PR #80, opened 2014-06-07
Add support for reading from HAR files

sirpengi:
Deserializer for HAR files.  A serializer would require information the Request class doesn't contain (like timing data), and would be intrusive to add at this point.


IvanMalison:
@sirpengi ping.

IvanMalison:
This has been inactive for too long

PR #136, opened 2015-02-24
Fix default port for HTTPS

abhinav:


abhinav:
master may be broken. pypy-httplib2 fails on my local copy of master as well (without this change).

IvanMalison:
Thanks. Did you look in to the test failure in the travis build?

abhinav:
Yes. The issue seems to be that the headers can be out of order for the assertion. Added a fix.

kevin1024:
Nice find!  What kind of bug did this cause?

abhinav:
None that I'm aware of. Just stumbled into it. :)

PR #138, opened 2015-03-23
update urllib3 patch/stub to be same as used for requests and add tests

### LINKED ISSUE #6: Add tests for urllib3
I'm testing requests, which tests urllib3 indirectly, but I need to add some tests which will test urllib3 directly, since the modules are monkeypatched at a different path.

### LINKED ISSUE #137: urllib3 patch does not seem to work
using `urllib3.PoolManager` directly (not via `requests`) but none of these `.request()` calls seem to be patched.

aisch:
what it does:

- reuse requests patching/stub code (since i think that is well tested/uses in vcrpy)
- add some integration tests for urllib3 (i used requests tests as base)
- update travis matrix to use those

closes #6, closes #137

### DIFF
--- a/tests/integration/test_urllib3.py
+++ b/tests/integration/test_urllib3.py
@@ -2,12 +2,10 @@
 
 # coding=utf-8
 
-import os
 import pytest
 import vcr
 from assertions import (
     assert_cassette_empty,
-    assert_cassette_has_one_response,
     assert_is_json
 )
 certifi = pytest.importorskip("certifi")
--- a/tests/integration/test_urllib3.py
+++ b/tests/integration/test_urllib3.py
@@ -4,10 +4,7 @@
 
 import pytest
 import vcr
-from assertions import (
-    assert_cassette_empty,
-    assert_is_json
-)
+from assertions import assert_cassette_empty, assert_is_json
 certifi = pytest.importorskip("certifi")
 urllib3 = pytest.importorskip("urllib3")
 
--- a/tests/integration/test_urllib3.py
+++ b/tests/integration/test_urllib3.py
@@ -1,0 +1,153 @@
+'''Integration tests with urllib3'''
+
+# coding=utf-8
+
+import os
+import pytest
+import vcr
+from assertions import (
+    assert_cassette_empty,
+    assert_cassette_has_one_response,
+    assert_is_json
+)
+certifi = pytest.importorskip("certifi")
+urllib3 = pytest.importorskip("urllib3")
+
+
+@pytest.fixture(params=["https", "http"])
+def scheme(request):
+    """
+    Fixture that returns both http and https
+    """
+    return request.param
+
+
+@pytest.fixture(scope='module')
+def verify_pool_mgr():
+    return urllib3.PoolManager(
+        cert_reqs='CERT_REQUIRED', # Force certificate check.
+        ca_certs=certifi.where()
+    )
+
+
+@pytest.fixture(scope='module')
+def pool_mgr():
+    return urllib3.PoolManager()
+
+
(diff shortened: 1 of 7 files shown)


IvanMalison:
Cool! This is great. Thanks for doing this.

Unfortunately, your build is failing. I think this is because you are assuming that urllib3 is going to present when it might not always be. It seems that the issue is caused by an import error. I also have a couple of nitpicky things that I'd like you to change. I'll comment on them in a moment.

aisch:
@IvanMalison thx for response so quick. will fix the build and then whatever issues u find on the pr.

tests/integration/test_urllib3.py
```diff
@@ -0,0 +1,153 @@
+'''Integration tests with urllib3'''
+
+# coding=utf-8
+
+import os
```
IvanMalison: This import is unused.

tests/integration/test_urllib3.py
```diff
@@ -0,0 +1,153 @@
...
+import pytest
+import vcr
+from assertions import (
+    assert_cassette_empty,
+    assert_cassette_has_one_response,
```
IvanMalison: So is this.

vcr/patch.py
```diff
@@ -229,6 +194,40 @@ def _boto(self):
         else:
             from .stubs.boto_stubs import VCRCertValidatingHTTPSConnection
             yield cpool, 'CertValidatingHTTPSConnection', VCRCertValidatingHTTPSConnection
+            
+    def _mock_urllib3_triples(self, cpool, stubs):
```
IvanMalison: This should probably be given a different name since it does not actually return triples, but context managers. Notice that though it does construct the triples internally, it applies `_build_patchers_from_mock_triples` at the end of the function definition.

Way to keep this DRY!

aisch:
@IvanMalison i agree w/ the test duplication (i copied from requests assuming they have very similar capability. but some cases, like e.g. sessions, not so those deleted and in others interface was pretty  different). isn't this a general issue for testing similar libs (httplib, urllib2, urllib3, etc)? e.g. `test_cross_scheme` is in test_urllib2, etc.

i can try to fix this just for urllib3 and requests since requests is interface over urrlib3 or we can create issue to consider how to solve it (e.g. creating a common testing http interface, implementing that for the ones supported and running common tests parameterized by each implementation).

IvanMalison:
> isn't this a general issue for testing similar libs (httplib, urllib2, urllib3, etc)? e.g. test_cross_scheme is in test_urllib2, etc.

Yeah, I guess it is. @kevin1024 wrote all the tests for everything except for requests, so I'm not really sure exactly what that looks like, but I agree that this is a problem that really ought to be solved in a pretty general.

> i can try to fix this just for urllib3 and requests since requests is interface over urrlib3 or we can create issue to consider how to solve it (e.g. creating a common testing http interface, implementing that for the ones supported and running common tests parameterized by each implementation).

Yeah I guess I'm fine if you just file an issue to try to clean up some of the duplication in testing.

kevin1024:
Yup, I can't think of a good way other than creating some kind of abstraction layer around making http calls including handling sessions and stuff.  Not even sure if that is a good idea or not.

aisch:
@IvanMalison :+1: thx for quick merge and release!

IvanMalison:
This broke tox because of the way the environments were specified. you need to put each version of urllib in the dependency list.

aisch:
sorry about that.. im not familiar w/ tox. perhaps it can be tested as part of travis?

IvanMalison:
No its fine, I already fixed it, it was just a problem with the way the environments were listed.

PR #144, opened 2015-03-31
Add ability to filter post data parameters

eadmundo:
Hi,

I needed to do this for some tests (Amazon Login requires passing the client secret in post data), and I didn't like having to put the values to filter in the method using `before_record`. Hope it's useful!

Cheers

Ed

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ with my_vcr.use_cassette('test.yml', filter_query_parameters=['api_key']):
 ```
 
 ### Filter information from HTTP post data
-Use the `filter_post_data_parameters` configuration option with a list of query
+Use the `filter_post_data_parameters` configuration option with a list of post data
 parameters to filter.
 
 ```python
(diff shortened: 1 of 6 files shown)


IvanMalison:
Thanks for doing this -- it seems like a good idea to me. Unfortunately, it seems that you have python 2.6 compatibility errors. See if you can fix those with a new commit and make the travis build pass.

eadmundo:
I hadn't done a fallback to the compat version of OrderedDict if the import of collections.OrderedDict failed.

kevin1024:
Yup, I think credentials in POST data is  a common enough case to merit specific support.  You can of course build a custom filter, but I originally justified the header filtering and the get param filtering as common cases.

My general philosophy is: if it's a common case, it should be easy.  If it's a weird case: it should be possible (but not necessarily easy).

eadmundo:
In fact I did originally build a custom filter, but I really disliked what I was going to have to do with the fields for filtering, since I couldn't pass them in as arguments - so I did the filter instead... agree with the general philosophy, and I suspect more than just Amazon Login requires secrets in POST data.

IvanMalison:
@kevin1024 
any reason why we aren't using the pypi package for ordered dict?
https://pypi.python.org/pypi/ordereddict
it could just be a dependency specified in setup.py, and that way we wouldn't have to have the code in the codebase/separately maintain it.

IvanMalison:
Actually, there is a backport of the entire collections module, which would allow us to get rid of the compat folder entirely.
https://pypi.python.org/pypi/backport_collections/0.1

kevin1024:
Nope, no reason that I can think of.  Nice find!

IvanMalison:
This is merged in fc4e985ee93a9180793058a33d5ed4e759c2d7cb

Issue #145, opened 2015-04-01
TypeError when using use_cassette(..., with_current_defaults=True)


samstav:
```pytb
Traceback (most recent call last):
  File "~/.virtualenvs/dev/lib/python2.7/site-packages/vcr/config.py", line 71, in use_cassette
    return Cassette.use(path, self.get_path_and_merged_config(path, **kwargs))
  File "~/.virtualenvs/dev/lib/python2.7/site-packages/vcr/cassette.py", line 84, in use
    return CassetteContextDecorator.from_args(cls, *args, **kwargs)
TypeError: from_args() takes exactly 3 arguments (4 given)
```

The following code will reproduce the error:

```python
import atexit
import shutil
import tempfile
import urllib2

import vcr

tempdir = tempfile.mkdtemp()
atexit.register(shutil.rmtree, tempdir)

recorder = vcr.VCR(cassette_library_dir=tempdir)

with recorder.use_cassette('hazip.yaml', record_mode='once'):
    ip = urllib2.urlopen('https://icanhazip.com').read()

with recorder.use_cassette('hazip.yaml', with_current_defaults=True,
                           record_mode='none'):
    ip = urllib2.urlopen('https://icanhazip.com').read()
```

PR #146, opened 2015-04-01
fix with_current_defaults causing TypeError

samstav:
The `from_args()` method in cassette.py was
throwing a `TypeError` when calling

```
use_cassette(..., with_current_defaults=True)
...
TypeError: from_args() takes exactly 3 arguments (4 given)
```

The path was then being passed to `use()` twice.

------------

Fixes https://github.com/kevin1024/vcrpy/issues/145

### DIFF
--- a/vcr/config.py
+++ b/vcr/config.py
@@ -69,7 +69,8 @@ class VCR(object):
 
     def use_cassette(self, path, with_current_defaults=False, **kwargs):
         if with_current_defaults:
-            return Cassette.use(path, self.get_path_and_merged_config(path, **kwargs))
+            path, config = self.get_path_and_merged_config(path, **kwargs)
+            return Cassette.use(path, **config)
         # This is made a function that evaluates every time a cassette is made so that
         # changes that are made to this VCR instance that occur AFTER the use_cassette
         # decorator is applied still affect subsequent calls to the decorated function.


IvanMalison:
Hey! Thanks for finding this. It would be great if there were a test that would not pass without this change. Do you think you could add that. If not I could do it myself.

samstav:
@IvanMalison I'll add the test.

I was experimenting with the `with_current_defaults` argument -- Can you explain what exactly this flag does?

IvanMalison:
I'm pretty sure I added it but I can't remember. Let me take a look.

IvanMalison:
So `with_current_defaults` makes it so that subsequent changes to the settings on the vcr do not affect a the settings used for a function decorated with use_cassette.

When it is not set, changes to settings on the vcr will affect the behavior of the decorated function.

samstav:
> So with_current_defaults makes it so that subsequent changes to the settings on the vcr do not affect a the settings used for a function decorated with use_cassette.

Subsequent, as in any changes made to the settings _after_ instantiating a `VCR` ?

IvanMalison:
> Subsequent, as in any changes made to the settings after instantiating a VCR ?

Well, yes, after the instantiation of a VCR, but more importantly after the decoration of a function.

> Asking because I will include tests for this logic too.

Of course, I'll give you an example below that might help you write a test.

> Also, does this only work when use_cassette() is used as a decorator, and not in a with statement? Are there any caveats/differences to using the decorator vs. the with statement?

There aren't really any differences as far as implementation is concerned, but because a with statement is a piece of code that executes immediately, this kwarg can't really have an effect.

Let me provide an example that might help clarify:

``` python
vcr = VCR()

@vcr.use_cassette('some_file.yaml')
def some_test():
   ... do some stuff

@vcr.use_cassette('some_file.yaml', with_current_defaults=True)
def other_test():
   ... do some stuff

vcr.cassette_library_dir = 'some_directory_to_keep_cassettes_in'

some_test()
other_test()
```

In this example, the cassette will be saved in the  'some_directory_to_keep_cassettes_in' in the case of `some_test`, but not in the case of other_test. `with_current_defaults` is not relevant to context manager use because the code is always executed immediately, so there is no oppotunity to change vcr settings before the code is executed.


This may seem like a strange/unimportant feature, but its actually really important because a lot of test runners import tests very eagerly aggresively, and that is the point at which the decorator runs. If you wanted to have for example, a flag that sets the cassette library youd have to worry about whether the tests get imported before the code that sets the library directory.

IvanMalison:
I went ahead and wrote the test myself since I realized it was sort of tricky after thinking about it for a bit (see https://github.com/kevin1024/vcrpy/blob/master/tests/unit/test_vcr.py#L110) if you are curious.

Thanks, again, for finding this.

samstav:
Cool. Thanks @IvanMalison

Issue #142, opened 2015-03-24
Use pytest-httpbin


kevin1024:
This would help builds run faster and probably pass more consistently.  It's also one of my main motivations for writing it.  I just need to get the motivation to actually *use* it :-P

[pytest-httpbin](https://github.com/kevin1024/pytest-httpbin)

IvanMalison:
Yeah, I had noticed that you hade made that library and was curious as to why vcrpy's tests don't use it. At least we have an issue for this now.

kevin1024:
Working on this a bit in a branch: https://github.com/kevin1024/vcrpy/tree/pytest-httpbin

danstender:
Just wanted to drop the note that we would appreciate it much for the Debian package [1] if the tests could run offline. Pytest-httpbin packaging is in progress.

[1] https://packages.qa.debian.org/v/vcr.py.html

kevin1024:
Cool!  Yeah, I've been super busy lately but if I ever get a free weekend I definitely want to finish this up.

danstender:
Welcome pytest-httpbin to Debian: https://tracker.debian.org/pkg/pytest-httpbin

yarikoptic:
or may be meanwhile there is a way to skip tests which require network connection @kevin1024 ?  thanks in advance for the clarification

danstender:
Any progress on this? Please combine your products :-)

kevin1024:
Would love to but been super busy lately!

kevin1024:
Now using it.  Not sure that the test suite is 100% offline yet but good progress!

PR #148, opened 2015-04-11
Ship extra bits with the pypi tarball.

ralphbean:
Thanks for the awesome lib!  :)

This change should make new tarballs uploaded to pypi include various
nice pieces:

- The README
- The LICENSE
- The tests

The text of the license actually specifies that the full text must be
distributed with all copies of the software.  So, you need it to be in
compliance with the MIT license.

The README is just nice to have, and the tests are particularly nice for
my use case.  I am packaging vcrpy for inclusion in the Fedora linux
distribution and:

- We like to use the tarball from pypi because it is the same source
  distribution that everyone else is using.
- We like to run the tests before we build the rpm in our build system
  to make sure nothing crazy is going on.

Of course, we can use the tarball for the source and then do a second
step to clone the source and get the tests.  But, this is more work than
we like if we can just get the tests added to the tarball.  Other
distributions (like Debian) like this too.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,4 @@
+include README.md
+include LICENSE.txt
+include tox.ini
+recursive-include tests *


kevin1024:
Wow, cool!  I don't see a problem with including that stuff in the tarball.  

I'm not sure why one of the tests failed on Travis, though.  Probably unrelated to this change?

If you're going to be running the tests in your build system you should know that sometimes they fail randomly due to failed network requests to httpbin.  At least until we do #142.

kevin1024:
Ran the tests again and they pass!

PR #153, opened 2015-05-07
Fix version of `six` dependency.

morganwahl:
`from six.moves.http_client import HTTPConnection` fails before version 1.5.0 of six. (on Python 2.7, at least.)

I had a particularly old copy of `six` in my virtualenv and ran into import errors in `vcrpy`.


kevin1024:
Thanks!

PR #154, opened 2015-05-13
Filter parameters from 'application/json' content-type POST requests

marco-santamaria:


IvanMalison:
test?

vcr/filters.py
```diff
@@ -31,13 +32,20 @@ def remove_query_parameters(request, query_parameters_to_remove):
...
-        post_data = OrderedDict()
-        for k, sep, v in [p.partition(b'=') for p in request.body.split(b'&')]:
-            if k in post_data:
-                post_data[k].append(v)
-            elif len(k) > 0 and k.decode('utf-8') not in post_data_parameters_to_remove:
-                post_data[k] = [v]
-        request.body = b'&'.join(
-            b'='.join([k, v])
-            for k, vals in post_data.items() for v in vals)
+        try:
```
IvanMalison: Why not look at the content type header instead of just try excepting?

marco-santamaria:
OK, yesterday I was in a hurry :)
I added some tests and removed the try/except.
Please note that with JSON post data it doesn't look a good idea to me accepting duplicate keys.
Also, with the proposed implementation the order of the keys can change: I don't think it could be a problem.

marco-santamaria:
Yes that was the reason. I could have rewritten with dict, but is already merged. Thanks

PR #155, opened 2015-05-15
Crash when cassette path contains cassette_library_dir

gazpachoking:
If an absolute path containing the `cassette_library_dir` is specified when using `vcr.use_cassette`, the transformer to add the cassette_library_dir will return None, resulting in a crash.
```
  File "...\vcr\persist.py", line 6, in load_cassette
    with open(cassette_path) as f:
TypeError: coercing to Unicode: need string or buffer, NoneType found
```
This adds a test for it, (and another one just testing a filename path to make sure the transformer is working,) as well as fixes the bug.

### DIFF
--- a/tests/unit/test_vcr.py
+++ b/tests/unit/test_vcr.py
@@ -147,6 +147,24 @@ def test_cassette_library_dir_with_decoration_and_no_explicit_path():
     function_name()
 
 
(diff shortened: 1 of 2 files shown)


IvanMalison:
Thanks for catching this/the test.

gazpachoking:
No prob! Thanks for the awesome library. :smile:

Issue #158, opened 2015-06-24
"TypeError: Can't convert 'bytes' object to str implicitly" whilst recording 'requests' session.


MrJohz:
Specifically, whilst removing post parameters:

```python
    def remove_post_data_parameters(request, post_data_parameters_to_remove):
        if request.method == 'POST' and not isinstance(request.body, BytesIO):
            if ('Content-Type' in request.headers and
                    request.headers['Content-Type'] == 'application/json'):
                json_data = json.loads(request.body.decode('utf-8'))
                for k in list(json_data.keys()):
                    if k in post_data_parameters_to_remove:
                        del json_data[k]
                request.body = json.dumps(json_data).encode('utf-8')
            else:
                post_data = OrderedDict()
                # fails on following line due to bytes/str incompatibility.
                for k, sep, v in [p.partition(b'=') for p in request.body.split(b'&')]:
                    if k in post_data:
```

Example file that will fail:
```python
import vcr
import requests

player = vcr.VCR(
    serializer='yaml',
    record_mode='once',
    cassette_library_dir='cassettes',
    match_on=['uri', 'method'],
    filter_post_data_parameters=['username', 'password'])

post_body = {'username': 'top-secret-needs-filtering',
             'password': 'top-secret-needs-filtering'}

with player.use_cassette('test_post.yaml'):
    requests.post('http://httpbin.org/post', data=post_body)
```

Example console shell of previous file:
```console
$ python --version
Python 3.4.2

$ pip freeze
backport-collections==0.1
contextlib2==0.4.0
mock==1.0.1
PyYAML==3.11
requests==2.7.0
six==1.9.0
vcrpy==1.5.2
wrapt==1.10.4

$ python fail.py 
Traceback (most recent call last):
  File "/home/jonathan/.pyenv/versions/failvcrpy/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py", line 372, in _make_request
    httplib_response = conn.getresponse(buffering=True)
  File "/home/jonathan/.pyenv/versions/failvcrpy/lib/python3.4/site-packages/vcr/stubs/__init__.py", line 219, in getresponse
    if self.cassette.can_play_response_for(self._vcr_request):
  File "/home/jonathan/.pyenv/versions/failvcrpy/lib/python3.4/site-packages/vcr/cassette.py", line 201, in can_play_response_for
    request = self._before_record_request(request)
[...]

PR #159, opened 2015-06-24
Fix issue #158

MrJohz:
I've included a requests-specific test, because I couldn't reproduce the bug with just urllib and I didn't want to fiddle around with all of the other libraries.  However, the fix should apply to any library that has the same issue: sending a request with the request body in a non-binary form.  What I've done is to check if the request body is non-binary (using six's `text_type`) and if so encode it using the 'utf-8' encoding that's also used later.


IvanMalison:
@eadmundo I'm going to merge with when Travis passes, but you should take a look at this since you added this functionality originally.

eadmundo:
sorry for the slow reply, but looks good to me

PR #161, opened 2015-06-30
packaging fixes

graingert:


IvanMalison:
Did you do this by hand or did you use something like pandoc?

graingert:
Pandoc. PyPI doesn't support anything except rst

IvanMalison:
yep. Just wanted to make sure it wasn't by hand.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[bdist_wheel]
```
IvanMalison: What does this do?

setup.cfg
```diff
@@ -0,0 +1,2 @@
...
+universal=1
```
IvanMalison: Nice. So this creates a python2/3 compatible distribution?

graingert: It allows you to create universal wheels when you run:

`python setup.py sdist bdist_wheel register upload`

PR #164, opened 2015-07-02
Tornado support

### LINKED ISSUE #120: Support for Tornado HTTP
Is Tornado's HTTP client not supported? Thanks

http://tornado.readthedocs.org/en/latest/httpclient.html

abhinav:
This PR adds support for Tornado's AsyncHTTPClient.

This doesn't support all of AsyncHTTPClient's functionality. An exception is thrown if that functionality is used inside the vcrpy context.

Resolves #120.

### DIFF
--- a/vcr/stubs/tornado_stubs.py
+++ b/vcr/stubs/tornado_stubs.py
@@ -1,0 +1,147 @@
+'''Stubs for tornado HTTP clients'''
+from __future__ import absolute_import
+
+from six import BytesIO
+
+from tornado import httputil
+from tornado.httpclient import AsyncHTTPClient
+from tornado.httpclient import HTTPResponse
+from tornado.simple_httpclient import SimpleAsyncHTTPClient
+
+from vcr.errors import CannotOverwriteExistingCassetteException
+from vcr.request import Request
+
+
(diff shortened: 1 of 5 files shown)


vcr/stubs/tornado_stubs.py
```diff
@@ -0,0 +1,147 @@
...
+        # TODO body_producer, header_callback, and streaming_callback are not
+        # yet supported.
+
+        unsupported_call = (
+            request.body_producer is not None or
+            request.header_callback is not None or
+            request.streaming_callback is not None
+        )
+        if unsupported_call:
+            response = HTTPResponse(
```
IvanMalison: Why not just actualy raise an exception here? Will this get caught somewhere in Tornado or something?

abhinav: It won't get caught in Tornado but it'll be raised in an unexpected context. `fetch` returns a future and exceptions for that operation are expected inside that future. If this was just a simple `raise`, the call to fetch will throw an exception instead of the future. It will also not respect the `raise_error` flag if `raise`d directly.

BTW, there was a bug here because the exception wasn't actually being raised (I forgot to call the callback). Fix and more tests in #165.

PR #167, opened 2015-07-04
Improve body matcher using Content-Type based approach

### LINKED ISSUE #166: Error with body matcher for json, xmlrpc and form urlencoded
This is a tricky issue I encountered when using the body matcher on xmlrpc requests.

Symptoms: Sometimes the request won't match, sometimes it will, and this only affects certain requests. This occurs on Python 3.4 and I believe other python 3 versions but not on 2.7

Cause: An XMLRPC request has a body with XML inside which is generated from the parameters passed to the function call. Some parameters can be of dict type. xmlrpclib (or xmlrpc.client) will loop over items of the dict and generate the appropriate XML which will be the body of our POST request. Now items order is not guaranteed in dict and the behavior changed in python 3 such that the order of the same dict *can change anytime* and is not *more or less constant on the same computer* as in python 2. So the generated XML won't be necessarily the same as the one you recorded.

Fix suggestion: A custom xmlrpc body matcher that takes that into account and will compare the `struct` XML elements in the correct order.

The gzip compression didn't help me debuging this as I couldn't even read directly the cassettes...

Diaoul:
Fixes #166

### DIFF
--- a/vcr/matchers.py
+++ b/vcr/matchers.py
@@ -1,6 +1,6 @@
 import json
 from six.moves import urllib, xmlrpc_client
-from .util import CaseInsensitiveDict
+from .util import CaseInsensitiveDict, read_body
 import logging
 log = logging.getLogger(__name__)
 
@@ -34,18 +34,12 @@ def query(r1, r2):
 
 
 def raw_body(r1, r2):
-    if hasattr(r1.body, 'read') and hasattr(r2.body, 'read'):
-        return r1.body.read() == r2.body.read()
-    return r1.body == r2.body
+    return read_body(r1) == read_body(r2)
 
 
 def body(r1, r2):
-    if hasattr(r1.body, 'read') and hasattr(r2.body, 'read'):
-        r1_body = r1.body.read()
-        r2_body  = r2.body.read()
-    else:
-        r1_body = r1.body
-        r2_body  = r2.body
+    r1_body = read_body(r1)
+    r2_body  = read_body(r2)
     r1_headers = CaseInsensitiveDict(r1.headers)
     r2_headers = CaseInsensitiveDict(r2.headers)
     if r1_headers.get('Content-Type') == r2_headers.get('Content-Type') == 'application/x-www-form-urlencoded':
--- a/vcr/matchers.py
+++ b/vcr/matchers.py
@@ -1,5 +1,6 @@
 import json
 from six.moves import urllib, xmlrpc_client
+from .util import CaseInsensitiveDict
 import logging
 log = logging.getLogger(__name__)
 
@@ -45,12 +46,14 @@ def body(r1, r2):
     else:
         r1_body = r1.body
         r2_body  = r2.body
-    if r1.headers.get('Content-Type') == r2.headers.get('Content-Type') == 'application/x-www-form-urlencoded':
+    r1_headers = CaseInsensitiveDict(r1.headers)
+    r2_headers = CaseInsensitiveDict(r2.headers)
+    if r1_headers.get('Content-Type') == r2_headers.get('Content-Type') == 'application/x-www-form-urlencoded':
         return urllib.parse.parse_qs(r1_body) == urllib.parse.parse_qs(r2_body)
-    if r1.headers.get('Content-Type') == r2.headers.get('Content-Type') == 'application/json':
+    if r1_headers.get('Content-Type') == r2_headers.get('Content-Type') == 'application/json':
         return json.loads(r1_body) == json.loads(r2_body)
-    if ('xmlrpc' in r1.headers.get('User-Agent', '') and 'xmlrpc' in r2.headers.get('User-Agent', '') and
-        r1.headers.get('Content-Type') == r2.headers.get('Content-Type') == 'text/xml'):
+    if ('xmlrpc' in r1_headers.get('User-Agent', '') and 'xmlrpc' in r2_headers.get('User-Agent', '') and
+        r1_headers.get('Content-Type') == r2_headers.get('Content-Type') == 'text/xml'):
         return xmlrpc_client.loads(r1_body) == xmlrpc_client.loads(r2_body)
     return r1_body == r2_body
 
--- a/vcr/matchers.py
+++ b/vcr/matchers.py
@@ -1,3 +1,5 @@
+import json
+from six.moves import urllib, xmlrpc_client
 import logging
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


Diaoul:
No unit test because dict order is not deterministic.

Diaoul:
More details in the commit message.

IvanMalison:
Your build is failing because it does not import parse/parse_qs in a python2/3 compatible way. It should be imported from the six library.

Diaoul:
Yep I saw that, fixed it now waiting for the new build to finish. Looks good so far.

IvanMalison:
This doesn't seem to deal with gunzipping the requests. Was that not an issue?

Diaoul:
The issue I had with gunzipping was that it makes hard to read cassettes for debugging and disabling it meant editing xmlrpclib which I was not comfortable doing. The body variable is plain text when debugging.

IvanMalison:
This needs unit tests. At the very least, you can create a json/xml strings manually that should match and hit the matcher directly. I'd also like to see the desired behavior when the content types don't match etc.

Diaoul:
Unfortunately we can't assume that the test will fail when not comparing JSON with JSON loads. Sometimes it will pass just fine due to the nature of the issue.
Maybe I can try using ordered dict here purely for testing purposes.

IvanMalison:
Thats why I'm saying that we should unit test the function directly by manually constructing the json/xml string.

IvanMalison:
You could also just pass the string data directly into the http library you are using in the same way, and order it differently in the two requests.

vcr/matchers.py
```diff
@@ -30,12 +32,29 @@ def query(r1, r2):
...
 
 
+def body(r1, r2):
+    if hasattr(r1.body, 'read') and hasattr(r2.body, 'read'):
+        r1_body = r1.body.read()
+        r2_body  = r2.body.read()
+    else:
+        r1_body = r1.body
+        r2_body  = r2.body
+    if r1.headers.get('Content-Type') == r2.headers.get('Content-Type') == 'application/x-www-form-urlencoded':
```
IvanMalison: hmmmm. I think that we need to do lowers/uppers on both 'Content-Type' and 'application/x-www-form-urlencoded'. We might even need to do a case insensitive regex match. I think that both of those strings are allowed to have different casings.

IvanMalison: This is why requests has

https://github.com/kennethreitz/requests/blob/v1.2.3/requests/structures.py#L37

@kevin1024 Might be useful to have this here, but I'm not sure we want to add a strict dependency on requests. Thoughts? Maybe we can just copy the implementation and add it vcrpy?

kevin1024: I would rather not add requests as a dependency.

We also have to be careful here - the requests implementation of CaseInsensitiveDict lowercases the header keys.  Someone might actually care about the capitalization of headers.  Yeah there's something in the RFC that says headers are case-insensitive, but I think keeping VCR transparent is very important.

Since VCR is used mainly for integration tests, it should pass through all the weirdness of the server and client that it's working with, that way it can help find bugs in both.  It shouldn't be helping out by changing headers.

But yeah, the header comparison should probably be case-insensitive for our "smart body matcher".  Just don't store them that way.

IvanMalison: > But yeah, the header comparison should probably be case-insensitive for our "smart body matcher". Just don't store them that way.

Oh yeah, agreed. What I was saying is that we can't check for the 'Content-Type' key or assume that 'application/x-www-form-urlencoded' will always be totally lower cased, and so that here we should take request.headers and insert it into a case insensitve dict in order to check for the header.

kevin1024: Yup, in that case it's fine, as long as it's just used for this matcher.  Maybe we could add it to `vcr/datastructs.py`?

I can picture someone wanting to write a custom matcher where they care about header capitalization, so I don't think I want to change the matcher API - just use the CaseInsensitiveDict for looking up the content-type of our "smart body matcher".

This raises questions about whether our header matcher should also do some kind of case-insensitive match, but I can see both sides on that one.

IvanMalison: I think leaving the header matcher as is is fine for now. We could maybe add another one that specifically indicates that the matching is case insensitive, but keep in mind that this matching should be happening on requests that are generated from the same test code so there probably shouldn't be differences in the casing the requst headers.

One edge case we might (or might not) want to think about is the case where two headers with different casing are sent with the request e.g.

'Content-Type' and 'content-type'. The behavior of CaseInsensitveDict is undefined in this case.

Diaoul: I didn't want to bring requests' CaseInsensitveDict with this PR as it is out of scope. I think headers should be case insensitive for the whole vcrpy not just this matcher. So use CaseInsensitveDict in the Request object. This will make the headers matcher more RFC compliant too and do what any developer would expect from header matching (i.e. case insensitive matching)

IvanMalison: I'm not necessarily against this idea, but the point is that this PR can not go in to master as is, because its behavior is inconsistent with respect to well formed content-type headers.

You'll have to convince @kevin1024 that this is the right thing to do since he said:
> I can picture someone wanting to write a custom matcher where they care about header capitalization, so I don't think I want to change the matcher API - just use the CaseInsensitiveDict for looking up the content-type of our "smart body matcher".

I can see the merits of both approaches, so I'm not going to take a side on this point, but I think that the capitlization issue definitely needs to be addressed in the content-type body matcher in some way or another.

IvanMalison: @kevin1024 Do you have any thoughts on ^?

@Diaoul Whats the status of this PR? Do you not want to carry it through if it requieres doing the caseinsensitve dict thing as well?

kevin1024:
Yeah, I consider the cassettes exposed to tests (`with vcr.use_cassette as cassette`) as part of the API so changing their behavior would be a major version bump.  I can see both sides too but I would a fairly convincing reason to for an API change like this.  A couple of points to consider:

1 - If at all possible, I *really* want to avoid the case where your test passes with VCR but fails without it.  I want to avoid this at all costs, what's what I mean when I want to keep VCR transparent.
2 - VCR is usually used for integration tests, and I don't want it covering up any kind of weird behavior in either the client or the server.  Requests actually does the header capitalization normalization automatically, but other HTTP client's don't.

Diaoul:
I don't think we're going to see a fairly convincing reason here. I can see the reason why you want to keep the header dict case sensitive. Doing otherwise would mean a potential loss of information.

I suggest we use the CaseInsensitiveDict inside the smart body matcher. I will update the PR and add a few unit tests when I have some time available.

Whether to extend it to the headers matcher that's up to you. Servers answering the request won't care about case sensitivity in the headers and that would be more RFC compliant. We can also add a raw_headers matcher if you want to keep case sensitivity matching capability. That's out of scope for this PR though.

IvanMalison:
> I suggest we use the CaseInsensitiveDict inside the smart body matcher. I will update the PR and add a few unit tests when I have some time available.

Yep. That is exactly what I want to do.

Diaoul:
@IvanMalison, @kevin1024: I added what we discussed. The travis build looks good so far.

If everything's good for you, could you make a release with this soon?

Diaoul:
Any updates on this?

IvanMalison:
Looks pretty good to me. Would be nice to still have a test that makes sure that tests that we still say the bodies don't match (all the tests test that things match. What about making sure things don't match?).

vcr/matchers.py
```diff
@@ -30,12 +33,31 @@ def query(r1, r2):
...
-def body(r1, r2):
+def raw_body(r1, r2):
     if hasattr(r1.body, 'read') and hasattr(r2.body, 'read'):
         return r1.body.read() == r2.body.read()
     return r1.body == r2.body
 
 
+def body(r1, r2):
+    if hasattr(r1.body, 'read') and hasattr(r2.body, 'read'):
+        r1_body = r1.body.read()
```
IvanMalison: Since you do the exact same thing to r1 and r2, it seems like it might be a good idea to write a function that performs all of these operations.

Think it would help a bunch with readability here.

Diaoul:
There are infinite possibilities for two requests not to match. Everything cannot be unit tested. Do you have a particular test in mind? Two xmlrpc requests with different headers? One xmlrpc and one json?

This is not a very complex function, more tests would probably end up testing the `CaseInsensitiveDict` (which is also quite simple and already proven working fine as it comes from requests) or the `==` operator. I was planning to add more tests if someone encounters a bug.

PR #168, opened 2015-07-05
Fix RST parse errors generated by pandoc

graingert:


graingert:
pasted into http://rst.ninjs.org/ and fixed the error(s) manually

graingert:
Woot looks like we're getting nice HTML on https://pypi.python.org/pypi/vcrpy/1.6.1

IvanMalison:
@graingert Nice! vcrpy really deserves sphinx docs. Would be awesome to get on readthedocs.org

Issue #169, opened 2015-07-05
Upgrading to 1.6.0 seems conditional extra_require are not working on setup


ialex:
I receive this error when i upgraded to 1.6.0:
```
Traceback (most recent call last):
  File "run_tests.py", line 14, in <module>
    from tests.app_login_tests import MainSiteTest  # noqa
  File "/my_app/tests/app_login_tests.py", line 7, in <module>
    from vcr_setup import vcrlib
  File "/my_app/tests/vcr_setup.py", line 1, in <module>
    import vcr
  File "/usr/local/lib/python2.7/dist-packages/vcr/__init__.py", line 2, in <module>
    from .config import VCR
  File "/usr/local/lib/python2.7/dist-packages/vcr/config.py", line 8, in <module>
    from .compat import collections
  File "/usr/local/lib/python2.7/dist-packages/vcr/compat.py", line 4, in <module>
    import mock
ImportError: No module named mock
```

Looking at the commits i think this new conditionals imports are not working https://github.com/kevin1024/vcrpy/commit/41949f7dc64a2ddb8a9c2cb845eac7bb8b94380b#diff-2eeaed663bd0d25b7e608891384b7298R36 i'm installing the package using pip im on python 2.7.

kevin1024:
Hmm, what version of pip are you running?  `pip -V`

I wonder if the conditional imports are a new feature.

ialex:
pip 1.5.4 from /usr/lib/python2.7/dist-packages (python 2.7)

IvanMalison:
hmm wouldn't it be the version of setup_tools that he has that matters? I guess it could be a pip thing, but setup_tools seems more likely.

IvanMalison:
@ialex you can run `python -c 'from setuptools.version import __version__ as version; print(version)'` to check on this

IvanMalison:
@kevin1024 this raises the question: Do we want to depend on new features in setuptools?

kevin1024:
Ah yeah probably a setuptools version.

We probably want this to work on old versions of setuptools.  A lot of people don't have much control over their environment (yum install setuptools on RHEL5 installs some ancient version) and I don't want there to be any pain getting vcrpy installed.

IvanMalison:
hmmm so you want to revert @graingert 's change in 41949f7dc64a2ddb8a9c2cb845eac7bb8b94380b?
I can see your reasoning. Maybe he can chime in on this?

graingert:
[...]

PR #170, opened 2015-07-07
Support conditional requirements in old versions of pip

graingert:
idea from https://github.com/jkbrzt/httpie/blob/master/setup.py#L46

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 import sys
 from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
+import pkg_resources
 
 long_description = open('README.rst', 'r').read()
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -20,6 +21,21 @@ def run_tests(self):
...
+
+
+extras_require = {
+    ':python_version in "2.4, 2.5, 2.6"':
+        ['contextlib2', 'backport_collections', 'mock'],
+    ':python_version in "2.7, 3.1, 3.2"': ['contextlib2', 'mock'],
+}
+
+
+if 'bdist_wheel' not in sys.argv:
```
IvanMalison: can we not do a version check here too. Won't this make it do it unconditionally?

graingert:
@IvanMalison I see you made a comment, (I got an email notification) but it's gone now...

IvanMalison:
Yeah i was worried that we would do things unconditionally, but I was just misreading the code. Looks good to me. I wish there were a way we could test.

graingert:
I did a:
```
pip install pip==1.5
pip install .
```
in my local checkout

IvanMalison:
yeah I guess we could add a travis environment that does the same thing?

IvanMalison:
closes #169

kevin1024:
Nice! Thanks @graingert.

PR #171, opened 2015-07-14
[RFC] Deserialize request bodies to byte string, not text string (fixes #163).

### LINKED ISSUE #163: Sharing cassettes between Python 2 and 3 is tricky (when matching on body)
I have a test that needs to match on the request body, because it sends a POST. This works fine in a pure Python 2 world, and it works fine in a pure Python 3 world. Things get tricky when you need the same code to work under both Python 2 and 3.

Specifically, if I first run my tests under Python 2, then my cassette files store bodies like this:
```
interactions:
- request:
    body: 'hello world

      how are you today?'
[...]
```

If I then run my test under Python 3, it fails because the bodies don't match:
```
E
======================================================================
ERROR: test_echo_with_body (tests.test_echo.TestEchoClient)
----------------------------------------------------------------------
  File "/home/gward/src/python-renesys-apis/tests/test_echo.py", line 24, in test_echo_with_body
    response = call_echo(body)
  File "/home/gward/src/python-renesys-apis/tests/test_echo.py", line 10, in call_echo
    return urllib.request.urlopen(req)
  File "/usr/lib/python3.4/urllib/request.py", line 161, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.4/urllib/request.py", line 463, in open
    response = self._open(req, data)
  File "/usr/lib/python3.4/urllib/request.py", line 481, in _open
    '_open', req)
  File "/usr/lib/python3.4/urllib/request.py", line 441, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.4/urllib/request.py", line 1210, in http_open
    return self.do_open(http.client.HTTPConnection, req)
  File "/usr/lib/python3.4/urllib/request.py", line 1185, in do_open
    r = h.getresponse()
  File "/var/lib/venv/ren3/lib/python3.4/site-packages/vcr/stubs/__init__.py", line 236, in getresponse
    self.cassette.record_mode)
nose.proxy.CannotOverwriteExistingCassetteException: No match for the request (<Request (POST) http://gerg.ca/echo.cgi>) was found. Can't overwrite existing cassette ('echo_with_body.yaml') in your current record mode ('once').
-------------------- >> begin captured logging << --------------------
vcr.cassette: DEBUG: Entered context for cassette at echo_with_body.yaml.
vcr.stubs: DEBUG: Got <Request (POST) http://gerg.ca/echo.cgi>
vcr.matchers: DEBUG: Requests <Request (POST) http://gerg.ca/echo.cgi> and <Request (POST) http://gerg.ca/echo.cgi> differ according to the following matchers: [(False, <function body at 0x7fb22cb21e18>)]
vcr.cassette: DEBUG: Exiting context for cassette at echo_with_body.yaml.
--------------------- >> end captured logging << ---------------------
```

The workaround appears to be to regenerate my cassettes under Python 3. Then they look like this:
```
interactions:
- request:
    body: !!binary |
      aGVsbG8gd29ybGQKaG93IGFyZSB5b3UgdG9kYXk/
[...]
```
and the tests pass under Python 2 and 3.

Two problems:
* regenerating all my cassettes is a pain the neck; I'd rather not if I can avoid it
* if that is the best we can do, it really should be documented!

gward:
If you like this fix, please DO NOT MERGE IT yet!! There is unfinished business in my refactoring of vcr.serializers.compat: I refactored `convert_body_to_bytes()` to be cleaner and simpler, but left `convert_body_to_unicode()` as-is. I don't need it for this fix, and didn't want to waste time on it in case you reject this PR.

But if you *do* like this approach, I should finish the refactoring that I started. So please let me know!

### DIFF
--- a/tests/fixtures/py2body_cassette.yaml
+++ b/tests/fixtures/py2body_cassette.yaml
@@ -1,0 +1,16 @@
+interactions:
+- request:
+    body: x=5&y=2
+    headers:
+      Content-Length: ['7']
+      Content-Type: [application/x-www-form-urlencoded]
+      Host: [httpbin.org]
+    method: POST
+    uri: http://httpbin.org/post
+  response:
+    body: {string: !!python/unicode "{}\n"}
+    headers:
+      content-length: ['3']
+      content-type: [application/json]
+    status: {code: 200, message: OK}
+version: 1
--- a/tests/fixtures/py3body_cassette.yaml
+++ b/tests/fixtures/py3body_cassette.yaml
@@ -1,0 +1,17 @@
+interactions:
+- request:
+    body: !!binary |
+      eD01Jnk9Mg==
+    headers:
+      Content-Length: ['7']
+      Content-Type: [application/x-www-form-urlencoded]
+      Host: [httpbin.org]
+    method: POST
+    uri: http://httpbin.org/post
+  response:
+    body: {string: "{}\n"}
+    headers:
+      content-length: ['3']
+      content-type: [application/json]
+    status: {code: 200, message: OK}
+version: 1
--- a/tests/unit/test_serialize.py
+++ b/tests/unit/test_serialize.py
@@ -27,6 +27,22 @@ def test_deserialize_new_json_cassette():
         deserialize(f.read(), jsonserializer)
 
 
+def test_deserialize_py2_yaml_cassette():
+    # A cassette generated under Python 2 stores the body as a string, but
+    # the same cassette generated under Python 3 stores it as "!!binary".
+    # Make sure we accept the Python 2 cassette, regardless of whether
+    # we're running under Python 2 or 3.
+    with open('tests/fixtures/py2body_cassette.yaml') as f:
+        (requests, responses) = deserialize(f.read(), yamlserializer)
+    assert requests[0].body == b'x=5&y=2'
+
+def test_deserialize_py3_yaml_cassette():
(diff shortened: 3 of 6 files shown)


graingert:
Can you check some round tripping of different encodings? Eg if I post something as UTF16

IvanMalison:
closes #163.
I'll Second what @graingert . I'd generally like to see a bit more testing. This is tricky business.
@kevin1024 Not sure I get the intention behind the original behavior of this function, can you clarify?

graingert:
I think on serialising vcrpy should decode body bytes using utf8+surrogateescape. This will need a backported implementation for python 2

gward:
The commit I just pushed should fix the tests I inadvertently broke. This has nothing to do with @graingert's and @IvanMalison's comments: I'll try to get to them later.

gward:
Oh yeah forgot to mention: commit 99770a7 means the refactoring in 778c8bd wasn't necessary after all. However, I still think it's a nice cleanup. But if you want to see a new PR that ditches the refactoring and focuses only on the bug, let me know.

gward:
@graingert said:
> I think on serialising vcrpy should decode body bytes using utf8+surrogateescape. This will need a backported implementation for python 2

It's unclear if you're asking me to add this as part of this bug fix, or this is a general "we should do this" comment.

gward:
New pull request coming with smaller patch, no refactoring required, and more extensive tests.

PR #175, opened 2015-07-16
Fix for #163, take 2.

gward:
Compared to my previous PR (#171):
* no refactoring of vcr/serializers/compat.py -- turns out to be unnecessary
* more extensive tests

### DIFF
--- a/tests/unit/test_serialize.py
+++ b/tests/unit/test_serialize.py
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 import pytest
 
 from vcr.compat import mock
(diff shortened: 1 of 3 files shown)


graingert:
I'd really like to be able to see the body in the cassette file where possible.

graingert:
Eg utf8 bodies display as possible text in the yaml file.

gward:
> I'd really like to be able to see the body in the cassette file where possible.

So would I. But this PR is focused on fixing a bug with accepting *existing* YAML cassettes. I need this fix so I can continue porting code to Python 3. I cannot use VCR for a Python 2/3 codebase without this fix (or without a gross workaround).

So can we please get this fix in, and then see about nicer-looking request bodies in YAML cassette files?

gward:
Alternately, please let me know if my fix looks good. Worst case, I'll extract the patch and apply it to our RPM of VCRpy.

IvanMalison:
This looks good to me. @kevin1024 @graingert thoughts?

PR #179, opened 2015-07-29
Support distribute, Fixes #178

### LINKED ISSUE #178: AttributeError: 'module' object has no attribute 'evaluate_marker'
I'm getting the following when trying to install `vcrpy`:

```
Downloading/unpacking vcrpy==1.6.1 (from -r requirements-tests.txt (line 12))
  Using download cache from ...
  Running setup.py egg_info for package vcrpy
      File "<string>", line 16, in <module>
      File ".../env/build/vcrpy/setup.py", line 36, in <module>
        if key.startswith(':') and pkg_resources.evaluate_marker(key[1:]):
    AttributeError: 'module' object has no attribute 'evaluate_marker'
    Complete output from command python setup.py egg_info:

  File "<string>", line 16, in <module>

  File ".../env/build/vcrpy/setup.py", line 36, in <module>

    if key.startswith(':') and pkg_resources.evaluate_marker(key[1:]):

AttributeError: 'module' object has no attribute 'evaluate_marker'
```

graingert:


graingert:
Fixes #178

graingert:
This is only an option, if we really need to support Distribute.

IvanMalison:
@graingert I don't really see the harm in adding this, so I'm inclined to merge. @charlax Do you really care about supporting distribute?

kevin1024:
:+1:, I think VCR should support whatever outdated packaging libraries that might be installed on someone's system (within reason) to lower the barrier to entry.

graingert:
When would you drop support for Distribute?

IvanMalison:
I'm with @kevin1024 on this one. @graingert I'm not really sure I see a good reason to ever drop support for distribute. It's good to avoid installing extraneous packages, but I think that it's ultimately more important that this libarary works for as many people as possible.

charlax:
Thanks for landing this! I actually had a similar PR I was going to submit.

I'm not sure why the setup is so outdated on the CI machine I was running this on, but in any case I agree that it's good thing to be as forgiving as possible.

PR #180, opened 2015-07-30
Fix exception catching in coroutines.

abhinav:
If you don't send exceptions back to the generator, they will raise and fail the test even if the generator is catching them at the yield point.

### DIFF
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -1,3 +1,4 @@
+import sys
 import inspect
 import logging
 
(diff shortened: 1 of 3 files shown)


vcr/cassette.py
```diff
@@ -87,16 +88,20 @@ def _execute_function(self, function, args, kwargs):
         return handler(function, args, kwargs)
 
     def _handle_coroutine(self, function, args, kwargs):
+        # Wrap a bi-directional generator so that we're inside the cassette
```
IvanMalison: Make this a proper docstring.

abhinav: Done

vcr/cassette.py
```diff
@@ -87,16 +88,21 @@ def _execute_function(self, function, args, kwargs):
         return handler(function, args, kwargs)
 
     def _handle_coroutine(self, function, args, kwargs):
+        """Wraps a bi-directional generator so that we're inside the cassette
```
IvanMalison: nitpick but
bi-directional generator == coroutine, right?

I'd prefer to use the typical terminolgy here.

vcr/cassette.py
```diff
@@ -87,16 +88,21 @@ def _execute_function(self, function, args, kwargs):
...
-                    to_yield = coroutine.send(to_send)
-                except StopIteration:
-                    break
-                else:
                     to_send = yield to_yield
+                except Exception:
+                    to_yield = coroutine.throw(*sys.exc_info())
+                else:
+                    to_yield = coroutine.send(to_send)
+            # We don't need to catch StopIteration. The caller (Tornado's
```
IvanMalison: Why is this comment at a different indentation level. Let's put it above `to_yield = `

vcr/cassette.py
abhinav: Right. Fixed.

vcr/cassette.py
abhinav: Fixed

PR #186, opened 2015-08-14
Capture effective url in Tornado

mvschaik:
Most of the libraries capture redirects in multiple requests, but tornado doesn't. This PR adds the 'url' property to the response keep the `.effective_url`.

Probably it would be more in line with the other libraries to capture all the responses separately to also populate the `.history` of the response.

I've also added some tests to other libraries, but they don't have this issue.

### DIFF
--- a/tests/integration/test_httplib2.py
+++ b/tests/integration/test_httplib2.py
@@ -56,6 +56,17 @@ def test_response_headers(scheme, tmpdir):
         resp, _ = httplib2.Http().request(url)
         assert set(headers) == set(resp.items())
 
+def test_effective_url(scheme, tmpdir):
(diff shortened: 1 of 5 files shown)


IvanMalison:
Thanks for writing tests! Super helpful.

Hmm so does that mean that recorded requests in tornado won't be replayed properly when replayed for other libs and vice versa?

mvschaik:
That is correct. A preferable solution would be to have the same behavior as the other libs, but that would mean capturing the requests in a different way.

IvanMalison:
@kevin1024 @abhinav I'm going to merge this, but you guys should take a glance at the above conversation. Probably not a huge deal though.

kevin1024:
:+1: thanks!!!!!

Issue #189, opened 2015-08-21
Environment variable configuration


Diaoul:
It would be nice to be able to specify a global `record_once` from environment variable. Something like `VCRPY_RECORD_ONCE` is unique enough not to conflict with anything.

Would you consider this feature?

kevin1024:
How about something like this:

```python
import vcr

my_vcr = vcr.VCR(
    record_mode=os.environ.get['VCR_RECORD_MODE']
)

with my_vcr.use_cassette('test.json'):
    # your http code here
```

Diaoul:
I know I can do it in my code, I just believe this could be a nice feature and would integrate well with Travis for example without extra code.

Diaoul:
But then the question is if we do it for record_once why not for other configuration options? That's why I opened the discussion on this feature.

PR #190, opened 2015-08-21
Add default for record_mode from environment variable

Diaoul:
See #189

### DIFF
--- a/vcr/config.py
+++ b/vcr/config.py
@@ -25,7 +25,7 @@ class VCR(object):
 
     def __init__(self, path_transformer=lambda x: x, before_record_request=None,
                  custom_patches=(), filter_query_parameters=(), ignore_hosts=(),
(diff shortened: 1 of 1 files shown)


Diaoul:
This is untested but I believe it should work.

IvanMalison:
I have a very weak preference for not including this just because it is SO easy to set up yourself. I guess it doesn't really hurt. @kevin1024 I'll leave this one up to you

kevin1024:
Yeah, I am against merging this just because it is limited to one variable. 

I'm not against trying to add an envvar-based configuration scheme, but we would need a good plan for that that (some kind of naming scheme?  A way to make it more explicit?)

I think adding just this one case would be confusing (why can I only change this one piece of config from the environment and not anything else?)

Closing for now, but I am open to discussion about building an envvar config scheme.  In the meantime, you can use the technique I showed in #189.

PR #192, opened 2015-08-24
Make request.headers always a CaseInsensitiveDict.

agriffis:
Previously request.headers was a normal dict (albeit with the
request.add_header interface) which meant that some code paths would do
case-sensitive matching, for example remove_post_data_parameters which
tests for 'Content-Type'. This change allows all code paths to get the same
case-insensitive treatment.

Additionally request.headers becomes a property to enforce upgrading it to
a CaseInsensitiveDict even if assigned.

### DIFF
--- a/vcr/request.py
+++ b/vcr/request.py
@@ -1,10 +1,11 @@
 from six import BytesIO, text_type
 from six.moves.urllib.parse import urlparse, parse_qsl
+from .util import CaseInsensitiveDict
 
 
 class Request(object):
     """
-    VCR's  representation of a request.
+    VCR's representation of a request.
 
     There is a weird quirk in HTTP.  You can send the same header twice.  For
     this reason, headers are represented by a dict, with lists as the values.
@@ -32,9 +33,19 @@ class Request(object):
             self.body = body.read()
         else:
             self.body = body
-        self.headers = {}
-        for key in headers:
-            self.add_header(key, headers[key])
+        self.headers = CaseInsensitiveDict()
+        for key, value in headers.items():
+            self.add_header(key, value)
+
+    @property
+    def headers(self):
+        return self._headers
+
+    @headers.setter
+    def headers(self, value):
+        if not isinstance(value, CaseInsensitiveDict):
+            value = CaseInsensitiveDict(value)
+        self._headers = value
 
     @property
     def body(self):
--- a/vcr/request.py
+++ b/vcr/request.py
@@ -1,3 +1,4 @@
+import warnings
 from six import BytesIO, text_type
 from six.moves.urllib.parse import urlparse, parse_qsl
 from .util import CaseInsensitiveDict
@@ -6,23 +7,6 @@ from .util import CaseInsensitiveDict
 class Request(object):
     """
     VCR's representation of a request.
-
-    There is a weird quirk in HTTP.  You can send the same header twice.  For
-    this reason, headers are represented by a dict, with lists as the values.
-    However, it appears that HTTPlib is completely incapable of sending the
-    same header twice.  This puts me in a weird position: I want to be able to
-    accurately represent HTTP headers in cassettes, but I don't want the extra
-    step of always having to do [0] in the general case, i.e.
-    request.headers['key'][0]
-
-    In addition, some servers sometimes send the same header more than once,
-    and httplib *can* deal with this situation.
-
-    Futhermore, I wanted to keep the request and response cassette format as
-    similar as possible.
-
-    For this reason, in cassettes I keep a dict with lists as keys, but once
-    deserialized into VCR, I keep them as plain, naked dicts.
     """
 
     def __init__(self, method, uri, body, headers):
@@ -33,9 +17,7 @@ class Request(object):
             self.body = body.read()
         else:
             self.body = body
-        self.headers = CaseInsensitiveDict()
-        for key, value in headers.items():
-            self.add_header(key, value)
+        self.headers = headers
 
     @property
     def headers(self):
@@ -43,8 +25,8 @@ class Request(object):
 
     @headers.setter
     def headers(self, value):
-        if not isinstance(value, CaseInsensitiveDict):
-            value = CaseInsensitiveDict(value)
+        if not isinstance(value, HeadersDict):
+            value = HeadersDict(value)
         self._headers = value
 
     @property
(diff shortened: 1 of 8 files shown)


agriffis:
Yowsers. Tests ran here. Hang on while I investigate.

agriffis:
Phewph, that's better.

IvanMalison:
This is what I always wanted all along. I believe that last time we discussed this there was some objection, but I may be remembering this incorrectly. Let me see if I can find the discussion

IvanMalison:
Yep @kevin1024 seems to be weakly against this idea in https://github.com/kevin1024/vcrpy/pull/167#issuecomment-119098603

Its really only mentioned in passing there, and I think that there was another discussion somewhere else, but my memory is failing me there.

@kevin1024 I am actually in favor of merging this, but I sort of understand your objection so I'll let you make the call.

agriffis:
Interesting re the prior discussion, thanks! I'm not quite sure I understand the objection, though. Using `CaseInsensitiveDict` shouldn't cause any unwanted header case normalization, because it preserves the original assignment case:
```python
>>> from vcr.util import CaseInsensitiveDict
>>> d = CaseInsensitiveDict([('Content-Type', 'text/html')])
>>> 'content-type' in d
True
>>> d
{'Content-Type': 'application/json'}
```

vcr/request.py
```diff
@@ -32,9 +33,19 @@ def __init__(self, method, uri, body, headers):
             self.body = body.read()
         else:
             self.body = body
-        self.headers = {}
-        for key in headers:
-            self.add_header(key, headers[key])
+        self.headers = CaseInsensitiveDict()
+        for key, value in headers.items():
```
IvanMalison: Use .update?

vcr/request.py
```diff
@@ -32,9 +33,19 @@ def __init__(self, method, uri, body, headers):
             self.body = body.read()
         else:
             self.body = body
-        self.headers = {}
-        for key in headers:
-            self.add_header(key, headers[key])
+        self.headers = CaseInsensitiveDict()
```
IvanMalison: or better yet, actually: doesn't your setter actually make it so you don't need a setter here at all?

i.e. just leave `self.headers = headers` and everything should just work

agriffis: The reason to call `.add_header()` is to preserve the behavior of extracting the first element from a list of header values. Re-examining that behavior might be something for a future excursion but it seems orthogonal to this PR. :-)

IvanMalison:
@agriffis cool. I hadn't looked at the code closely and it seems that case is preserved. I guess the only objection then would be that the behavior WOULD (potentially) be changed when you try to send along different headers for lower and upper case values. CaseInsensitiveDict would collapse it before it is passed to the real http connection, whereas the current implementation wouldn't.

vcr/request.py
IvanMalison: I honestly think the current behavior of `add_header` is pretty weird, but given that this the case, the setter SHOULD also do the same thing, right?

In general, I think its better to have logic for populating values like that live in exactly one place, and I think that the setter is the right place for that. The behavior should be abstracted into the dict object in any case.

IvanMalison: > seems orthogonal to this PR. :-)

Not sure I completely agree. This is a significant (at least minor version bump) change to the behavior of headers. I'd rather not have to do ANOTHER minor version bump for more breaking changes to the way that headers behave.

I suppose that it could be in a separate pr before the minor version bump, but I think that now seems like a good time to handle this to me.

kevin1024:
Yeah, IIRC my main concern is covering up behavior on clients / servers that *are* case-sensitive, despite the RFC saying that headers are not case-sensitive.  I haven't looked closely enough at this change to see if it could cause a problem with that.

vcr/request.py
agriffis: I'd be happy to kill `add_header` (in fact I'd started to do that before I reined in the changes) but in that case it's not just the `headers` setter that needs changing, rather `headers` itself needs to have different behavior.

It would be like this (untested, just for discussion):
```python
class HeadersDict(CaseInsensitiveDict):
    def __setitem__(key, value):
        if isinstance(value, (tuple, list)):
            value = value[0]
        super(HeadersDict, self).__setitem__(key, value)
```

and then I'd move the giant docstring from `VCR` down to `HeadersDict`

IvanMalison:
> Yeah, IIRC my main concern is covering up behavior on clients / servers that are case-sensitive, despite the RFC saying that headers are not case-sensitive. I haven't looked closely enough at this change to see if it could cause a problem with that.

The only way something could be influenced by this change is if someone were to say send along both
`'Content-Type'` and `'content-type'`, but the server only accepted one of them (or even more crazily accepted both). Other than that I don't think it would have an effect.

Sending along `'content-type'` vs `'Content-Type'` would still work, as long as they were sent alone.

kevin1024:
OK, well in that case let's go for it if you are +1 and it seems to clean up the code.  I don't think we need to support a server *that* broken.  Probably want a version bump since it could potentially be a breaking change.  I'm pretty much +0 on this.

agriffis:
@kevin1024 cool, thanks.

@IvanMalison Hang on while I do the `HeadersDict` change

IvanMalison:
@agriffis Thanks! If you throw the minor version bump in there (and CHANGELOG edits) I'd be happy to release to pypi for you.

agriffis:
@IvanMalison I didn't do the changelog or version bump because I have another PR in progress. Maybe after that one. ;-)

agriffis:
@IvanMalison This should be all set now. Subclassing `CaseInsensitiveDict` to `HeadersDict` enabled a minorly beneficial behavior change: the case of the original key is now sticky. See the commit message and tests in 7312229 for more info.

However I didn't update the changelog or version bump because I have #196 in the works. I imagine that you'd like all these changes at once before the version bump, but please let me know. Thanks!

IvanMalison:
I have a few tweaks to HeadersDict that I'm working on (I want it to actually store multiple values for each header even if it doesn't send them along) that I want to finish before I merge this. I'll try to finish them up by the end of this weekend.

agriffis:
As I understand it, the current (previous) implementation was so that users of `Request` (mainly filters) would have easy access to single-value headers rather than always needing to deal with lists. In what context are you thinking that storing a list would be appropriate? When calling from `putheader`? Wouldn't we still want filters to overwrite headers with a single value rather than appending to a list? Just curious how you're thinking this will work.

By the way I'm on freenode as agriffis in case you'd like to discuss realtime, or you could make a new chat for this project on gitter.im (I can't because I'm not a contributor)

kevin1024:
> (I can't because I'm not a contributor)

@agriffis you are now!  Thanks for your help on the project.

PR #203, opened 2015-08-28
Insensitive headers

IvanMalison:
WIP. letting people take a look to see if I should abandon or not

### DIFF
--- a/tests/unit/test_request.py
+++ b/tests/unit/test_request.py
@@ -1,6 +1,7 @@
 import pytest
 
-from vcr.request import Request, HeadersDict
+from vcr.request import Request
+from vcr.util import HeadersDict
 
 
 def test_str():
@@ -29,7 +30,7 @@ def test_add_header_deprecated():
     ('https://go.com/', 443),
     ('https://go.com:443/', 443),
     ('https://go.com:3000/', 3000),
-    ])
+])
 def test_port(uri, expected_port):
     req = Request('GET', uri,  '', {})
     assert req.port == expected_port
--- a/tox.ini
+++ b/tox.ini
@@ -15,6 +15,7 @@ deps =
     pytest
     pytest-localserver
     PyYAML
+    ipdb
     requests1: requests==1.2.3
     requests27: requests==2.7.0
     requests26: requests==2.6.0
--- a/vcr/request.py
+++ b/vcr/request.py
@@ -1,13 +1,11 @@
 import warnings
 from six import BytesIO, text_type
 from six.moves.urllib.parse import urlparse, parse_qsl
-from .util import CaseInsensitiveDict
+from .util import HeadersDict
 
 
 class Request(object):
-    """
-    VCR's representation of a request.
-    """
+    """Represent an HTTP request in vcrpy."""
 
     def __init__(self, method, uri, body, headers):
         self.method = method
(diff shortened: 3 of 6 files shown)


agriffis:
Hey @IvanMalison 

Lots of thoughts here. Let's talk about design, then purpose, then design...

Regarding design:
  * `DictDecorator` tries to implement a proxy, but it's broken because it doesn't call back to the MRO leaf as proper Python inheritance would. See https://code.activestate.com/recipes/519639-true-lieberman-style-delegation-in-python/
  * `DictDecorator` implements too many methods. The raison d'être of `MutableMapping` is that it's hard to inherit from the built-in `dict` because the implementation takes shortcuts internally rather than calling its own methods, so `MutableMapping` provides an implementation where the subclass only needs to implement a small set of core methods. See https://hg.python.org/cpython/file/tip/Lib/_abcoll.py#l493

Regarding purpose:
  * I'm definitely on-board with making VCR.py properly preserve multiple header lines, both in the real request and the recorded cassette. It's probably rare in practice, but it's [supposed to mean the same thing as a comma-separated list](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)
  * While we're making this effort, we should also preserve the order of all the headers. It's not supposed to matter (ref above) but for the sake of transparency it's worthwhile.
  * We probably want separate matchers, `headers` which compares a dict (with duplicate headers comma-joined!) and `headers_strict` which compares the fully ordered headers.
  * Filters still want a simplified case-insensitive dict-based view of headers with single values, probably exactly like the `headers` matcher above, including comma-joining. (I will also want this to be mutable for #196)

Back to design:
  * It seems like the underlying storage should actually be a list of KV tuples.
  * Proxies (delegation) are most useful when you need to wrap an instance of something, especially to change the behavior when you have no opportunity to specify a subclass. I'm not sure they're the right answer here, even if implemented properly. It probably makes more sense to implement a `MutableMapping` over the top of the KV list, then have a mixin to implement case normalization, and finally instantiate `HeadersDict` inheriting from both, something like this (off the top of my head, completely untested):

```python
class ConcatDict(collections.MutableMapping):

    def __init__(self, kvlist):
        self._kvlist = kvlist

    def __getitem__(self, key):
        values = [v for k, v in self._kvlist if k == key]
        if not matches:
            raise KeyError(key)
        return ','.join(values)

    def __setitem__(self, key, value):
        try:
            i = self._kvlist.find(key)
        except ValueError:
            self._kvlist.append((key, value))
        else:
            del self._kvlist[key]
            self._kvlist.insert(i, (key, value))

    def __delitem__(self, key, value):
        self._kvlist[:] = [(k, v) for k, v in self._kvlist if k != key]

    def __iter__(self, key, value):
        return (k for k, v in self._kvlist)

    def __len__(self):
        return len(self._kvlist)


class CaseInsensitiveMixin(object):

    def __getitem__(self, key):
        try:
            k, v = super(CaseInsensitiveMixin, self).__getitem__(key.lower())
        except KeyError:
            raise KeyError(key)
        else:
            return v

    def __setitem__(self, key, value):
        super(CaseInsensitiveMixin, self).__setitem__(key.lower(), (k, v))

    def __delitem__(self, key):
        super(CaseInsensitiveMixin, self).__delitem__(key.lower())


class HeadersDict(CaseInsensitiveMixin,
                  ConcatDict):
    pass


class Request(object):

    def __init__(self):
        self._headers = []

    @property
    def headers(self):
        return HeadersDict(self._headers)

```

Final notes:
  * This is going to change the cassette format, necessitating a major version bump. We'll also want either a converter for old cassette files, or to read them transparently. This is begging for versioned cassette files (unless they're already versioned; I haven't checked).
  * I'd really like to get #196 merged before fully tackling this. That PR is finished except for tweaking docs, and I need it for vcrpy-facebook. It doesn't break compatibility and it would make sense to push out a release after we get it in.

IvanMalison:
> Hey @IvanMalison
> 
> Lots of thoughts here. Let's talk about design, then purpose, then design...
> 
> Regarding design:
> 
> DictDecorator tries to implement a proxy, but it's broken because it doesn't call back to the MRO leaf as proper Python inheritance would. See https://code.activestate.com/recipes/519639-true-lieberman-style-delegation-in-python/

I don't think that this is an implementation of a proxy, but the decorator pattern, and I would argue that the behavior that you are describing here is NOT desired in the case of the decorator pattern

> DictDecorator implements too many methods. The raison d'être of MutableMapping is that it's hard to inherit from the built-in dict because the implementation takes shortcuts internally rather than calling its own methods, so MutableMapping provides an implementation where the subclass only needs to implement a small set of core methods. See https://hg.python.org/cpython/file/tip/Lib/_abcoll.py#l493

uhhh what? Which methods are not part of this interface?

> Regarding purpose:
> 
> I'm definitely on-board with making VCR.py properly preserve multiple header lines, both in the real request and the recorded cassette. It's probably rare in practice, but it's supposed to mean the same thing as a comma-separated list
> While we're making this effort, we should also preserve the order of all the headers. It's not supposed to matter (ref above) but for the sake of transparency it's worthwhile.
> We probably want separate matchers, headers which compares a dict (with duplicate headers comma-joined!) and headers_strict which compares the fully ordered headers.
> Filters still want a simplified case-insensitive dict-based view of headers with single values, probably exactly like the headers matcher above, including comma-joining. (I will also want this to be mutable for #196)
> Back to design:
> 
> It seems like the underlying storage should actually be a list of KV tuples.
> Proxies (delegation) are most useful when you need to wrap an instance of something, especially to change the behavior when you have no opportunity to specify a subclass. I'm not sure they're the right answer here, even if implemented properly. It probably makes more sense to implement a MutableMapping over the top of the KV list, then have a mixin to implement case normalization, and finally instantiate HeadersDict inheriting from both, something like this (off the top of my head, completely untested):

Mixins are very close to constituting an anti-pattern in my mind. I don't think that there are many legitimate use cases for them.

With that said, I don't think that what I did here is particularly useful/the right way to do now that I'm looking at it again. We could probably get away with doing something very straightforward (putting everything in one class that handles all of the behavior), and that is probably the best way to proceed. The lover of abstraction in me hates that answer...

Issue #130, opened 2014-12-23
Problems with Python 2.7.9 and urllib2


vanclevstik:
It seems like the release of Python 2.7.9 brought some changes to urllib2 that are incompatible with vcrpy.

Running tests with python 2.7.9 throws similar error on most of the `test_urllib2` tests:

```
TypeError: unbound method connect() must be called with 
VCRHTTPConnection/tmp/pytest-4/test_response_headers_https_0/headers.yaml instance as first argument (got HTTPSConnection instance instead)
```

Re-running Travis build now that Travis has upgraded to 2.7.9 would probably finish with failed build.

PR #213, opened 2015-09-23
Avoid force_reset to make patching thread-safe

### LINKED ISSUE #212: force_reset makes vcr non thread-safe
I tried to use vcr to write tests for an old program that uses threadpools internally. It turned out that this doesn't work and I believe it's due to the use of `force_reset` when the unpatched base classes are instantiated. The reset is necessary since `httplib.HTTPSConnection` refers to `httplib.HTTPConnection` by name.

The following program shows unexpected errors on my machine:

``` python
import sys, urllib2

from concurrent import futures
from vcr import use_cassette
from vcr.errors import CannotOverwriteExistingCassetteException

@use_cassette('not-actually-used.yaml', record_mode='none')
def load_all(*urls):
    print 'Submitting URLs...',
    with futures.ThreadPoolExecutor(max_workers=2) as executor:
        fs = [executor.submit(urllib2.urlopen, url) for url in urls]
        print 'done!'
        for f in futures.as_completed(fs):
            try:
                result = f.result()
                print 'Unexpected result:', result
                sys.exit(1)
            except CannotOverwriteExistingCassetteException as e:
                pass  # Success!
            except Exception as e:
                print 'Unexpected error:', e
                sys.exit(1)


if __name__ == '__main__':
    load_all('https://httpbin.org/get?q=testing+vcrpy+with',
             'https://httpbin.org/get?q=https+and+threads')
```

The error shown is

```
unbound method __init__() must be called with VCRHTTPConnectionnot-actually-used.yaml
instance as first argument (got HTTPSConnection instance instead)
```

and is triggered when the unpatched `httplib.HTTPSConnection.__init__` calls the patched `httplib.HTTPConnection.__init__`.

I see the error about 50% of the time, so I normally run the script in a little loop (Zsh syntax):

``` sh
errors=0; for i in {1..20}; do python test.py; ((errors += $?)); echo; echo "*** Successes: $((i - errors))/$i, $((100 * (i - errors) / i))%"; done
```

mgeisler:
Before, `force_reset` was used twice when instantiating the original baseclasses. It was necessary to unpatch things, since the classes were subclasses of patched classes *and* because the subclass referred to the super class explicitly by name.

That is, `httplib` has code like:
```python
class HTTPConnection:
    ...

class HTTPSConnection(HTTPConnection):
    "This class allows communication via SSL."

    def __init__(self, host, port=None, ...):
        HTTPConnection.__init__(self, host, port, ...)
```
Creating a real `HTTPSConnection` instance is now impossible as long as `httplib.HTTPConnection` is patched.

To get around this, `force_reset` was used to temporary unpatch things. However, this means modifying global state and is thus not thread-safe. Modifying the classes once when vcr is loaded is better since no further patching and unpatching needs to be done after that.

This code in this commit uses byteplay, which is not Python 3 compatible. I think byteplay could be upsed with Python 3 compatibility, but I haven't looked much into it. The project also seems semi-abandoned with the last release done in 2010.

Also, I will agree that this kind of bytecode level patching is at best controversial :-) However, since I only use it to fix a reference to a super-class, I think it could be acceptable here.

Fixes #212 and makes the test program there run stable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ class PyTest(TestCommand):
         sys.exit(errno)
 
 
-install_requires = ['PyYAML', 'wrapt', 'six>=1.5']
+install_requires = ['PyYAML', 'wrapt', 'six>=1.5', 'byteplay']
 
 
 extras_require = {
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -5,12 +5,35 @@ import itertools
 from .compat import contextlib, mock
 from .stubs import VCRHTTPConnection, VCRHTTPSConnection
 from six.moves import http_client as httplib
+from byteplay import LOAD_GLOBAL, LOAD_ATTR, Code
+
+
+def _patch_name(func, old, new):
+    """Replace references to old with new in func bytecode
+
+    This effectively renames variables and attributes in the code for
+    the function passed.
+    """
+    bytecode = Code.from_code(func.func_code)
+    for i, (opcode, arg) in enumerate(bytecode.code):
+        if opcode in [LOAD_GLOBAL, LOAD_ATTR] and arg == old:
+            bytecode.code[i] = (opcode, new)
+    func.func_code = bytecode.to_code()
 
 
 # Save some of the original types for the purposes of unpatching
 _HTTPConnection = httplib.HTTPConnection
 _HTTPSConnection = httplib.HTTPSConnection
 
+httplib._HTTPConnection = _HTTPConnection
+httplib._HTTPSConnection = _HTTPSConnection
+# Make the original HTTPSConnection class refer to the original
+# HTTPConnection class, always. This means that we don't need to
+# unpatch things when creating HTTPSConnection instances.
+_patch_name(_HTTPSConnection.__init__.im_func,
+            'HTTPConnection', '_HTTPConnection')
+_patch_name(_HTTPSConnection.connect.im_func,
+            'HTTPConnection', '_HTTPConnection')
 
 # Try to save the original types for requests
 try:
@@ -42,6 +65,12 @@ else:
     _HTTPSConnectionWithTimeout = httplib2.HTTPSConnectionWithTimeout
     _SCHEME_TO_CONNECTION = httplib2.SCHEME_TO_CONNECTION
 
+    # Make the original classes refer to the original httplib classes.
+    _patch_name(_HTTPConnectionWithTimeout.__init__.im_func,
+                'HTTPConnection', '_HTTPConnection')
+    _patch_name(_HTTPSConnectionWithTimeout.__init__.im_func,
+                'HTTPSConnection', '_HTTPSConnection')
+
 
 # Try to save the original types for boto
 try:
@@ -50,6 +79,9 @@ except ImportError:  # pragma: no cover
     pass
 else:
     _CertValidatingHTTPSConnection = boto.https_connection.CertValidatingHTTPSConnection
+    # Make the original class refer to the original httplib class.
+    _patch_name(_CertValidatingHTTPSConnection.__init__.im_func,
+                'HTTPConnection', '_HTTPConnection')
 
 
 # Try to save the original types for Tornado
--- a/vcr/stubs/__init__.py
+++ b/vcr/stubs/__init__.py
@@ -244,16 +244,12 @@ class VCRConnection(object):
                     self._vcr_request
                 )
             )
-            # This is imported here to avoid circular import.
-            # TODO(@IvanMalison): Refactor to allow normal import.
-            from vcr.patch import force_reset
-            with force_reset():
-                self.real_connection.request(
-                    method=self._vcr_request.method,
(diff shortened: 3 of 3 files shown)


kevin1024:
Wow, this is pretty mindblowing.  I have to admit I haven't thought much about making VCR.py threadsafe.  Is this all that is needed for it to work?

I'm obviously not too excited about adding another dependency / doing bytecode-level patching.   The goal is noble though.

IvanMalison:
> Before, force_reset was used twice when instantiating the original baseclasses. It was necessary to unpatch things, since the classes were subclasses of patched classes and because the subclass referred to the super class explicitly by name.

Sigh. This is why you are supposed to use super.

> This code in this commit uses byteplay, which is not Python 3 compatible. I think byteplay could be upsed with Python 3 compatibility, but I haven't looked much into it. The project also seems semi-abandoned with the last release done in 2010.

That would be a dealbreaker for its inclusion in this library.

We could do something sort of hacky where we just patch the function with an exact rewrite of the existing function except with real super calls instead of explicit superclass calls (instead of the force reset).

Is this really the only reason we do a force reset here? I feel like I vaguely remember some other reason.

> Wow, this is pretty mindblowing. I have to admit I haven't thought much about making VCR.py threadsafe. Is this all that is needed for it to work?

This is only threadsafety assuming that there is exactly one active cassette at any given moment, but i do believe it would give us threadsafety in thise case.

kevin1024:
> Sigh. This is why you are supposed to use super.

Even better: they are old-style classes so they actually can't use super.

mgeisler:
@kevin1024 wrote:
> Wow, this is pretty mindblowing.

Thanks, that's a much better response than I had hoped for :-)

@IvanMalison wrote:
> >This code in this commit uses byteplay, which is not Python 3 compatible. I think byteplay could be updated with Python 3 compatibility, but I haven't looked much into it. The project also seems semi-abandoned with the last release done in 2010.
>
> That would be a dealbreaker for its inclusion in this library.

Yeah, I would call the patch a proof of concept at best. I plan to rewrite it as a standalone piece of code that uses the `custom_patches` argument for a project at work. That way people who use Python 2.7 can take it and play with it without you having to actually merge it into vcr.py directly.

I think making vcr.py threadsafe would be a good and reasonable thing. After all, making parallel HTTP requests is the prime use case of packages like `concurrent.futures` :-)

> We could do something sort of hacky where we just patch the function with an exact rewrite of the existing function except with real super calls instead of explicit superclass calls (instead of the force reset).
>
> Is this really the only reason we do a force reset here? I feel like I vaguely remember some other reason.

I think it is the main reason. One `force_reset` was added in e1f65bcbdc8e to fix #130, which I believe is the same issue I was seeing.

IvanMalison:
> Yeah, I would call the patch a proof of concept at best. I plan to rewrite it as a standalone piece of code that uses the custom_patches argument for a project at work. That way people who use Python 2.7 can take it and play with it without you having to actually merge it into vcr.py directly.

Cool, I'd love to take a look at that when you are done

> I think making vcr.py threadsafe would be a good and reasonable thing. After all, making parallel HTTP requests is the prime use case of packages like concurrent.futures :-)

yeah agreed, I just want to make sure we do so in a way that doesn't break any existing features/compatibility

mgeisler:
In the end I simply went with a bit of code that is imported at the top of my tests:
```python
import byteplay

def _patch_name(func, old, new):
    """Replace references to old with new in func bytecode.

    This effectively renames variables and attributes in the code for
    the function passed. When the function used to access 'old', it
    will now instead access 'new'.
    """
    bytecode = byteplay.Code.from_code(func.func_code)
    wanted_opcodes = [byteplay.LOAD_GLOBAL, byteplay.LOAD_ATTR]
    for i, (opcode, arg) in enumerate(bytecode.code):
        if opcode in wanted_opcodes and arg == old:
            bytecode.code[i] = (opcode, new)
    func.func_code = bytecode.to_code()

# Make the original HTTPSConnection class refer to the original
# HTTPConnection class, always. This avoids "unbound method __init__()
# must be called with VCRHTTPConnection<something>.yaml instance as
# first argument (got HTTPSConnection instance instead)" errors.
#
# Please see https://github.com/kevin1024/vcrpy/issues/212 for more
# background and discussion.
httplib._HTTPConnection = httplib.HTTPConnection
httplib._HTTPSConnection = httplib.HTTPSConnection
_patch_name(httplib._HTTPSConnection.__init__.im_func,
            'HTTPConnection', '_HTTPConnection')
_patch_name(httplib._HTTPSConnection.connect.im_func,
            'HTTPConnection', '_HTTPConnection')
```

Since the code is safe to run always (it doesn't change the behavior when no patching is done), I decided that was the easiest way.

The `force_reset` done by vcrpy no longer matter when `__init__` and `connect` is patched like above, so the part of my original pull request that removed the `force_reset` calls can be left in.

MSeal:
FYI there is now a python3 library for byteplay:
https://pypi.python.org/pypi/byteplay3/3.5.0

PR #217, opened 2015-10-02
Fix use_cassette when used as decorator does not return the wrapped function's return value

bcen:
`CassetteContextDecorator._handle_function` should return the value back to caller.

Let me know if the test is good enough or not, thanks.


kevin1024:
Cool, :+1: from me.  Paging @IvanMalison  - you know more about the implementation around this.

IvanMalison:
Ahh my bad. Can't believe I missed this when I wrote it. Thanks!

PR #229, opened 2015-11-25
Fix pyflakes and pep8 errors

jayvdb:
Use extra asserts to use previously unused variables in tests,
such as `cass` and `response`.

Fix only pyflakes errors in docs/conf.py

### DIFF
--- a/tests/integration/test_record_mode.py
+++ b/tests/integration/test_record_mode.py
@@ -1,4 +1,3 @@
-import os
 import pytest
 import vcr
 from six.moves.urllib.request import urlopen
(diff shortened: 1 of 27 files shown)


jayvdb:
fwiw, the `.travis.yml` changes will be mostly unnecessary after #228

kevin1024:
Looks like I caused some conflicts when merging #228.  Can you rebase?

Also, I noticed quite a few new asserts.  Is this purely for pep8 reasons?  Does flake8 require an assert per test or something?

jayvdb:
pyflakes requires that all variables are used.
Often the tests had `with blah as cass: something` which did not access `cass`.  Adding an assert for `cass` fixes the problem.  The alternative is to remove `.. as cass:`, but I guessed it is more useful to add asserts than remove `cass` which may be useful for debugging or future asserts.

kevin1024:
Ohh gotcha.

I think I would prefer to do the first option (remove ..as cass) since I prefer to have fewer asserts per test.  Asserts without much meaning to the test can dillute the meaning of the important assert.  I strive for a single assert per test but often don't achieve it :)

tests/integration/test_record_mode.py
```diff
@@ -9,16 +8,21 @@ def test_once_record_mode(tmpdir):
     with vcr.use_cassette(testfile, record_mode="once"):
         # cassette file doesn't exist, so create.
         response = urlopen('http://httpbin.org/').read()
```
jayvdb: should I also change lines like this to `urlopen('http://httpbin.org/').read()` , and remove the `assert response` ?

(It is very late here, so I'll pick this up tomorrow and revise it to better suit your tastes.)

kevin1024: Yeah, if we're not doing anything with the variable I suppose we don't need to create it.  (BTW thanks a lot for doing this!  Having all the code be pep8 gives a nice clean feeling :) :+1:

kevin1024:
Thanks for the cleanup!

Issue #222, opened 2015-10-13
Serialization errors with JSON adapter and Python3.4


aliaksandrb:
Hi!

Assume we have `vcr` configured with `json` adapter:

``` python
import vcr
vcr = vcr.VCR(serializer='json')
```

and a simple method posting some data:

``` python

def login(username, password):
    credentials = {
        'username': username,
        'password': password,
    }

    response_obj = requests.post('https://example.com/login', json=credentials)
    return response_obj
```

Then the wrapped test method fails while trying to dump the request-response into json file:

``` python
    def test_login(self):
        with vcr.use_cassette('fixtures/vcr_cassettes/invalid_login.json'):
            with self.assertRaises(client.Unauthorized):
                self.client.log_in(username='***', password='***')
```

Error outputs is reduced:

```
  File "~/.tox/py34/lib/python3.4/site-packages/vcr/serializers/jsonserializer.py", line 13, in serialize
    return json.dumps(cassette_dict, indent=4)
  File "/usr/lib/python3.4/json/__init__.py", line 237, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.4/json/encoder.py", line 194, in encode
    chunks = list(chunks)
  File "/usr/lib/python3.4/json/encoder.py", line 422, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.4/json/encoder.py", line 396, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.4/json/encoder.py", line 317, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.4/json/encoder.py", line 396, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.4/json/encoder.py", line 396, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.4/json/encoder.py", line 429, in _iterencode
    o = _default(o)
  File "/usr/lib/python3.4/json/encoder.py", line 173, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: b'{"username": "***", "password": "***"}' is not JSON serializable
```

As I understand the problem occurs because `vcr` doesn't try to encode `request` body a counterweight to `response`.
And because in Python2.7 `body` is simple `str` type `json.dump` works correctly with it but fails on `byte` type of Python3.4.

[...]

PR #232, opened 2015-12-02
Fix for Serialization errors with JSON adapter

aliaksandrb:
Hi!

Here is first attemp to fix the issue #222.
It might be not perfect (specially the `jsonserializer.py` part :flushed:) but if general idea seems reasonable I could update it according to comments.

So the idea itself to try to serialize `request body` as well as `response body` because in`py3` its string representation could be binary that leads to an error described in issue #222.

See you!

### DIFF
--- a/tests/unit/test_serialize.py
+++ b/tests/unit/test_serialize.py
@@ -2,7 +2,8 @@
 import pytest
 
 from vcr.compat import mock
-from vcr.serialize import deserialize
+from vcr.request import Request
+from vcr.serialize import deserialize, serialize
 from vcr.serializers import yamlserializer, jsonserializer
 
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_serialize.py
```diff
@@ -83,3 +84,47 @@ def test_deserialize_py2py3_yaml_cassette(tmpdir, req_body, expect):
 def test_serialize_constructs_UnicodeDecodeError(mock_dumps):
     with pytest.raises(UnicodeDecodeError):
         jsonserializer.serialize({})
+
```
IvanMalison: two newlines between top level definitions per pep8

IvanMalison:
@kevin1024 can you take a look at this. I never really worked with the json serializer so I don't have much context.

kevin1024:
Check out the code here: https://github.com/kevin1024/vcrpy/blob/master/vcr/serializers/compat.py

Probably just need to do the same thing for the request body.  Which is pretty much what you did, but maybe move it into serializers.compat and try to reuse the code in there if possible.

Also this comment: https://github.com/kevin1024/vcrpy/blob/master/vcr/serializers/compat.py#L16  says "request" but I think should say "response".  Confusing :)

aliaksandrb:
Thanks for response guys!

I have checked `compact..py` the only problem is its usage of `resp['body']['string']` rather than simple `resp['body']`.

Anyway, I will try to refactor this patch a bit according to your notes.

gabrielpugliese-luizalabs:
Hi,

I do need this feature and I'm kinda lost here. How could I help you integrating this with master?

kevin1024:
This patch is pretty good already, just needs a small refactor.  Instead of adding the `_serialize_request_body(request)` to the json serializer, see if you can reuse the methods in the serizalizers/compat.py file.

kevin1024:
If not that's ok too, just want someone to try :)

gabrielpugliese-luizalabs:
I'm using your fork already :) Working great.

aliaksandrb:
Totally forgot about it, shame for me :disappointed: 

I have rebased the patch to a new version, maybe it will look better :)

kevin1024:
Rebased and merged in a different branch.  Thanks!!

Issue #230, opened 2015-12-01
Add decode_compressed_response VCR option


tobiowo:
The Ruby version of VCR supports a [decode_compressed_response](https://www.relishapp.com/vcr/vcr/v/2-9-3/docs/cassettes/decode-compressed-response) option that decompresses response bodies encoded with gzip or deflate before recording and makes cassettes human-readable/editable. This could be implemented in vcrpy as a before_record_response function using zlib.

IvanMalison:
seems like a good idea pr?

kevin1024:
:+1: sounds useful to me. Would merge a pull request for this.

agriffis:
I have this in vcrpy-facebook if you'd like to pull from that code. https://github.com/agriffis/vcrpy-facebook/blob/master/vcr_facebook/response.py#L50

PR #149, opened 2015-04-12
WIP: add pytest-httpbin

kevin1024:
This adds support to most of the tests that rely on httpbin.  There are still a few that need to hit the internet.  Not sure if I can make the entire test suite run offline or not.  Might need to mock DNS for this to work.

PR #234, opened 2015-12-03
Decode compressed response option

tobiowo:
Add decode_compressed_response VCR option #230 

httpbin tests might conflict with #149

### DIFF
--- a/vcr/config.py
+++ b/vcr/config.py
@@ -35,7 +35,8 @@ class VCR(object):
                  before_record_response=None, filter_post_data_parameters=(),
                  match_on=('method', 'scheme', 'host', 'port', 'path', 'query'),
                  before_record=None, inject_cassette=False, serializer='yaml',
-                 cassette_library_dir=None, func_path_generator=None):
+                 cassette_library_dir=None, func_path_generator=None,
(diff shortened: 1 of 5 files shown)


vcr/config.py
```diff
@@ -163,7 +165,12 @@ def _build_before_record_response(self, options):
         before_record_response = options.get(
             'before_record_response', self.before_record_response
         )
+        decode_compressed_response = options.get(
+            'decode_compressed_response', self.decode_compressed_response
+        )
         filter_functions = []
+        if decode_compressed_response:
+            filter_functions.append(filters.decode_response)
```
IvanMalison: Nit: I would put this filter BEFORE any of the explicitly passed in filters, so that those filters could access the decoded values if they needed to.

IvanMalison:
Great work and nice tests! Just make that one small change that I suggested and I'll merge.

vcr/config.py
tobiowo: That's the way it works now: decode_response (if the option is set) is added as the first filter, then everything passed in from `before_record_response` is appended to the end of the filter_functions list.

kevin1024:
Sweet!  Thanks!

PR #236, opened 2015-12-08
Run flake8 on all environments

jayvdb:
dc9cd422 introduced Travis checking of flakes using tox.
However tox.ini only defined 'flakes-py34', so Travis
was only invoking flake8 on Python 3.4, and invoking
py.test on other Python versions.

As tox factors only work properly on the default `testenv`,
use a generic tox environment 'flakes' which uses the
default Python version, which Travis will set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,4 +42,5 @@ install:
 script:
 - v=$TRAVIS_PYTHON_VERSION
 - if [[ ${v%py} != $v ]]; then TOXPY=$v; else TOXPY=py"${v/./}"; fi
-- tox -e "${TOXPY}-${TOX_SUFFIX}"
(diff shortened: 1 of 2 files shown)


kevin1024:
Thanks for the PR.  Looks like there are some tests failing?

jayvdb:
Looks like a bug in the latest pytest <https://pypi.python.org/pypi/pytest>

jayvdb:
Looks like the problem is https://github.com/pytest-dev/pytest/issues/1238 , which has a large impact and a tested PR so a new release should be out soon.

nicoddemus:
`2.8.5` is out now. :smile:

jayvdb:
I've found tox-travis, which does a lot of voodoo for us. So, I prefer #238 , but both do the trick, so over to you to make the decision ;-)

kevin1024:
Closed in favor of #238

PR #237, opened 2015-12-11
Add record_on_exception flag. Addresses #205.

### LINKED ISSUE #205: Don't save requests from decorated tests if decorated test fails
As the title says: Don't save requests from decorated tests if decorated test fails

Does VCR already do this? If so awesome, and you guys should definitely say so in the README somewhere. Otherwise this seems like a solid feature. We have one test in particular where our 3rd party goes down more often than I'd care to admit, and if that bad response gets saved that will be quite aggravating.

dan-passaro:
When `record_on_exception=False`, cassettes will not be written when the decorated/context-managed code raises an uncaught exception. Defaults to `True`, which maintains historical behavior.

Did not know an appropriate place to document this so this PR includes no docs. Would be happy to add them but just couldn't decide what section of the docs was appropriate.

This change is discussed in #205.


neozenith:
@dan-passaro sorry the ball was dropped on this for four years. 

Is this a feature you're still wanting to get included?

If so could you rebase with master to trigger CI? Thanks.

PR #238, opened 2015-12-12
Use tox-travis

jayvdb:
dc9cd42 introduced Travis checking of flakes using tox.
However tox.ini only defined 'flakes-py34', so Travis
was only invoking flake8 on Python 3.4, and invoking
py.test on other Python versions.

As tox factors only work properly on the default testenv,
use a generic tox environment 'flakes', and use tox-travis
to select the correct basepython.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,7 @@ python:
 - 3.4
 - pypy
 install:
-- pip install tox
+- pip install tox-travis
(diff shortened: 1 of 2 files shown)


kevin1024:
Nice find with tox-travis - thanks @ryanhiebert for the library and thanks @jayvdb for the PR(s).

ryanhiebert:
Well coveralls was a little late...

PR #242, opened 2016-01-10
Exclude __pycache__ dirs & compiled files in sdist

koobs:
The current vcrpy sdist on PyPI includes __pycache__ dirs and compiled files which causes tests to fail when running tests via setup.py test

import file mismatch:
imported module 'test_persist' has this __file__ attribute:
  /Users/imalison/Projects/vcrpy/tests/unit/test_persist.py
which is not the same as the test file we want to collect:
  /usr/home/user/repos/freebsd/ports/devel/py-vcrpy/work/vcrpy-1.7.4/tests/unit/test_persist.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules

This change recursively excludes these directories and files from future source distributions

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,5 @@ include README.rst
 include LICENSE.txt
 include tox.ini
 recursive-include tests *
+recursive-exclude * __pycache__
+recursive-exclude * *.py[co]


kevin1024:
Thanks! :cake:

koobs:
Cake is win. You're welcome Kevin :]

Also, it's worth mentioning you have one of the best packaged python .. uh .. packages I've seen in a while (and I've ported *many* to FreeBSD), with all the mod cons:

* All the (correct, and version conditional) *_require{s}
* test command integration
* no forcing of tox for tests because its not actually a *requirement*
* No conflation of *_require{s} and requirements files
* LICENSE in sdist

Good stuff!

kevin1024:
I think @graingert and @IvanMalison figured out all the conditional require mojo - glad to hear it's working!

Does this mean vcrpy is going to be a FreeBSD package?  Woohoo!

PR #239, opened 2015-12-16
Fix a TypeError in concatenation when request.body is empty

osantana:

PR #244, opened 2016-01-19
Avoid concatenating bytes with strings

jaysonsantos:
I stumbled upon an error while using boto and python3.5 while it tries to send chunks of data, vcrpy tries to concatenate string with bytes, because `self._vcr_request.body` is an empty `byte` it will fall on `or` which initialize an empty `str` and try to concatenate with `data` which is a byte  and it would break.

### DIFF
--- a/tests/integration/test_boto.py
+++ b/tests/integration/test_boto.py
@@ -1,12 +1,12 @@
 import pytest
 boto = pytest.importorskip("boto")
 
-import boto
-import boto.iam
-from boto.s3.connection import S3Connection
-from boto.s3.key import Key
-from ConfigParser import DuplicateSectionError
-import vcr
+import boto  # NOQA
+import boto.iam  # NOQA
+from boto.s3.connection import S3Connection  # NOQA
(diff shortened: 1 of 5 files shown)


osantana:
I already tried to fix this issue before: https://github.com/kevin1024/vcrpy/pull/239 :smile:

kevin1024:
happy to merge if you can make the tests pass / add a test that fails without the change!

jaysonsantos:
@kevin1024 the tests that broke (actually flask) were not from my change, but I can fix those anyway and add a regression test (which I didn't because I thought it was already covered).
@osantana, with that appoach on python <= 2.7, it will convert `byte` to `str` if `data` is `str` I don't know if its ok to "force" conversion on what user sent.

jaysonsantos:
Hey @kevin1024 can you check it again? I created a test that can reproduce exactly what I said and I had to put some `# NOQA` on some testes because of pytest skip stuff.

kevin1024:
Not sure how I missed your update before, sorry about that!

Issue #251, opened 2016-05-01
httpbin intermittent 50x error


jayvdb:
On Travis builds I am seeing lots of intermittent errors from httpbin, returning 50x.

Maybe pytest-httpbin has some feature to automatically recover from these intermittent errors, in which case #149 will solve this.

Alternatively, we could use
https://pypi.python.org/pypi/flaky

There are also some tests which I suspect are failing due to 50x , but the test doesnt check the response is OK before inspecting other parts of the response. e.g. `test_requests.py` 's `test_effective_url` as seen here
https://travis-ci.org/jayvdb/vcrpy/jobs/126960307

kevin1024:
Yeah, I'm seeing that too, thus the renewed interest in #149.  Getting kind of bogged down by the parametrized tests in test_matchers though.

jayvdb:
Another case where a 50x is probably hiding behind another problem is `test_stubs.py:test_case_insensitivity`

```
tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_case_insensitivity0')
    def test_case_insensitivity(tmpdir):
        testfile = str(tmpdir.join('case_insensitivity.yml'))
        # check if headers are case insensitive outside of vcrpy
        outside = _headers_are_case_insensitive()
        with vcr.use_cassette(testfile):
            # check if headers are case insensitive inside of vcrpy
            inside = _headers_are_case_insensitive()
            # check if headers are case insensitive after vcrpy deserializes headers
            inside2 = _headers_are_case_insensitive()
    
        # behavior should be the same both inside and outside
>       assert outside == inside == inside2
E       assert False == True
```

https://travis-ci.org/jayvdb/vcrpy/jobs/127038815#L568

jayvdb:
And `test_wild.py:test_cookies`

```
tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_cookies0')
    def test_cookies(tmpdir):
        testfile = str(tmpdir.join('cookies.yml'))
        with vcr.use_cassette(testfile):
            s = requests.Session()
            s.get("http://httpbin.org/cookies/set?k1=v1&k2=v2")
    
            r2 = s.get("http://httpbin.org/cookies")
>           assert len(r2.json()['cookies']) == 2
E           assert 0 == 2
E            +  where 0 = len({})
tests/integration/test_wild.py:74: AssertionError
```
https://travis-ci.org/jayvdb/vcrpy/jobs/127041061#L473

[...]

PR #252, opened 2016-05-01
Add Python 3.5 and PyPy 3 to Travis

jayvdb:


jayvdb:
successful build on my side https://travis-ci.org/jayvdb/vcrpy/builds/126961687
but likely #251 will mean some jobs need restarting

kevin1024:
Nice!  Think it will pass Travis?

jayvdb:
Youll need kick some of the jobs again and they will pass.

PR #253, opened 2016-05-02
Remove pytest-localserver from test dependencies

jayvdb:
pytest-localserver is no longer needed after
the switch to using pytest-httpbin.


kevin1024:
Good catch!

PR #257, opened 2016-06-09
Handle empty body

JanLikar:


kevin1024:
Thanks for the PR!  Can you provide some example code that shows how this situation comes up?  Or even better - a test?

JanLikar:
If `body` is an empty string, `json.loads()` raises an `ValueError`. This conforms to the JSON spec, in the sense that an empty string has no deserialized counterpart, but I think that for all intents and purposes of vcrpy, `""` and `"''"` should be treated as equivalent.

mark0978:
How would you create a test case for this that shows an empty body?

I disagree with this patch, but am at a loss for how I mock out a mocking library to produce an empty body....

mbachry:
Here's a test that crashes vcrpy with `AttributeError: 'NoneType' object has no attribute 'decode'`. This pr fixes it:

    diff --git a/tests/integration/test_requests.py b/tests/integration/test_requests.py
    index 84a992b..804ab2e 100644
    --- a/tests/integration/test_requests.py
    +++ b/tests/integration/test_requests.py
    @@ -38,6 +38,18 @@ def test_body(tmpdir, httpbin_both):
             assert content == requests.get(url).content


    +def test_get_empty_content_type_json(tmpdir, httpbin_both):
    +    '''Ensure GET with application/json content-type and empty request body doesn't crash'''
    +    url = httpbin_both + '/status/200'
    +    headers = {'Content-Type': 'application/json'}
    +
    +    with vcr.use_cassette(str(tmpdir.join('get_empty_json.yaml')), match_on=('body',)):
    +        status = requests.get(url, headers=headers).status_code
    +
    +    with vcr.use_cassette(str(tmpdir.join('get_empty_json.yaml')), match_on=('body',)):
    +        assert status == requests.get(url, headers=headers).status_code
    +
    +
     def test_effective_url(tmpdir, httpbin_both):
         '''Ensure that the effective_url is captured'''
         url = httpbin_both.url + '/redirect-to?url=/html'

kevin1024:
Thanks @JanLikar and @mbachry  - I have combined your two contributions into #278

PR #259, opened 2016-06-21
Fix deepcopy issue for response headers when `decode_compressed_response` is enabled

nickdirienzo:
Hi @kevin1024 and @IvanMalison,

I ran into this issue after the 1.8.0 release where the first time I would request gzipped data inside a cassette context, my test would fail with a JSON deserialization issue.

I narrowed down the problem to be the data being returned by the stub would not have content encoding set as gzip but the content was not decoded (as the original response should not be modified), so `requests` was failing when I would call `.json()`.

In the filter to decode gzipped data, `headers` is a reference to the original response headers`gzip` is in a list and therefore being shallow copied in the `__init__` of `util.CaseInsensitiveDict`.

Looking forward to your feedback. Cheers!

(Aside: after going through some of the cassette code, I think a better place for the `self._before_record_request` and `self._before_record_response` calls would be in `_save` so `append` doesn't have side-effects. If you think so too happy to send this over in a separate PR.)

### DIFF
--- a/tests/integration/test_stubs.py
+++ b/tests/integration/test_stubs.py
@@ -68,7 +68,7 @@ def test_original_decoded_response_is_not_modified(tmpdir, httpbin):
 
     # Even though the above are raw bytes, the JSON data should have been
     # decoded and saved to the cassette.
-    with vcr.use_cassette(testfile) as cass:
+    with vcr.use_cassette(testfile):
         inside2 = urlopen(request)
         assert 'content-encoding' not in inside2.headers
         assert_is_json(inside2.read())
--- a/tests/integration/test_stubs.py
+++ b/tests/integration/test_stubs.py
@@ -1,5 +1,8 @@
 import vcr
 import six.moves.http_client as httplib
+from six.moves.urllib.request import urlopen, Request
+
+from assertions import assert_is_json
 
 
 def _headers_are_case_insensitive(host, port):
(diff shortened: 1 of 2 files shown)


kevin1024:
Thanks!  I really appreciate the regression test.  And I apologize for... whatever is going on with the travis build.

tests/integration/test_stubs.py
```diff
@@ -44,3 +47,28 @@ def test_multiple_headers(tmpdir, httpbin):
...
+        # to the casssette.
+        assert 'gzip' == inside.headers['content-encoding']
+
+        # They should be the same response.
+        assert inside.headers.items() == outside.headers.items()
+        assert inside.read() == outside.read()
+
+    # Even though the above are raw bytes, the JSON data should have been
+    # decoded and saved to the cassette.
+    with vcr.use_cassette(testfile) as cass:
```
kevin1024: flake8 is angry about `cass` being an unused var

kevin1024:
Nothing to complain about here.  Happy to merge if you can make flake8 stop complaining.

> I think a better place for the self._before_record_request and self._before_record_response calls would be in _save so append doesn't have side-effects.

Sounds good to me in principle but wondering if changing that behavior would break anybody's use case.

From the docs:

> If none of these covers your request filtering needs, you can register a callback that will manipulate the HTTP request before adding it to the cassette. Use the before_record configuration option to so this.

So the docs seem to imply that it prevents the requests from going into the cassette *at all*, not just preventing the request from saving.  Since you can grab the cassette (`with vcr.use_cassette as cass`) and use it to make assertions about http calls that have been made (alternative to mock) this might end up breaking some test cases assuming the previous behavior.

tests/integration/test_stubs.py
nickdirienzo: Whoops. Thanks.

nickdirienzo:
Awesome to hear. Thanks.

Re the aside:
Hmm.. that's a good point. Thanks for entertaining the idea; I didn't read the docs regarding that functionality before suggesting so it was a little out of left field in hindsight.

We've primarily been using it as a "fully-featured" mock, so I'm not sure how others use it (i.e. the `with vcr.use_cassette as cass`). 

I'll leave it alone for now as it's more of a nice-to-have from a code perspective than something mission critical.

nickdirienzo:
Updated the test to not depend on `requests`. I misunderstood when VCR comes into play. Hope this still looks good.

kevin1024:
Thanks!

PR #260, opened 2016-06-24
fix tests in stubs, requests ssl verification and httpbin+flask

### LINKED ISSUE #258: Tests are failing
`test_records_multiple_header_values` and `test_amazon_doctype` are failing with `urllib.error.HTTPError: HTTP Error 500: INTERNAL SERVER ERROR` and
`requests.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:646)`, respectively.

foobarna:
assert in stubs failed because the order differs in lists
`requests` fails sometimes SSL verification, went with `verify=False`
httbin + Flask 0.11 fails, so i pinned it down

this should fix #258

### DIFF
--- a/tests/integration/test_wild.py
+++ b/tests/integration/test_wild.py
@@ -76,7 +76,7 @@ def test_amazon_doctype(tmpdir):
     # amazon gzips its homepage.  For some reason, in requests 2.7, it's not
     # getting gunzipped.
     with vcr.use_cassette(str(tmpdir.join('amz.yml'))):
-        r = requests.get('http://www.amazon.com')
+        r = requests.get('http://www.amazon.com', verify=False)
     assert 'html' in r.text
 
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_wild.py
```diff
@@ -76,7 +76,7 @@ def test_amazon_doctype(tmpdir):
     # amazon gzips its homepage.  For some reason, in requests 2.7, it's not
     # getting gunzipped.
     with vcr.use_cassette(str(tmpdir.join('amz.yml'))):
-        r = requests.get('http://www.amazon.com')
+        r = requests.get('http://www.amazon.com', verify=False)
```
fxfitz: Amazon.com now redirects to https, which is why this test started failing.  Since this test is doing nothing where SSL/TLS is actually needed, I think setting `verify=False` is fine.

For posterity's sake, the underlying reason why this is happening is because `runtests.sh` overrides `REQUESTS_CA_BUNDLE`, needed for some httpbin TLS tests.  This combined with Amazon redirecting is what caused tests to stop failing.

fxfitz:
LGTM 👍

tests/integration/test_wild.py
foobarna: Was thinking about the http/https redirect but they were still failing. Didn't cross my mind about the `runtests.sh`

kevin1024:
Thank you so much!!!!

Issue #209, opened 2015-09-13
Support aiohttp


omerbenamram:
Hi,

I love vcrpy!
now that tornado support is included, any chance of supporting aiohttp anytime soon?

Cheers

IvanMalison:
No plans to add it currently but it looks straightforward enough to do

drgarcia1986:
:+1:  to aiohttp suport

kevin1024:
just added, thanks @lamenezes

PR #266, opened 2016-08-04
Aiohttp support

lamenezes:
Solves #209


kevin1024:
😯wow!

kevin1024:
Thanks for the contribution.  See if you can goof with the travis/pytest config so it only runs the aiohttp tests on python3.4+

lamenezes:
The newest version of flake8 (>3) [1] does not support python 2.6, 3.2 and 3.3 and it is causing  this build to fail when running tests for py26-flakes

[1] https://flake8.pycqa.org/en/latest/release-notes/3.0.0.html

kevin1024:
Heh ok.  See if you can disable flakes on py26 I'm that case

lamenezes:
It finally passed :smile:

drgarcia1986:
👏 👏 👏 
Thanks a lot @lamenezes

PR #267, opened 2016-08-16
Asyncio stub doesn't handle the params passed in inside :params: named argument

anovikov1984:


kevin1024:
Thanks for the PR.  Can you also add a test that shows the bug this patch is fixing?

kevin1024:
Combined into #277

PR #271, opened 2016-09-01
Exclude aiohttp from python < 3 setup

### LINKED ISSUE #270: build failure under python2 with 1.10.0
byte-compiling /builddir/build/BUILDROOT/python-vcrpy-1.10.0-1.fc26.noarch/usr/lib/python2.7/site-packages/vcr/stubs/aiohttp_stubs.py to aiohttp_stubs.pyc
  File "/usr/lib/python2.7/site-packages/vcr/stubs/aiohttp_stubs.py", line 16
    def json(self, *, encoding='utf-8', loads=json.loads):  # NOQA: E999
                    ^
SyntaxError: invalid syntax

I thought aiohttp was python3 only? perhaps it should be excluded for python2 builds? Or python2 support dropped?

lamenezes:
Fixes #270

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,10 @@ except Exception:
             install_requires.extend(value)
 
 
+excluded_packages = ["tests*"]
+if sys.version_info[0] == 2:
+    excluded_packages.append("vcr.stubs.aiohttp_stubs")
+
 setup(
     name='vcrpy',
     version='1.9.0',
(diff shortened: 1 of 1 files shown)


puiterwijk:
Excluding from find_packages only works with packages, not modules.
#272 makes this stub a package so that the exclusion takes effect.

PR #272, opened 2016-09-09
Move vcr.stubs.aiohttp_stub to a package

puiterwijk:
find_packages(exclude=) only works with packages, not modules.
So this fixes install_lib for python2 by correctly excluding that module.

Fixes: #270
Fixes: #271
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>


puiterwijk:
This fixes building on Fedora: http://koji.fedoraproject.org/koji/taskinfo?taskID=15562892

kevin1024:
Thanks!

PR #275, opened 2016-09-20
AttributeError: 'VCRRequestsHTTPSConnectionfixtures/cassettes/test_' object has no attribute 'ssl_context' in with urllib3/python 2.7.12

number5:
This is working for `urllib3` 1.16 but failed for 1.17
vcrpy version: 1.10.0
```py
self = <vcr.patch.VCRRequestsHTTPSConnectionfixtures/cassettes/test_dowload_wavefile object at 0x104375d90>
key_file = None, cert_file = None, cert_reqs = None, ca_certs = None
assert_hostname = None, assert_fingerprint = None, ca_cert_dir = None

    def set_cert(self, key_file=None, cert_file=None,
                 cert_reqs=None, ca_certs=None,
                 assert_hostname=None, assert_fingerprint=None,
                 ca_cert_dir=None):
        """
            This method should only be called once, before the connection is used.
            """
        # If cert_reqs is not provided, we can try to guess. If the user gave
        # us a cert database, we assume they want to use it: otherwise, if
        # they gave us an SSL Context object we should use whatever is set for
        # it.
        if cert_reqs is None:
            if ca_certs or ca_cert_dir:
                cert_reqs = 'CERT_REQUIRED'
>           elif self.ssl_context is not None:
E           AttributeError: 'VCRRequestsHTTPSConnectionfixtures/cassettes/test_' object has no attribute 'ssl_context'
```


hartwork:
@number5 I'm optomistic that with the latest merge to `master` [the test suite](https://travis-ci.org/kevin1024/vcrpy/builds/329568738) will return to zero failures. Please rebase this pull request against latest `master` so we can see if it breaks any tests. Thank you!

neozenith:
@number5 sorry this got dropped for the last 2.5 years. 

If you'd still like to put this in, could you rebase with master and add a test case that capture this? 

Thanks.

neozenith:
Sorry to ask again but could you rebase against the latest master when you get the chance. Looks like Travis was run against an environment configuration that doesn't match tox

number5:
the tests looks good

PR #276, opened 2016-10-01
Add tests for aiohttp params fix

lamenezes:
I've been using vcr with aiohttp recently and I've encountered the same bug that PR #267 solves. My PR implement tests for #267 PR. Tests on this branch will fail because they are lacking code on the other PR.


kevin1024:
Combined into #277

PR #280, opened 2016-11-07
Fix up to support aiohttp 1.x

madninja:
This should fix up aiohttp 1.x support.


kevin1024:
Thanks!  Looks like a couple tests are failing.  Can you check those out?  Might just be a pyflakes issue.

madninja:
@kevin1024 all set, thanks for the catch

kevin1024:
Just pushed a new release including this PR, thanks for the contribution 👍

PR #282, opened 2016-12-18
Remove duplicate mock triple

FooBarQuaxx:


kevin1024:
Nice catch.  Was this causing a bug?

FooBarQuaxx:
No I don't think so. I stumbled upon it while exploring the code.

IvanMalison:
Was this supposed to be the normal HTTP version

PR #284, opened 2016-12-21
Add command line utility for editing cassettes

### LINKED ISSUE #249: Add script, or document how to convert yaml binary response body into human-readable
My vcrpy responses come back gzipped, so yaml encodes them as binary objects. It would be good if there were some easy way to read these as a human. I'm not sure what that would look like. Perhaps it could be a script which decodes these objects?

I also posted on Stack Overflow: https://stackoverflow.com/questions/36366234

sirex:
When you have large response or request bodies it is almoust imposible to edit
then directly in YAML file.

To fix that, I created a command line tool, that will open request or response
body for editing in an external editor and when you save you changes, cassette
file will be updated with you changes.

This saves a lot of time and prevents errors when trying to edit large
responses directly in YAML file.

Teststed this on py27,py34,py35, all tests pass.

Added new section in documentaiton about how to use command line tool.

Probably this fixes #249.


setup.py
```diff
@@ -68,7 +68,12 @@ def run_tests(self):
     install_requires=install_requires,
     extras_require=extras_require,
     license='MIT',
-    tests_require=['pytest', 'mock', 'pytest-httpbin'],
+    tests_require=['pytest', 'mock', 'pytest-mock', 'pytest-httpbin'],
+    entry_points={
+        'console_scripts': [
+            'vcr = vcr.main:main',
```
IvanMalison: maybe call this vcredit or something like that

neozenith:
Is there still a need for this PR? It has been 2.5 years and the library has moved on a bit.

PR #288, opened 2016-12-26
Errorfix: Add missing requirement yarl

lamenezes:
This PR fixes the following error that was introduced on vcrpy==1.10.4:

```python
    from aiohttp import ClientResponse
>   from yarl import URL
E   ImportError: No module named 'yarl'

../virtualenvs/venv-3.5.1/lib/python3.5/site-packages/vcr/stubs/aiohttp_stubs/__init__.py:9: ImportError
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ extras_require = {
     ':python_version in "2.4, 2.5, 2.6"':
         ['contextlib2', 'backport_collections', 'mock'],
     ':python_version in "2.7, 3.1, 3.2"': ['contextlib2', 'mock'],
+    ':python_version in "3.4, 3.5, 3.6"': ['yarl'],
 }
 
 


kevin1024:
Hmm.  What is yarl and is there a stdlib equivalent?  Guess I missed this when merging the PR.  Maybe we should have an aiohttp test or two.

lamenezes:
yarl [1] is a lib to handle URLs. its equivalent on stdlib is urllib.

there are tests for aiohttp [2] but they won't break because newer versions of aiohttp have yarl as a requirement.

[1] http://yarl.readthedocs.io/en/latest/
[2] https://github.com/kevin1024/vcrpy/blob/master/tests/integration/test_aiohttp.py

lamenezes:
we can't get rid of yarl and mantain support to aiohttp. it is mandatory to use `yarl.URL` on aiohttp requests [1] and responses [2] on its 1.x versions

[1] https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client_reqrep.py#L74
[2] https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client_reqrep.py#L519

lamenezes:
we also can't leave yarl as an global requirement to vcrpy as it won't work for python 2.x. i'm gonna take some time and think on a better solution for this.

kevin1024:
I believe we already have some requirements that are conditional on python version, maybe we can use the same technique: https://github.com/kevin1024/vcrpy/blob/master/setup.py#L30

lamenezes:
Fixed, build passing.

PR #289, opened 2017-01-04
added a fix to httplib2

carlosds730:
If we are doing a "real" connect, we need to `force_reset`, otherwise we get and unbound method error.

### DIFF
--- a/vcr/stubs/__init__.py
+++ b/vcr/stubs/__init__.py
@@ -287,7 +287,9 @@ class VCRConnection(object):
             # Cassette is write-protected, don't actually connect
             return
 
-        return self.real_connection.connect(*args, **kwargs)
+        from vcr.patch import force_reset
+        with force_reset():
+            return self.real_connection.connect(*args, **kwargs)
 
     @property
     def sock(self):


kevin1024:
Thanks!  Could you add a test that reproduces the issue?

carlosds730:
To reproduce it just create a `VCRHTTPSConnection` whose base class is `httplib.HTTPSConnection` and call its connect, if you don't have a recorded response for that request and the cassette is not write protected the connect method will call `self.real_connection.connect()` and this will raise an Exception.

carlosds730:
```python
@mock.patch('vcr.cassette.Cassette.can_play_response_for', return_value=False)
def testing_connect(self):
    vcr_connection = VCRHTTPSConnection('www.google.com')
    vcr_connection.cassette.write_protected = False
    vcr_connection.real_connection.connect()
    assert vcr_connection.real_connection.sock is not None
```

carlosds730:
The test needs to be fixed,
```python
@mock.patch('vcr.cassette.Cassette.can_play_response_for', return_value=False)
def testing_connect(self):
    vcr_connection = VCRHTTPSConnection('www.google.com')
    vcr_connection.cassette = Cassette('test', record_mode='all')
    vcr_connection.real_connection.connect()
    assert vcr_connection.real_connection.sock is not None
```

carlosds730:
Thanks

PR #292, opened 2017-01-12
Allow injection of persistence methods

j-funk:
I'm using the ability to inject custom persistence methods for `save_cassette` and `load_cassette` in order to save cassettes to a DB in a service I've built which uses vcrpy as a dependency.  Perhaps other folks will appreciate the ability to override the persistence methods?

You can see the custom methods I inject for my project [here](https://github.com/j-funk/bond/blob/master/bond/contract_manager/contract_manager.py#L201-L224).

I've added a couple of new integration tests for save/load injection. Let me know if there's anything else you'd like for this PR!


kevin1024:
Thanks!  I think this is a great idea.  I especially appreciate that you have tests in this PR. I would be happy to add this but I would like to change the API a bit.  

If you go here, you can see that we have support for custom serializers and custom matchers:

https://github.com/kevin1024/vcrpy/blob/master/docs/advanced.rst#register-your-own-serializer

I would like to let you define custom persisters using a similar API, rather than a callback.  What do you think?

j-funk:
Great thanks for the feedback. Seems pretty straightforward.  I'll update the PR and see what you think!

j-funk:
Hey @kevin1024 - so it's refactored and almost all tests are passing, however there is some dark mock / lambda voodoo mixed in with [one of the tests](https://github.com/kevin1024/vcrpy/pull/292/files#diff-a3a5b0fd081ddc14b3e309c2782b4cd7R88).  I've spent quite a lot of time attempting to make this test pass. Looks like we need to get the class instance into the lambda somehow, but it's beyond me.  Would you be willing to take a look?

```
TypeError: unbound method <lambda>() must be called with FilesystemPersister 
instance as first argument (got str instance instead)
```

kevin1024:
Dark mock and lambda voodoo?  Paging @IvanMalison 😛

IvanMalison:
I'll take a look later today

IvanMalison:
Its just because that lambda is not a class or static method. This means that when it is retrieved from the FilesystemClass it comes back as an unbound method, which causes an issue because the function is then called without a FileSystemClass instance as its first argument.

I made the appropriate change in https://github.com/j-funk/vcrpy/pull/1

kevin1024:
Thanks @Ivanmalison :)

j-funk:
@IvanMalison thank you :)

kevin1024:
@j-funk is this PR ready to merge?

kevin1024:
Oh, actually, can you add a section to the docs showing how to use it? :cake:

j-funk:
@kevin1024 I've updated the docs and made a minor adjustment to the code.  Assuming tests pass, the only thing I am not sure about is the rst link I've made to the `FilesystemPersister` class.

kevin1024:
Awesome, thanks!!!

Issue #293, opened 2017-01-23
VCRConnection does not support Python 3.6 - 'encode_chunked'


derekbekoe:
In Python 3.6 http.client.HTTPConnection.request, a new keyword argument of `encode_chunked` was added. Currently, it doesn't look like VCR does not support this change so you get a stacktrace.
https://docs.python.org/3.6/library/http.client.html#http.client.HTTPConnection.request

When VCR patches `HTTPConnection` with `vcr.patch.VCRHTTPSConnection`, it does not accept this new keyword argument and so you see the following stacktrace for example:
```
  File "/usr/local/lib/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/lib/python3.6/urllib/request.py", line 526, in open
    response = self._open(req, data)
  File "/usr/local/lib/python3.6/urllib/request.py", line 544, in _open
    '_open', req)
  File "/usr/local/lib/python3.6/urllib/request.py", line 504, in _call_chain
    result = func(*args)
  File "/usr/local/lib/python3.6/urllib/request.py", line 1361, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/usr/local/lib/python3.6/urllib/request.py", line 1318, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
TypeError: request() got an unexpected keyword argument 'encode_chunked'
```

Similar to: https://github.com/boto/botocore/issues/1079 where the solution was in PR https://github.com/boto/botocore/pull/1107 (accept the kwargs but don't use them).

---

More info:

In the `do_open` method of `/usr/local/lib/python3.6/urllib/request.py`, a `http_class` class is expected. VCR passes in the class `vcr.patch.VCRHTTPSConnection` (an instance of `VCRConnection`) for this.

In `do_open`, Python calls `http_class(...).request(..., encode_chunked)`. Since VCR uses `VCRHTTPSConnection` here, `VCRHTTPSConnection.request()`. This method (https://github.com/kevin1024/vcrpy/blob/master/vcr/stubs/__init__.py#L156) does not accept the `encode_chunked` keyword and so we get `TypeError: request() got an unexpected keyword argument 'encode_chunked'`.

PR #294, opened 2017-01-23
Correctly patch HTTPConnection.request for Python 3.6

derekbekoe:
Fixes https://github.com/kevin1024/vcrpy/issues/293

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ python:
 - 3.3
 - 3.4
 - 3.5
+- 3.6
 - pypy
 - pypy3
 install:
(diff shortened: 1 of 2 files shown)


lamenezes:
Nice PR @derekbekoe! 

Do you think you could add some tests that reproduce the bug reported in issue #293?

derekbekoe:
Will take a look!

derekbekoe:
Testing this requires adding Python 3.6 to Travis CI.
I did this and other tests fail because of changes to HTTPConnection (https://docs.python.org/3.6/library/http.client.html) which `VCRConnection` is based from (though not a subclass of).

I don't have time right now to fix up all tests for full Python 3.6 support with VCR.
I have reverted the change to `.travis.yml` for now.

PR #300, opened 2017-03-10
use a lock to make force_reset thread safe

### LINKED ISSUE #212: force_reset makes vcr non thread-safe

### LINKED ISSUE #295: Second thread not recorded
When run with multiple threads `THREADS = 2`, only the one call is recorded:

```python
import requests
import concurrent.futures
import vcr


THREADS = 2


@vcr.use_cassette('first-only-used.yaml')
def load_urls(urls):
    with concurrent.futures.ThreadPoolExecutor(
            max_workers=THREADS) as executor:
        fs = [executor.submit(requests.get, url) for url in urls]
        for f in concurrent.futures.as_completed(fs):
            try:
                result = f.result()
                print('Result: %s' % result.json())
            except Exception as e:
                print('Error %s' % e)


if __name__ == '__main__':
    urls = [
        'https://httpbin.org/get?q=first',
        'https://httpbin.org/get?q=second'
    ]
    load_urls(urls)
```

First call:
```
% env/bin/python thread_vcr.py                                                                                                                                       
Result: {u'origin': u'xx.180.xx.95', u'headers': {u'Host': u'httpbin.org', u'Accept-Encoding': u'gzip, deflate', u'Accept': u'*/*', u'User-Agent': u'python-requests/2.12.1'}, u'args': {u'q': u'second'}, u'url': u'https://httpbin.org/get?q=second'}
Result: {u'origin': u'xx.180.xx.95', u'headers': {u'Host': u'httpbin.org', u'Accept-Encoding': u'gzip, deflate', u'Accept': u'*/*', u'User-Agent': u'python-requests/2.12.1'}, u'args': {u'q': u'first'}, u'url': u'https://httpbin.org/get?q=first'}
```

Second call:
```
% env/bin/python thread_vcr.py
Error No match for the request (<Request (GET) https://httpbin.org/get?q=second>) was found. Can't overwrite existing cassette ('first-only-used.yaml') in your current record mode ('once').
Result: {u'origin': u'xx.180.xx.95', u'headers': {u'Host': u'httpbin.org', u'Accept-Encoding': u'gzip, deflate', u'Accept': u'*/*', u'User-Agent': u'python-requests/2.12.1'}, u'args': {u'q': u'first'}, u'url': u'https://httpbin.org/get?q=first'}
```

Yaml:
```
% cat first-only-used.yaml 
interactions:
- request:
    body: null
    headers:
      Accept: ['*/*']
      Accept-Encoding: ['gzip, deflate']
      Connection: [keep-alive]
      User-Agent: [python-requests/2.12.1]
    method: GET
    uri: https://httpbin.org/get?q=first
  response:
    body: {string: !!python/unicode "{\n  \"args\": {\n    \"q\": \"first\"\n  },\
        \ \n  \"headers\": {\n    \"Accept\": \"*/*\", \n    \"Accept-Encoding\":\
        \ \"gzip, deflate\", \n    \"Host\": \"httpbin.org\", \n    \"User-Agent\"\
        : \"python-requests/2.12.1\"\n  }, \n  \"origin\": \"xx.180.xx.95\", \n \
        \ \"url\": \"https://httpbin.org/get?q=first\"\n}\n"}
    headers:
      access-control-allow-credentials: ['true']
      access-control-allow-origin: ['*']
      connection: [keep-alive]
      content-length: ['267']
      content-type: [application/json]
      date: ['Tue, 07 Feb 2017 14:29:38 GMT']
      server: [nginx]
    status: {code: 200, message: OK}
version: 1
```

graingert:
Fixes #295, Fixes #212

### DIFF
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -1,6 +1,7 @@
 '''Utilities for patching in cassettes'''
 import functools
 import itertools
+import threading
 
 from .compat import contextlib, mock
 from .stubs import VCRHTTPConnection, VCRHTTPSConnection
@@ -447,9 +448,13 @@ def reset_patchers():
         )
 
 
+_force_reset_lock = threading.RLock()
+
+
 @contextlib.contextmanager
 def force_reset():
     with contextlib.ExitStack() as exit_stack:
+        exit_stack.enter_context(_force_reset_lock)
         for patcher in reset_patchers():
             exit_stack.enter_context(patcher)
         yield


IvanMalison:
What about nested calls to force reset?

The docs on threading.Lock make it sound as though it doesnt do that thing where the same thread can reacquire the lock if needed

> A factory function that returns a new primitive lock object. Once a thread has acquired it, subsequent attempts to acquire it block, until it is released; any thread may release it.

graingert:
ah probably best to use RLock then

IvanMalison:
I'm also not really convinced that this makes anything threadsafe. What if you have some other thread in a portion of the code that hasn't yet hit a force reset? It will still be allowed to operate, and thus be exposed to the effects of the force reset.

I'm pretty sure that the only way to guarantee thread safety (assuming that force_reset is used) is to completely lock all activity whenever we enter a cassette.

IvanMalison:
@graingert As mentioned in my comment above, I don't think this actually gives us thread safety. Feel free to reopen if you think it does.

selevit:
@IvanMalison, I agree that this fix may not cover all multithreading corner cases, but it works and makes the library's threadsafety definitely better than it was before. Maybe it makes sense to merge this change until the better solution is designed?

Now I have to monkeypatch `vcrpy` to make it work in different threads, and to be hones, that's not the way I like a lot :-)

IvanMalison:
Nothing about this is threadsafe at all. Doesn't make sense to do anything to try to improve thread safety if we are not fully guaranteeing it, which we cannot do. These aren't just edge cases. People will but these issues if we start indicating that vcrpy is in any way threadsafe.

PR #301, opened 2017-03-14
handle pytest-asyncio async def coroutines

graingert:
I'm currently using:

``` python
from vcr.cassette import CassetteContextDecorator

patchy.mc_patchface(CassetteContextDecorator._handle_coroutine, """\
    @@ -1,4 +1,4 @@
    -def _handle_coroutine(self, function, args, kwargs):
    +async def _handle_coroutine(self, function, args, kwargs):
         """Wraps a coroutine so that we're inside the cassette context for the
         duration of the coroutine.
         """
    @@ -6,11 +6,4 @@
             coroutine = self.__handle_function(cassette, function, args, kwargs)
             # We don't need to catch StopIteration. The caller (Tornado's
             # gen.coroutine, for example) will handle that.
    -        to_yield = next(coroutine)
    -        while True:
    -            try:
    -                to_send = yield to_yield
    -            except Exception:
    -                to_yield = coroutine.throw(*sys.exc_info())
    -            else:
    -                to_yield = coroutine.send(to_send)
    +        await coroutine
""")
```

to handle:

```python
@my_vcr.use_cassette(filter_headers=['authorization'])
@pytest.mark.asyncio
async def test_some_thing() -> None:
    res = await some_thing()
    assert res.to_dict() == {'text': 'a', 'score': 0.1, 'entities': []}
```

if I don't include the patch, I get:

```pytb
  File "/home/graingert/projects/x/datascience/.tox/py35/lib/python3.5/site-packages/pytest_asyncio/plugin.py", line 77, in pytest_pyfunc_call
    asyncio.async(pyfuncitem.obj(**testargs), loop=event_loop))
  File "/usr/lib/python3.5/asyncio/base_events.py", line 466, in run_until_complete
    return future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
    raise self._exception
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/home/graingert/projects/x/datascience/.tox/py35/lib/python3.5/site-packages/vcr/cassette.py", line 113, in _handle_coroutine
    to_yield = next(coroutine)
TypeError: 'coroutine' object is not an iterator
```

### DIFF
--- a/tests/integration/aiohttp_utils.py
+++ b/tests/integration/aiohttp_utils.py
@@ -1,7 +1,13 @@
 import asyncio
+import aiohttp
 
 
 @asyncio.coroutine
-def aiohttp_request(session, method, url, as_text, **kwargs):
-    response = yield from session.request(method, url, **kwargs)  # NOQA: E999
-    return response, (yield from response.text()) if as_text else (yield from response.json())  # NOQA: E999
+def aiohttp_request(loop, method, url, as_text, **kwargs):
+    with aiohttp.ClientSession(loop=loop) as session:
+        response = yield from session.request(method, url, **kwargs)  # NOQA: E999
+        if as_text:
+            content = yield from response.text()  # NOQA: E999
+        else:
+            content = yield from response.json()  # NOQA: E999
+        return response, content
--- a/tests/integration/async_def.py
+++ b/tests/integration/async_def.py
@@ -1,0 +1,13 @@
+import aiohttp
+import pytest
+import vcr
+
+
+@vcr.use_cassette()
+@pytest.mark.asyncio
+async def test_http():  # noqa: E999
+    async with aiohttp.ClientSession() as session:
+        url = 'https://httpbin.org/get'
+        params = {'ham': 'spam'}
+        resp = await session.get(url, params=params)  # noqa: E999
+        assert (await resp.json())['args'] == {'ham': 'spam'}  # noqa: E999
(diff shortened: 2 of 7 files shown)


lamenezes:
This PR looks good. Good job :+1: :cake: 

Do you think it is possible to add some tests here? :slightly_smiling_face:

lamenezes:
This PR looks good. Good job :+1: :cake: 

Do you think it is possible to add some tests here? :slightly_smiling_face:

graingert:
@lamenezes @kevin1024 and it finally passes!

kevin1024:
Woohoo!  Thanks so much for this.

I noticed the addition of pytest-asyncio as a dependency.  Since vcrpy is packaged in a few Linux distros, and as part of the packaging process they like to run the tests, that means the distros will also have to package pytest-asyncio.  Not sure if that's going to be a problem for them or not, but if it's just syntactic sugar is there a way to do without it?

graingert:
@kevin1024 This is the specific usecase where my issue cropped up, so I want to test it.

lamenezes:
@graingert @kevin1024 maybe we should skip pytest-asyncio tests if it isn't installed?

graingert:
Meh, it's a teeny tiny pure python package. Package maintainers can patch
out those tests if they really can't install it

PR #304, opened 2017-03-27
Allow errors parameter in MockClientResponse.text()

ntpz:
This is to match aiohttp behavior

### DIFF
--- a/vcr/stubs/aiohttp_stubs/__init__.py
+++ b/vcr/stubs/aiohttp_stubs/__init__.py
@@ -18,8 +18,8 @@ class MockClientResponse(ClientResponse):
         return loads(self.content.decode(encoding))
 
     @asyncio.coroutine
-    def text(self, encoding='utf-8'):
-        return self.content.decode(encoding)
+    def text(self, encoding='utf-8', errors='strict'):
+        return self.content.decode(encoding, errors=errors)
 
     @asyncio.coroutine
     def release(self):


kevin1024:
Thanks!  Could you please add a test that shows the bug this fixes?

notapresent:
Here's the code: 

```
async def fetch(session, url):
    response = await session.get(url)
    text = await response.text(errors='ignore')
    return text

class DownloaderTestCase(VCRMixin):
    def setUp(self):
        super(DownloaderTestCase, self).setUp()
        self.loop = asyncio.new_event_loop()
        asyncio.set_event_loop(self.loop)

    def test_fetch_downloads(self):
        async def go():
            async with aiohttp.ClientSession(loop=self.loop) as session:
                return await fetch(session, 'http://httpbin.org/')

        body = self.loop.run_until_complete(go())
        self.assertIn('ENDPOINTS', body)
```

Without this change, it raises: 
```
TypeError: text() got an unexpected keyword argument 'errors'
```

I'm not sure how to write test for this though. Should I write a unit test for `vcr.stubs.aiohttp_stubs.MockClientResponse` with  mocked `content` attribute and use something like `content.decode.assert_called_once_with(errors='ignore')` ?

lamenezes:
@notapresent thanks for your PR! Sorry for the delay, but your propose really makes sense.

If you're still interested in contributing I think you can test this by calling `await response.text(errors='something')` on the current test base. What do you think?

arthurHamon2:
@ntpz I'm closing this as it looks like it has been fixed in another PR

PR #306, opened 2017-05-04
aiohttp multidict query parameters

ginoledesma:
aiohttp supports multidict for headers and query parameters, allowing the user
to specify either a dict or a list of 2-item tuples as input. For example:

```python
async with aiohttp.ClientSession() as s:
    params = [('a', 'alpha'), ('b', 'bravo'), ('a', 'Alpha')]
    async with s.get('http://127.0.0.1:8888', params=params) as r:
        buf = await r.text()
```

This patches vcrpy to handle this specific case instead of only dict-like
objects.

### DIFF
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -111,6 +111,19 @@ def test_params(tmpdir, scheme):
         assert cassette.play_count == 1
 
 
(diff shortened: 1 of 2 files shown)


IvanMalison:
@ginoledesma Any chance you can resolve conflicts here?

PR #307, opened 2017-05-09
When filtering post data, handle ContentType with specified charset (ex: "application/json;UTF-8")

poussik:
Hello,

Regarding the website Amundi.com, I was unable to filter the POST data username & password.
The Content-Type set is "application/json;UTF-8", so strict comparison with "application/json" is not possible.

Thanks for this great python module!

### DIFF
--- a/vcr/filters.py
+++ b/vcr/filters.py
@@ -84,7 +84,7 @@ def replace_post_data_parameters(request, replacements):
     """
     replacements = dict(replacements)
     if request.method == 'POST' and not isinstance(request.body, BytesIO):
-        if request.headers.get('Content-Type') == 'application/json':
+        if request.headers.get('Content-Type').startswith('application/json'):
             json_data = json.loads(request.body.decode('utf-8'))
             for k, rv in replacements.items():
                 if k in json_data:


graingert:
FYI json must be in UTF8, so that flag is redundant. 

Matching the prefix is also incorrect, because it would include `application/json-not-really-json`

graingert:
Maybe https://pypi.python.org/pypi/python-mimeparse will help

poussik:
Hi,

Probably you read too quickly so you misunderstood the patch, FYI d.get(X, '') returns None when d[X] == None.
Regarding mimeparse I do not think adding a dependency for a so much small problem is a good idea.
I have updated this PR using standard cgi module (like mimeparse), and kept the logic as it was before.

Imho the best way to do would be to try except a json.loads instead of relying on Content-Type, but it is not aligned with the coding style.

vcr/filters.py
```diff
@@ -84,7 +85,9 @@ def replace_post_data_parameters(request, replacements):
     """
     replacements = dict(replacements)
     if request.method == 'POST' and not isinstance(request.body, BytesIO):
-        if request.headers.get('Content-Type') == 'application/json':
+        # Handles headers['Content-Type'] in [application/json, application/json;charset=XXX, None]
+        # Does not accept Content-Type in [application/json-not-really-json, application/custom-json]
```
graingert: please remove these comments, and convert them into unit tests.

poussik: Indeed! Many thanks for the review.

poussik:
Hello @graingert ,
The last build has failed because of a flickr timeout, can you please relaunch it ?

lamenezes:
@poussik thanks for your PR! Sorry for the delay, but your propose really makes sense.

Can you update your branch so we can follow with the review/merge process? :slightly_smiling_face:

PR #321, opened 2017-07-01
Update usage.rst

mikaeilorfanian:
"i.e." is misleading because it means "in other words". In the context of this paragraph, "i.e." implies that "unexpected requests" are those that have a different URI.
I think a URI change is just one example of what could constitute an "unexpected" request. vcr does request matching based on method, port, body, etc.
So, I think "e.g." which means "for example" is the right phrase in this context.


kevin1024:
Learned something new.  Thanks!

PR #324, opened 2017-08-09
Fix of registering custom persister and test.

### LINKED ISSUE #334: It seems that "register_persister()" does not work
It seems that the "self.persister" attribute is never propagated to new Cassettes created, since get_merged_config() doesn't look it up.

The code and the related test works because only FileSystemPersister is used in both.

petr-bulusek:
Old test_register_persister.py used FilesystemPersister, which is used as default, so didn't catch the problem. Without added missing persister config to config.py the test fails, because it's using the default even if register_persister is called.

Fixes #334

### DIFF
--- a/tests/integration/test_register_persister.py
+++ b/tests/integration/test_register_persister.py
@@ -10,10 +10,27 @@ import vcr
 from vcr.persisters.filesystem import FilesystemPersister
 
 
+class CustomFilesystemPersister(object):
+    """
+    like default FilesystemPersister but
+    adds .test extension to cassette file
+    """
+    @staticmethod
+    def load_cassette(cassette_path, serializer):
+        cassette_path += '.test'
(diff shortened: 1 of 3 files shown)


tests/integration/test_register_persister.py
```diff
@@ -10,10 +10,25 @@
 from vcr.persisters.filesystem import FilesystemPersister
 
 
+class CustomFilesystemPersister(object):
```
IvanMalison: I'm going to merge this as is, but I think it would be much better to write a test that just asserts that the function is called with the right stuff. Doing it this way is just so much more work.

IvanMalison:
@petr-bulusek Congrats on your first github pull request. Cool that github started showing that, and always good to see people dipping their feet in the world of open source!

PR #340, opened 2017-12-07
Drop support for EOL Python 2.6 and 3.3

### LINKED ISSUE #338: Tests fail on master beacause of Python 2.6 support
Tests are currently failing on master because a number of libraries have stopped supporting Python 2.
What I found so far, the following don't support Python 2.6:
[wheel>=0.30.0](https://github.com/pypa/wheel/blob/7ca7b3552e55030b5d78cd90d53f1d99c9139f16/CHANGES.txt#L15) 
[ruamel.yaml>=0.12.0](https://bitbucket.org/ruamel/yaml/commits/6eca2e8f28dc2ed64f065972327be5a0e381549d?at=0.12.0) 

(There were other failing tests that I fixed in https://github.com/kevin1024/vcrpy/pull/319/commits/36b51622301b68b6e46243f83f3c9ab87809c162)

hugovk:
Fixes #338.

Also removes redundant code that was required to maintain support for Python 2.6.

Here's the pip installs for vcrpy-unittest from PyPI for the last month (via `pypinfo --percent --pip vcrpy-unittest pyversion`) showing nothing for Python 2.6 or 3.3.

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   47.5% |            699 |
| 3.6            |   43.7% |            643 |
| 3.5            |    4.8% |             70 |
| 3.4            |    4.0% |             59 |

Also includes @samuelfekete's CI fixes from 36b5162 to (mostly) get the CI green! Thanks!
(I would have cherry-picked that commit but git couldn't find it.)


hartwork:
Looks pretty good already. Can you drop the unrelated `.idea` patch and squash the 3.3 removal bits into related commits to obtain a cleaner history?

For the pypy3 failures, maybe make a commit dropping it or adding it to allowed failures? Error
```
ImportError: /home/travis/build/kevin1024/vcrpy/.tox/requests22/site-packages/_regex.pypy3-58-x86_64-linux-gnu.so: undefined symbol: PyUnicode_Substring
```
does not look like "our fault" to me. That would leave only 3.6 to be fixed.

samuelfekete:
Thanks @hugovk!  Looking good so far.

hugovk:
Will do! 

May I add the .idea ignoring in another PR? I guess you don't want xml files from PyCharm ending up in the repo.

hugovk:
Updated!

(Not changed .gitignore, let me know).

The failures are intermittent, I restarted the failures on my Travis and got them all green: https://travis-ci.org/hugovk/vcrpy/builds/313402284

tests/integration/test_ignore.py
```diff
@@ -26,9 +26,9 @@ def test_ignore_localhost(tmpdir, httpbin):
     with overridden_dns({'httpbin.org': '127.0.0.1'}):
         cass_file = str(tmpdir.join('filter_qs.yaml'))
         with vcr.use_cassette(cass_file, ignore_localhost=True) as cass:
-            urlopen('http://localhost:{0}/'.format(httpbin.port))
+            urlopen('http://localhost:{}/'.format(httpbin.port))
```
vladcalin: It think it's still a good practice to keep the index in the format. Just an observation, afaik pep8 or any other standard does not enforce anything for this.

hugovk: I can revert those if needed. I prefer to omit them if possible to aid readability and reduce cognitive load: if I see some numbers in there then I know it's a special case, like different ordering.

vladcalin: It's just personal preference, and what you just said makes sense. In this case makes sense to keep the indexes out of it 😄

hartwork:
> The failures are intermittent

Do you have permissions to hit the retry button on Travis?
Any ideas on getting the tests less flakey?

@kevin1024 what do you think about this pull request?

hugovk:
> Do you have permissions to hit the retry button on Travis?

I do for https://travis-ci.org/hugovk/vcrpy/ (where I hit it for some jobs), but not for https://travis-ci.org/kevin1024/vcrpy/

> Any ideas on getting the tests less flaky?

I'm not familiar with these tests, but some of them look like intermittent failures from making network calls (eg. `ConnectionResetError: [Errno 104] Connection reset by peer`)

Either rewrite to avoid using the network, or perhaps use a pytest plugin such as [flaky](https://github.com/box/flaky#flaky) or [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures#pytest-rerunfailures)?

paultiplady:
@hugovk you can put `.idea` and other system-specific ignores in your global gitignore file (`$HOME/git/ignore`), rather than adding it to individual dependencies. 

https://git-scm.com/docs/gitignore

hugovk:
@paultiplady Unfortunately I cannot, this belongs in the project gitignore and cannot go in the global gitignore because some projects require some of the files from this directory to be in Git.

See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839 for more info.

That is, none of those files from the .idea dir should be in this project's repo.

But some other projects do want to share IDE config via some files from the .idea dir. So it's not a global level thing, it's a project-level thing.

I assume vcrpy does not want any of that config, hence ignoring the entire dir in this case.

What sort of problems does it cause to have an extra entry in this gitignore? 

I would think it is better to make sure irrelevant files don't end up being committed here.

graingert:
> What sort of problems does it cause to have an extra entry in this gitignore?

we don't want .idea specific config here, otherwise we'd have to maintain an ignore for every single editor.

hartwork:
> But some other projects do want to share IDE config via some files from the .idea dir. So it's not a global level thing, it's a project-level thing.

It should be possible to have it in global ignore and still force Git to add it to certain repos where wanted.

graingert:
@hugovk it's irrelevant whether or not it's useful in vcrpy adding .idea to .gitignore is a separate concern to `Drop support for EOL Python 2.6 and 3.3` and is blocking merging. Take it out of this PR make a separate PR.

hugovk:
Aha, thanks for the tips on unignoring the global ignore, I'll give that a shot! I've removed it from this project's one!

.travis.yml
```diff
@@ -52,21 +48,15 @@ matrix:
...
       python: pypy
-    - env: TOX_SUFFIX="aiohttp"
-      python: pypy3
 python:
-- 2.6
 - 2.7
 - 3.5
 - 3.6
 - pypy
-- pypy3
```
graingert: any chance of adding:
```yaml
  - "pypy2.7-5.8.0"
  - "pypy3.5-5.9.0"
```

graingert: eg instead of removing `pypy3` replace it with `pypy3.5-5.9.0`

hugovk: Added.

.travis.yml
```diff
@@ -52,21 +48,17 @@ matrix:
...
     - env: TOX_SUFFIX="requests1"
       python: 3.6
-    - env: TOX_SUFFIX="aiohttp"
-      python: 2.6
     - env: TOX_SUFFIX="aiohttp"
       python: 2.7
     - env: TOX_SUFFIX="aiohttp"
       python: pypy
-    - env: TOX_SUFFIX="aiohttp"
-      python: pypy3
```
graingert: this pypy3 should be pypy3.5-5.9.0

.travis.yml
```diff
@@ -42,8 +40,6 @@ matrix:
       python: 3.5
     - env: TOX_SUFFIX="flakes"
       python: pypy
-    - env: TOX_SUFFIX="flakes"
-      python: pypy3
```
graingert: this pypy3 should be pypy3.5-5.9.0

hugovk: Will update these. Should the vanilla `- pypy` be removed?

graingert: dunno. I'm tempted to have both. @kevin1024 ^

.travis.yml
graingert: oh I'm dumb this is under 'exclude' it shouldn't be here.

.travis.yml
hugovk: And should `python: pypy` be replaced with `python: "pypy2.7-5.8.0"`?

graingert: oh I'm dumb this is under 'exclude' it shouldn't be here.

graingert: > And should python: pypy be replaced with python: "pypy2.7-5.8.0"?

I can't decide it if `pypy` should be replaced with `pypy2.7-5.8.0` or `pypy2.7-5.8.0` should be added in addition to the existing plain `pypy` @kevin1024 ^

.travis.yml
```diff
@@ -52,21 +48,19 @@ matrix:
...
       python: 3.6
-    - env: TOX_SUFFIX="aiohttp"
-      python: 2.6
     - env: TOX_SUFFIX="aiohttp"
       python: 2.7
     - env: TOX_SUFFIX="aiohttp"
       python: pypy
     - env: TOX_SUFFIX="aiohttp"
-      python: pypy3
+      python: "pypy3.5-5.9.0"
```
graingert: <del>this pypy3 should be pypy3.5-5.9.0</del>

I'm dumb. This is under exclude. and we don't want to exclude pypy3.5 from aiohttp tests.

tests/integration/test_wild.py
```diff
@@ -80,13 +81,25 @@ def test_amazon_doctype(tmpdir):
     assert 'html' in r.text
 
 
-def test_xmlrpclib(tmpdir):
+@pytest.yield_fixture(scope='session')
+def rpc_server():
```
graingert: what's this for?

.travis.yml
hugovk: OK, this is back to how it was :)

tests/integration/test_wild.py
```diff
@@ -80,13 +81,25 @@ def test_amazon_doctype(tmpdir):
...
+    httpd = xmlrpc_server.SimpleXMLRPCServer(('', 0))
+    httpd.register_function(pow)
+    proxy_process = multiprocessing.Process(
+        target=httpd.serve_forever,
+    )
+    proxy_process.start()
```
graingert:
```python
try:
    proxy_process.start()
    yield 'http://{}:{}'.format(*httpd.server_address)
    finally:
        proxy_process.terminate()
```

tests/integration/test_wild.py
hugovk: It's to remove dependency on bugs.python.org/xmlrpc which has a 301 Moved Permanently.

> Also includes @samuelfekete's CI fixes from 36b5162 to (mostly) get the CI green!

graingert:
looks like there's a flake8 violation:
```shell
./tests/integration/test_wild.py:99:1: E303 too many blank lines (3)
```

hugovk:
Whoops! Fixed!

.travis.yml
hugovk: I'd suggest just one, and whichever is the latest. The build matrix is already huge and takes hours to run.

In fact, they're both the same at the moment:

`pypy` => PyPy 5.8.0
`pypy2.7-5.8.0` => PyPy 5.8.0

I would probably go for just `pypy`, as it will be upgraded at some point.

graingert: oh cool, I thought pypy was still the built in Ubuntu one. OK I'll go with your judgment on this.

hartwork:
From a quick look at the Python 3.5 failures, I'm unsure what they are about. Any ideas?

@hugovk, for the eventual merge it would be cool to have these commits squashed (using `git rebase -i`) for cleaner history:
* `Remove dependency on bugs.python.org/xmlrpc: 301 Moved Permanently`
* `Use try/finally for proxy_process.start()/terminate()`

And these two dropped:
* `Ignore IDE metadata`
* `Revert "Ignore IDE metadata"`

Thanks!

hugovk:
@hartwork Rebased!

hugovk:
> From a quick look at the Python 3.5 failures, I'm unsure what they are about. Any ideas?

It's only affecting builds with `TOX_SUFFIX="requestsXX"` and looks like it's from some external (dependency?) change.

My last green commit was on 8th December:
https://travis-ci.org/hugovk/vcrpy/builds/328963247

The next commit was 11th January, just reverting .gitignore, but failed for 3.5:
https://travis-ci.org/hugovk/vcrpy/builds/327579092

Pushing that last green commit in a new branch also fails for 3.5 today:
https://travis-ci.org/hugovk/vcrpy/builds/328963247

graingert:
it works for pypy3.5 which is odd.

graingert:
@hugovk how about a skipif on test_post_chunked_binary_secure if we're on cpython 3.5

graingert:
`pytest.mark.xfail((3, 5) > sys.version_info < (3, 6) and platform.python_implementation() == 'CPython', fails on CPython 3.5')`

tests/integration/test_requests.py
```diff
@@ -114,7 +114,7 @@ def test_post_chunked_binary(tmpdir, httpbin):
     assert req1 == req2
 
 
-@pytest.mark.xfail('sys.version_info >= (3, 6)', strict=True, raises=ConnectionError)
+@pytest.mark.xfail('sys.version_info >= (3, 5)', strict=True, raises=ConnectionError)
```
graingert: `pytest.mark.xfail(sys.version_info >= (3, 6) and platform.python_implementation() == 'CPython', strict=True, ...)`

graingert: because it works fine on pypy

tests/integration/test_requests.py
```diff
@@ -114,7 +114,8 @@ def test_post_chunked_binary(tmpdir, httpbin):
...
+@pytest.mark.xfail((3, 5) > sys.version_info < (3, 6) and platform.python_implementation() == 'CPython', 'Fails on CPython 3.5')
```
graingert: probably need to import these if platform doesn't work in the pytest.mark strings.

.travis.yml
hugovk: @graingert I've removed `pypy2.7-5.8.0` and left the more general `pypy`. 

Should `pypy3.5-5.9.0` also be excluded from `TOX_SUFFIX="flakes"`? Everything else but 3.6 is. 

`pypy` is also excluded from `TOX_SUFFIX="aiohttp"`. Should `pypy3.5-5.9.0` be excluded too?

graingert: > Should pypy3.5-5.9.0 be excluded too?

pypy3.5 should support aiohttp.

hugovk: Yep. And pypy3.5 with flake?

graingert:
Does it pass?

hugovk:
pypy3.5 with flake does pass, but I noticed everything else (except 3.6) is excluded for flake, so was wondering if the original idea was only to run on one version.

Hang on, no need to wonder, git blame confirms it: 
https://github.com/kevin1024/vcrpy/commit/ee17233aa080f3bc1ac8eb3898b3aebc98015047

graingert:
Hmm we should run flake8 on at least a V2 and a V3 of python. Because they
have different rules

hugovk:
Then I'd suggest excluding pypy3.5 and including 2.7.

hartwork:
For [the remaining failure](https://travis-ci.org/kevin1024/vcrpy/jobs/329041188) I think we are dealing with an incompatibility of requests 1.x with Python >=3.4 (including `pypy3.5-5.9.0` in particular). My guess is based on two things:

1. This quote from [`http.client` Python 3 docs](https://docs.python.org/3/library/http.client.html) regarding removal of `strict` parameter:
> Changed in version 3.4: The strict parameter was removed. HTTP 0.9-style “Simple Responses” are not longer supported.

2. These existing excludes from [`.travis.yml` line 49 to 54](https://github.com/kevin1024/vcrpy/blob/master/.travis.yml#L49):
```
matrix:
  [..]
  exclude:
  [..]
    - env: TOX_SUFFIX="requests1"
      python: 3.4
    - env: TOX_SUFFIX="requests1"
      python: 3.5
    - env: TOX_SUFFIX="requests1"
      python: 3.6
  [..]
```

So shall we exclude that case?

graingert:
@hartsock nice research, can we hack it so we don't pass strict along on bad versions?

hartwork:
Unless we replace `VerifiedHTTPSConnection` by something self-made (which I didn't find) it's not us, it's code from the requests module (`requests\packages\urllib3\connectionpool.py`).

hugovk:
I've replaced the flakes test on pypy3.5 with 2.7, so it's now doing flakes on 2.7 and 3.6.

---

In that case, rather than hacking, there's no point testing against Requests 1.x on Python versions that Requests 1.x doesn't support. It makes more sense to exclude.

Or more fundamentally, does VCR.py still need to support Requests 1.x? Does the Requests project still support 1.x? The last 1.x release was v1.2.3, released nearly five years ago in 2013. That's longer than the last CPython 2.6.

hartwork:
I'd be happy either way. @graingert?

graingert:
> I've replaced the flakes test on pypy3.5 with 2.7, so it's now doing flakes on 2.7 and 3.6.

good stuff

graingert:
@hartwork @hugovk let's exclude

```
Python: pypy3.5-5.9.0
TOX_SUFFIX="requests1"
```
for now

then we can drop requests 1 some other time

hugovk:
Excluded!

hugovk:
And two hours later, the CI has turned green!

graingert:
Cool. Can you rebase and clean up commits?

hugovk:
The branch is already up-to-date with master:
```
⌂121% [hugo:/private/tmp/vcrpy] rm-2.6 ± git rebase upstream/master
Current branch rm-2.6 is up to date.
```

How exactly would you like the commits cleaned up?

graingert:
Ok lgtm

kevin1024:
:+1: thanks looks awesome!!

hartwork:
@hugovk thanks for your patience and the great work!

PR #344, opened 2018-01-18
Reduce test matrix from 96 to 30 jobs?

hugovk:
Opening this for discussion.

I'd like to suggest running 96 build jobs is a bit unweildy. It uses around 4.5 hours of CI time in total, about 1 hour from start to finish ([eg.](https://travis-ci.org/hugovk/vcrpy/builds/329772785)).

That's a long time to wait to see if a build passes, and holds up other builds.

It's testing 6 Python versions against 11 Requests versions, 3 urllib3 versions, 2 Tornado versions, 2 boto versions, plus some aiohttp and flake8 jobs.

It looks like new Requests versions were added to the matrix as they were released. I think it's good to take stock from time to time and re-evaluate what is actually important.

* Is it really necessary to test against all those old versions of Requests, urllib3, Tornado and boto? Requests 1.2.3 was released nearly five years ago in 2013, longer than the last CPython 2.6.

* More fundamentally (as touched upon briefly in https://github.com/kevin1024/vcrpy/pull/340#issuecomment-357776765), does VCR.py need to make promises to the community to support so old releases of dependencies, and so many versions of them, or is it acceptable to ask of them to have recent versions?

* Do each of those dependencies make promises to support those old versions?

Further, until recently many of the build jobs were failing, giving less confidence in the CI/tests as a whole.

Finally, https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix says:

> *All build matrixes are currently limited to a maximum of* **200 JOBS** *for both private and public repositories. If you are on an open-source plan, please remember that Travis CI provides this service free of charge to the community. So please only specify the matrix you actually need.*

I'd recommend only testing against a single, recent version of each dependency, 30 jobs ([eg.](https://travis-ci.org/hugovk/vcrpy/builds/329776537)):

| Build | Number of jobs | Total CI time | Start to finish time |
| ----- | -------------- | ------------- | -------------------- |
| Lots of versions | 96 | 4 hrs 21 min 55 sec | 57 min 30 sec |
| Single version   | 30 | 1 hr 14 min 53 sec  | 19 min 17 sec |

Of course, I don't know the full history, so perhaps an extra old version may be necessary, but 96 seems a bit much, especially the 11 versions of Requests.

Thank you!

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = {py27,py35,py36,pypy}-{flakes,requests218,requests216,requests213,requests211,requests27,requests26,requests25,requests24,requests23,requests22,requests1,httplib2,urllib319,urllib3110,urllib3121,tornado3,tornado4,boto,boto3,aiohttp}
+envlist = {py27,py35,py36,pypy}-{flakes,requests27,httplib2,urllib3121,tornado4,boto3,aiohttp}
 
 [testenv:flakes]
 skipsdist = True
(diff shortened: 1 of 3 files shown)


kevin1024:
I’m inclined to agree with Hugo here.  The CI matrix is pretty unwieldy!  Is there any documentation about which versions of requests are supported by the requests team?  Is it only the newest release?

IvanMalison:
I vote for even more aggressive trimming, given that it seems that requests doesn't release new patch versions for old minor versions ever, and they use semantic versioning.

Basically, we should only test the latest versions of requests 1 and requests 2.

hugovk:
I couldn't find any docs about supported Requests releases. I sent a question and will report back if I hear anything.

Latest Requests 1.x and latest Requests 2.x is certainly still a big improvement on 11 versions :)

hugovk:
Update: the Requests team only support the latest release.

https://twitter.com/Lukasaoz/status/954437063675383810

tox.ini
```diff
@@ -18,28 +18,12 @@ deps =
...
-    urllib319: urllib3==1.9.1
-    urllib3110: urllib3==1.10.2
     urllib3121: urllib3==1.21.1
-    {py27,py35,py36,pypy}-tornado3: tornado>=3,<4
     {py27,py35,py36,pypy}-tornado4: tornado>=4,<5
-    {py27,py35,py36,pypy}-tornado3: pytest-tornado
     {py27,py35,py36,pypy}-tornado4: pytest-tornado
-    {py27,py35,py36}-tornado3: pycurl
     {py27,py35,py36}-tornado4: pycurl
-    boto: boto
```
graingert: update the readme to reflect this.

hugovk: Updated.

lamenezes:
@hugovk thanks for your contribution :smile:

PR #345, opened 2018-01-30
Add support for repeated get params in aiohttp

dedsm:
aiohttp supports receiving list of pairs or multidict in case you want to send repeated ids, this PR adds support for this.


lamenezes:
@carsonyl thanks for your PR :slightly_smiling_face: 

can you update your branch? this way we can follow the review/merge of this PR

tox.ini
```diff
@@ -18,6 +18,7 @@ deps =
     pytest
     pytest-httpbin
     PyYAML
+    multidict
```
lamenezes: IMO this requirement is aiohttp only, so maybe it would better to tweak tox.ini by changing this line (as below) and moving it to the end of this section.

```ini
aiohttp: multidict
```

dedsm:
@lamenezes I guess you were referring to me to update the PR and not @carsonyl 

I updated the PR and added the proposed changes.

I played around with dropping the fixed "<3" requirement for aiohttp in the tox.ini but I think it won't play nice with python < 3.6

lamenezes:
> @lamenezes I guess you were referring to me to update the PR and not @carsonyl

Oops. Really sorry about that :sweat_smile: I should pay better attention when replying to multiple PRs

tests/integration/test_aiohttp.py
```diff
@@ -1,3 +1,4 @@
+import multidict
```
lamenezes: I think here we'll have to use `pytest.importorskip('multidict')` to avoid all those CI builds failures. We had to do th same with aiohttp (on line 3)

arthurHamon2:
@dedsm can you rebase your branch on master in order to resolve the conflicts ?
I'll be happy to merge if the CI pass !

dedsm:
@arthurHamon2  I rebased it, hopefully it still stands

PR #346, opened 2018-02-22
Convert extras_require conditional deps to PEP 508 form.

carsonyl:
This replaces `extras_require` with the [PEP 508 spec](https://www.python.org/dev/peps/pep-0508/) for specifying conditional dependencies.

I was motivated to do this while trying to use pipenv and vcrpy with a Python 2/3 project, and found pipenv trying to install yarl's multidict dependency on Python 2. This happened even though pipenv noted that yarl was being skipped for its Python-3-only restriction specified by vcrpy. I'm hoping this change will fix the problem.

See also: pyca/cryptography#4090.


lamenezes:
@carsonyl thanks for your PR :slightly_smiling_face: 

can you update your branch? this way we can follow the review/merge of this PR

carsonyl:
@lamenezes Done! :)

lamenezes:
@carsonyl thx a bunch

PR #353, opened 2018-05-01
Fix aiohttp stub to support version >= 3.1.0

### LINKED ISSUE #348: aiohttp ClientResponse signature has changed
Hi! Firstly, a quick thank you for this library.

I've hit an issue with the aiohttp stubs, following [this aiohttp commit](https://github.com/aio-libs/aiohttp/pull/2820/commits/98ab5cc4ae7bad883c2e3066b5f616741e848a36) which changes the signature of their ClientResponse class.

This breaks lines 56 and 66 in `aiohttp_stubs\.__init__.py`:

`response = MockClientResponse(method, URL(vcr_response.get('url')))`

I only started using vcrpy today, so naturally I assumed I was doing something wrong when I consistently got the following Exception on cassette playback:

`  File "C:\Users\chris.mcevoy\Documents\GitHub\{private repo}\env\lib\site-packages\vcr\stubs\aiohttp_stubs\__init__.py", line 57, in new_request
    response = MockClientResponse(method, URL(url))
TypeError: __init__() missing 8 required keyword-only arguments: 'writer', 'continue100', 'timer', 'request_info', 'auto_decompress', 'traces', 'loop', and 'session'`

For the most part, I think this could be relatively easily worked around by giving vcrpy's MockClientResponse an `__init__()` which calls `super().__init__()` with the old defaults...?

However, there are two new args in aiohttp ClientResponse's signature to consider, specifically `loop` and `session`.

lamenezes:
Also remove python 3.4 async syntax such as `@asyncio.coroutine` and `yield from` as aiohttp now supports only new syntax (`async with` etc.). Update aiohttp stub `MockClientResponse` to match new signature.

Fixes #348.


kevin1024:
Yes help would be great!  I’ll send you an invite.

graingert:
Why doesn't it support `yield from` anymore?

lamenezes:
Because aiohttp dropped support to python 3.4 on aiohttp 3.0 [1].

For example to perform requests using aiohttp now it's mandatory to use `async with` which is supported only on python >= 3.5.

[1] https://github.com/aio-libs/aiohttp/releases/tag/v3.0.1

graingert:
It literally says right there that "`yield from` is still supported"

lamenezes:
I've tested the `yield from` syntax using aiohttp 3.1.1 and it doesn't work:

```python
import asyncio
import aiohttp


@asyncio.coroutine
def aiohttp_request(loop, method, url, output='text', **kwargs):
    with aiohttp.ClientSession(loop=loop) as session:
        response = yield from session.request(method, url, **kwargs)  # NOQA: E999
        if output == 'text':
            content = yield from response.text()  # NOQA: E999
        elif output == 'json':
            content = yield from response.json()  # NOQA: E999
        elif output == 'raw':
            content = yield from response.read()  # NOQA: E999
        return response, content
```

error:

```

______________________________________________________________________________________________________________ test_post[https] ______________________________________________________________________________________________________________

tmpdir = local('/tmp/pytest-of-luiz/pytest-24/test_post_https_0'), scheme = 'https'

    def test_post(tmpdir, scheme):
        data = {'key1': 'value1', 'key2': 'value2'}
        url = scheme + '://httpbin.org/post'
        with vcr.use_cassette(str(tmpdir.join('post.yaml'))):
>           _, response_json = post(url, data=data)

tests/integration/test_aiohttp.py:104:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/integration/test_aiohttp.py:36: in post
    return request('POST', url, output='text', **kwargs)
tests/integration/test_aiohttp.py:28: in request
    return run_in_loop(run)
tests/integration/test_aiohttp.py:21: in run_in_loop
    return loop.run_until_complete(task)
../../.pyenv/versions/3.6.4/lib/python3.6/asyncio/base_events.py:467: in run_until_complete
    return future.result()
tests/integration/aiohttp_utils.py:7: in aiohttp_request
    with aiohttp.ClientSession(loop=loop) as session:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <aiohttp.client.ClientSession object at 0x7f4d6800f438>

    def __enter__(self):
>       raise TypeError(Use async with instead)
E       TypeError: Use async with instead

.tox/py36-aiohttp/lib/python3.6/site-packages/aiohttp/client.py:746: TypeError
```

graingert:
Maybe it's possible if you yield from the dunder methods directly?

lamenezes:
> Maybe it's possible if you yield from the dunder methods directly?

It seems doable, but is it worth the trouble maintaining 3.4 support even if aiohttp itself won't support?

lamenezes:
> It seems doable, but is it worth the trouble maintaining 3.4 support even if aiohttp itself won't support?

Answering my own question: yes, it might be a good a idea to maintain support to python3.4/aiohttp for a little longer because some might still be using latest vcrpy and aiohttp<3.1

PR #354, opened 2018-05-02
Fix broken tests + build quicker

lamenezes:
Many CI builds are currently broken. After some investigation I found the major causes for it are:

- Old python and library versions. Fixed by removing tests from tox/travis (thanks to hugovk on #344)
- Some new releases of test-only libs such as Flask>=1.0 and aiohttp >= 3. Fixed by pinning versions on tox while the support is implemented. 

This PR fixes that :soccer: :top:


kevin1024:
💯

PR #356, opened 2018-05-09
fixes before_record_response mutates response

### LINKED ISSUE #355: before_record_response DOES change real response
### Versions:
python: v3.6
vcrpy: 1.11.1

### Issue:
I tried using the `before_record_response` callback to scrub responses of sensitive auth info before they are written to disc. However, when I have a test that exercises a function that makes 2 requests, the first requesting an auth token, and the second using that auth token, I'm seeing that the actual response from the first request has been mutated.

There was an issue opened previously (https://github.com/kevin1024/vcrpy/issues/291), although a while ago, that asserted the opposite, and @kevin1024 seemed to agree and explain why the response wasn't/shouldn't be changed.

### Repro:
I wrote the following script to reproduce the issue.
```python
import json
import requests
import vcr


VCR = vcr.VCR(
    serializer='yaml',
    cassette_library_dir='cassettes',
    record_mode='once',
)


def filter_json_response_body(fields, replacement='[REDACTED]'):
    """This is a very naive scrubber, for demo purposes.
    """
    def before_record_callback(response):
        string_body = response['body']['string'].decode('utf8')
        body = json.loads(string_body)

        for field in fields:
            if field in body:
                body[field] = replacement

        response['body']['string'] = json.dumps(body).encode()
        return response

    return before_record_callback


with VCR.use_cassette(
    'it_mutates_the_response',
    before_record_response=filter_json_response_body(['text'])
) as cass:

    response = requests.get('http://numbersapi.com/42?json')
    print(response.json())
```

Here is my `requirements.txt`:
```
requests==2.18.4
vcrpy==1.11.1
```

Please let me know if I forgot to include anything!

kgraves:
When no cassette exists, it's expected that the response returned should be
the original, unchanged response. The response recorded in the cassette should be
that which is returned by the before_record_response callback.

But on subsequent requests/responses (when a cassette exists), the responses
returned should be exactly what is in the cassette.

resolves #355

### DIFF
--- a/tests/integration/test_stubs.py
+++ b/tests/integration/test_stubs.py
@@ -1,5 +1,6 @@
 import vcr
 import zlib
+import json
 import six.moves.http_client as httplib
 
 from assertions import assert_is_json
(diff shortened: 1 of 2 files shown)


kgraves:
hmmm that error seems unrelated to my change...

Here's the failure:
```
=================================== FAILURES ===================================
_______________________ test_post_chunked_binary_secure ________________________
[XPASS(strict)] condition: sys.version_info >= (3, 6)
----------------------------- Captured stdout call -----------------------------
b'<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>501 Not Implemented</title>\n<h1>Not Implemented</h1>\n<p>Chunked requests are not supported for server WSGIServer/0.2</p>\n'
----------------------------- Captured stderr call -----------------------------
127.0.0.1 - - [09/May/2018 21:29:49] "POST /post HTTP/1.1" 501 184
------------------------------ Captured log call -------------------------------
connectionpool.py          272 WARNING  Connection pool is full, discarding connection: 127.0.0.1
connectionpool.py          272 WARNING  Connection pool is full, discarding connection: 127.0.0.1
```

lamenezes:
@kgraves thx for your PR! it really does make sense your proposal :slightly_smiling_face: 

> hmmm that error seems unrelated to my change...

Just now I fixed that error, updating the branch should solve the failure

kgraves:
I should probably change my test to use httpbin, right?

lamenezes:
@kevin1024 would you please check this PR and the related issue #355? I reviewed the code and it's all good, but I'm not sure if this is a bug or the expected/correct behavior. Your feedback is much appreciated.

kevin1024:
Yeah, the PR looks good and I think this is probably the right behavior, but I am a bit worried about breaking compatibility in case anyone was depending on this.

I think maybe we should do a major version bump just in case.  What do you think?

lamenezes:
> I think maybe we should do a major version bump just in case. What do you think?

Looks good to me. As we have some bugfixes on master I think it would be better to release a new version before merging this PR and then releasing a major version.

PR #361, opened 2018-06-18
Preserve existing query parameters in request url

gciding:
Currently when running VCRpy over a project using aiohttp, the cassettes lose the query parameters that existed in the urls.  

E.g. A request for `https://www.kogan.com/au/?name=hello` is saved in the cassette as `https://www.kogan.com/au/`.  This causes problems when you are testing multiple urls that have different query parameters on the same host.

This crude change preserves the original query parameters in the request_url so that the corresponding cassettes don't lose their query strings.


lamenezes:
@gciding great PR!

I've checked here and that really is a bug on the aiohttp stub. I'm taking the liberty to close this PR and use your commits and open a new PR with tests so we can merge this.

Thanks for your contribution :)

PR #371, opened 2018-07-09
Fix aiohttp patch to work with aiohttp >= 3.3

### LINKED ISSUE #367: Failing with aiohttp 3.3.2
Tried against vcrpy master 

```python
import pytest
import aiohttp.web

import vcr

async def hello(request):
    return aiohttp.web.Response(text='Hello, world')


@pytest.fixture
async def client(aiohttp_client):
    app = aiohttp.web.Application()
    app.router.add_get('/', hello)
    client = await aiohttp_client(app)
    return client


@vcr.use_cassette('localhost')
async def test_vcrpy(client): 
    await client.get('/')
```
When run raises

```
test_aiohttp.py:20: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../.virtualenvs/test/lib/python3.6/site-packages/aiohttp/test_utils.py:217: in request
    method, self.make_url(path), *args, **kwargs
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <aiohttp.client._RequestContextManager object at 0x7f53daa3cfc0>

    def __await__(self):
>       ret = self._coro.__await__()
E       AttributeError: 'generator' object has no attribute '__await__'
```

I managed to fix this for me by dropping Python 3.4 support and changed `vcr.stubs.aiohttp.__init__.vcr_request.new_requst` to use `async` and `await` syntax.

Tested against vcrpy==master, Python 3.6.5, aiohttp 3.3.2

stj:
Aiohttp expects an awaitable instance to be returned from
`ClientSession._request` though `asyncio.coroutine` decorated function
do not implement `__await__`. By changing the syntax and dropping Python
3.4 support we fix this issue.

Resolves #367


lamenezes:
@stj thanks for your PR. 

can you check and fix the errors on CI? if you need any help just ask :)

stj:
Fixed the issues with Python 3.5 and 3.6. No clue how to tackle the SyntaxError for 2.7 and what is going on with PyPy.

lamenezes:
@stj I'm still investigating this issue. vcrpy works fine with `aiohttp.ClientSession`, but for some strange reason it doesn't work with `aiohttp.test_utils.TestClient` that is being used on the test of this PR. I'm trying to figure out why that happens before dropping support on python 3.4.

stj:
When I looked at this I figured that it had to do with [`TestClient.request`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py#L216). The `session.request` returns a `_RequestContextManager`, if you change that to the async method `session._request` the test passes.

With that knowledge I looked into the `_RequestContextManager` class and figuring out that in line [851](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L851) the result of `__await__` is returned. Methods decorated with `asyncio.coroutine` do not implement `__await__` only `async def` methods do.

Maybe that helps you understand my changes.

lamenezes:
ok, I won't spend much time analyzing this trying to find a workaround to this bug that works with python 3.4. since we're releasing a major version on vcrpy and IMHO think it's time to drop support on python 3.4 as not even aiohttp supports it anymore.

stj:
Figured out how to have aiohttp tests ignored in py27 and pypy.
Flakes still fails against py27 and I haven't figure out how to configure tox to ignore files when run with py27.

stj:
Forgot about my PR here ... took some time to get the build green today.

I think I could add Python 3.4 without async support in the same way 2.7 is supported. Could look into it if there is a strong desire to still support it.

lamenezes:
@stj thanks for your contribution. I'm closing this PR as I moved the code from here to #390 with some other updates.

PR #374, opened 2018-07-18
Update docs' lists of supported HTTP clients

adamchainz:
I noticed these were out of sync, especially weirdly one mentioning boto and the other mentioning boto3. I figure the list in the README is redundant when the Installation docs section lists the supported libraries, so I've removed it. I also:

* alphabetically sorted the list
* Highlighted the library names as code
* Added both `boto` and `boto3` to the list since there is support for both
* Removed the comment about Tornado's AsyncHTTPClient since that's an implementation detail plus the patch happens on a couple different classes
* Removed the note about `http.client` being part of Python 3, because everyone is Python 3 these days anyway :)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -20,7 +20,7 @@ The following HTTP libraries are supported:
 -  ``http.client``
 -  ``httplib2``
 -  ``requests`` (both 1.x and 2.x versions)
--  ``tornado``
+-  ``tornado.httpclient``
 -  ``urllib2``
 -  ``urllib3``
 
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -12,15 +12,17 @@ Compatibility
...
--  Tornado's AsyncHTTPClient
+The following HTTP libraries are supported:
+
+-  ``aiohttp``
+-  ``boto``
+-  ``boto3``
+-  ``http.client``
+-  ``httplib2``
+-  ``requests`` (both 1.x and 2.x versions)
+-  ``tornado``
```
graingert: `tornado.httpclient`

lamenezes:
@adamchainz thanks for your contribution!

PR #319, opened 2017-06-22
Fix host and port for proxy connections

samuelfekete:
This fixes errors when using a proxy, as described in this issue:
https://github.com/kevin1024/vcrpy/issues/214

PR #389, opened 2018-09-18
Fix proxy

lamenezes:
Improve tests on #319. Thanks @samuelfekete @daneoshiga.

### DIFF
--- a/tests/integration/test_proxy.py
+++ b/tests/integration/test_proxy.py
@@ -51,5 +51,10 @@ def proxy_server():
 def test_use_proxy(tmpdir, httpbin, proxy_server):
     '''Ensure that it works with a proxy.'''
     with vcr.use_cassette(str(tmpdir.join('proxy.yaml'))):
-        requests.get(httpbin.url, proxies={'http': proxy_server})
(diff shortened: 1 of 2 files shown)


sathieu:
It looks like this commit broke the pyvmomi testsuite. See https://github.com/vmware/pyvmomi/pull/736#issuecomment-439176546.

sathieu:
The root cause is probably:
```
>>> from yarl import URL
>>> str(URL("foo:8009").port)
'None'
>>> str(URL("//foo:8009").port)
'8009'
```
And [http.client](https://docs.python.org/3/library/http.client.html#module-http.client) accepts `foo:8009`.

sathieu:
[this](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) is more precise:
> Following the syntax specifications in RFC 1808, urlparse recognizes a netloc only if it is properly introduced by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component.

sathieu:
The hostport logic of http.client is here:
https://github.com/python/cpython/blob/3.7/Lib/http/client.py#L871,L889

PR #397, opened 2018-09-26
ClientResponse.release isn't a coroutine

stj:
Therefore it should not be one in the MockClientResponse class.

https://github.com/aio-libs/aiohttp/blob/d0af887e3121d677b32339adff6540b6de01d167/aiohttp/client_reqrep.py#L832


lamenezes:
@stj thanks!

Issue #396, opened 2018-09-25
tornado_exception_can_be_caught fails in py3.7


P-EB:
Hi,

It seems that there are still some small issues in python3.7 with tornado.

This is the output I get for test_tornado_exception_can_be_caught.

```
tests/integration/test_tornado.py::test_tornado_exception_can_be_caught[simple] FAILED [ 50%]

=================================== FAILURES ===================================
_________________ test_tornado_exception_can_be_caught[simple] _________________

get_client = <function get_client.<locals>.<lambda> at 0x7fb3f890aea0>

    @pytest.mark.gen_test
    @vcr.use_cassette(path_transformer=vcr.default_vcr.ensure_suffix('.yaml'))
    def test_tornado_exception_can_be_caught(get_client):
        try:
            yield get(get_client(), 'http://httpbin.org/status/500')
        except http.HTTPError as e:
            assert e.code == 500
    
        try:
>           yield get(get_client(), 'http://httpbin.org/status/404')

tests/integration/test_tornado.py:337: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <vcr.cassette.CassetteContextDecorator object at 0x7fb3f8a81c88>
fn = <function CassetteContextDecorator._execute_function.<locals>.handle_function at 0x7fb3f8b252f0>

    def _handle_generator(self, fn):
        """Wraps a generator so that we're inside the cassette context for the
            duration of the generator.
            """
        with self as cassette:
            coroutine = fn(cassette)
            # We don't need to catch StopIteration. The caller (Tornado's
            # gen.coroutine, for example) will handle that.
            to_yield = next(coroutine)
            while True:
                try:
>                   to_send = yield to_yield

vcr/cassette.py:137: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tornado.gen.Runner object at 0x7fb3f7fe0828>

    def run(self):
        """Starts or resumes the generator, running until it reaches a
            yield point that is not ready.
            """
        if self.running or self.finished:
            return
        try:
            self.running = True
            while True:
                future = self.future
                if not future.done():
                    return
                self.future = None
                try:
                    orig_stack_contexts = stack_context._state.contexts
[...]

PR #403, opened 2018-10-14
Uses yield from in _handle_generator instead of a lookalike that doesn't work anymore

P-EB:
Dear @kevin1024,

PEP479 is active in python3.7 now, and this breaks _handle_generator. I've filed an issue to report this, and worked on finding a fix.

Essentially, this fix consists in using yield from coroutine in _handle_generator, but as this pattern is not python2.7 compliant, I designed a full patch.

This PR would close issue #396

### DIFF
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -25,6 +25,10 @@ if sys.version_info[:2] >= (3, 5):
 else:
     def handle_coroutine(*args, **kwags):
         raise NotImplementedError('Not implemented on Python 2')
+if sys.version_info[:2] >= (3, 5):
+    from .cassette_yf import _loc_handle_generator
+else:
+    from .cassette_nyf import _loc_handle_generator
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


vcr/cassette_yf.py
```diff
@@ -0,0 +1,7 @@
+def _loc_handle_generator(self, fn):
+    """Wraps a generator so that we're inside the cassette context for the
+    duration of the generator.
+    """
+    with self as cassette:
+        coroutine = fn(cassette)
```
graingert: Any reason you don't: `yield from fn(cassette)`

P-EB: For readability mostly.

About test: this code is directly used by plenty tests so I'm not sure to follow what you'd like.

graingert: unless trying to fit into 80 cols it's not worth breaking lines

graingert:
it looks like the linter is failing

vcr/cassette.py
```diff
@@ -25,6 +25,10 @@ def iscoroutinefunction(*args, **kwargs):
 else:
     def handle_coroutine(*args, **kwags):
         raise NotImplementedError('Not implemented on Python 2')
```
graingert: actually I think these version checks should be feature detected eg:
```
try:
    from ._handle_coroutine import handle_coroutine
except ImportError:
    def handle_coroutine(*args, **kwargs):
        raise NotImplementedError('Not implemented on Python 2')

try:
    from ._handle_generator_native import handle_generator
except ImportError:
    from ._handle_generator_emulation import handle_generator
```

vcr/cassette.py
```diff
@@ -123,25 +127,7 @@ def handle_function(cassette):
 
         return self._handle_function(fn=handle_function)
 
-    def _handle_generator(self, fn):
```
graingert: what does `_loc` mean?

graingert: also no need to expose this as a method - follow handle_coroutine's implementation

graingert:
@P-EB few change requests here

P-EB:
@graingert I changed my approach. Essentially I just added a catch of StopIteration in the handle_generator method. Do you wish me to post it?

PR #411, opened 2018-11-26
Fixes #382 boto3 compatibility.

### LINKED ISSUE #382: boto3 integration broken
Hey,
I've recently started seeing boto3 related failures in my unit tests. The recordings had been working previously, but it seems that something has changed in boto, and the vcrpy support is no longer working. 

```
------------------------------------------------------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------------------------------------------------------
/Users/analogj/repos/lexicon/venv/lib/python2.7/site-packages/urllib3/connectionpool.py:857: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
```

1oglop1:
As mentioned in #382 boto3 is not using vendored requests anymore.

### DIFF
--- a/tests/integration/test_boto3.py
+++ b/tests/integration/test_boto3.py
@@ -1,15 +1,46 @@
 import pytest
+import os
+
 boto3 = pytest.importorskip("boto3")
 
 import boto3  # NOQA
+import botocore  # NOQA
 import vcr  # NOQA
 
-bucket = 'boto3-demo-1337'              # a bucket you can access
-key = 'test/my_test.txt'                # key with r+w access
-content = 'hello world i am a string'   # content to put in the test file
+ses = boto3.Session(
+    aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
+    aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
+    aws_session_token=None,
+    region_name=os.environ['AWS_DEFAULT_REGION'],
+    # botocore_session=None,
+    # profile_name=None
+)
+
+IAM_CLIENT = ses.client('iam')
+
+try:
+    from botocore import awsrequest  # NOQA
+
+    botocore_awsrequest = True
+except ImportError:
+    botocore_awsrequest = False
 
 
-def test_boto_stubs(tmpdir):
+# skip tests if boto does not use vendored requests anymore
+# https://github.com/boto/botocore/pull/1495
+boto3_skip_vendored_requests = pytest.mark.skipif(
+    botocore_awsrequest,
+    reason='botocore version {ver} does not use vendored requests anymore.'.format(
+        ver=botocore.__version__))
+
+boto3_skip_awsrequest = pytest.mark.skipif(
+    not botocore_awsrequest,
+    reason='botocore version {ver} still uses vendored requests.'.format(
+        ver=botocore.__version__))
+
+
+@boto3_skip_vendored_requests
+def test_boto_vendored_stubs(tmpdir):
     with vcr.use_cassette(str(tmpdir.join('boto3-stubs.yml'))):
         # Perform the imports within the patched context so that
         # HTTPConnection, VerifiedHTTPSConnection refers to the patched version.
@@ -23,45 +54,59 @@ def test_boto_stubs(tmpdir):
         VerifiedHTTPSConnection('hostname.does.not.matter')
 
 
+@boto3_skip_awsrequest
+def test_boto3_awsrequest_stubs(tmpdir):
+    with vcr.use_cassette(str(tmpdir.join('boto3-stubs.yml'))):
+        from botocore.awsrequest import AWSHTTPConnection, AWSHTTPSConnection
+        from vcr.stubs.boto3_stubs import VCRRequestsHTTPConnection, VCRRequestsHTTPSConnection
+        assert issubclass(VCRRequestsHTTPConnection, AWSHTTPConnection)
+        assert issubclass(VCRRequestsHTTPSConnection, AWSHTTPSConnection)
+        AWSHTTPConnection('hostname.does.not.matter')
+        AWSHTTPSConnection('hostname.does.not.matter')
+
+
 def test_boto3_without_vcr():
-    s3_resource = boto3.resource('s3')
-    b = s3_resource.Bucket(bucket)
-    b.put_object(Key=key, Body=content)
+    username = 'user'
+    response = IAM_CLIENT.get_user(UserName=username)
 
-    # retrieve content to check it
-    o = s3_resource.Object(bucket, key).get()
-
-    # decode for python3
-    assert content == o['Body'].read().decode('utf-8')
+    assert response['User']['UserName'] == username
 
 
 def test_boto_medium_difficulty(tmpdir):
-    s3_resource = boto3.resource('s3')
-    b = s3_resource.Bucket(bucket)
+    username = 'user'
+
     with vcr.use_cassette(str(tmpdir.join('boto3-medium.yml'))):
-        b.put_object(Key=key, Body=content)
-        o = s3_resource.Object(bucket, key).get()
-        assert content == o['Body'].read().decode('utf-8')
+        response = IAM_CLIENT.get_user(UserName=username)
+        assert response['User']['UserName'] == username
 
     with vcr.use_cassette(str(tmpdir.join('boto3-medium.yml'))) as cass:
-        b.put_object(Key=key, Body=content)
-        o = s3_resource.Object(bucket, key).get()
-        assert content == o['Body'].read().decode('utf-8')
+        response = IAM_CLIENT.get_user(UserName=username)
+        assert response['User']['UserName'] == username
         assert cass.all_played
 
 
 def test_boto_hardcore_mode(tmpdir):
+    username = 'user'
     with vcr.use_cassette(str(tmpdir.join('boto3-hardcore.yml'))):
-        s3_resource = boto3.resource('s3')
-        b = s3_resource.Bucket(bucket)
-        b.put_object(Key=key, Body=content)
-        o = s3_resource.Object(bucket, key).get()
-        assert content == o['Body'].read().decode('utf-8')
+        ses = boto3.Session(
+            aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
+            aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
+            region_name=os.environ['AWS_DEFAULT_REGION'],
+        )
+
+        iam_client = ses.client('iam')
+        response = iam_client.get_user(UserName=username)
+        assert response['User']['UserName'] == username
 
     with vcr.use_cassette(str(tmpdir.join('boto3-hardcore.yml'))) as cass:
-        s3_resource = boto3.resource('s3')
-        b = s3_resource.Bucket(bucket)
-        b.put_object(Key=key, Body=content)
-        o = s3_resource.Object(bucket, key).get()
-        assert content == o['Body'].read().decode('utf-8')
+        ses = boto3.Session(
+            aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
+            aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
+            aws_session_token=None,
+            region_name=os.environ['AWS_DEFAULT_REGION'],
+        )
+
+        iam_client = ses.client('iam')
+        response = iam_client.get_user(UserName=username)
+        assert response['User']['UserName'] == username
         assert cass.all_played
--- a/tests/integration/test_boto3.py
+++ b/tests/integration/test_boto3.py
@@ -17,6 +17,7 @@ ses = boto3.Session(
 )
 
 IAM_CLIENT = ses.client('iam')
+IAM_USER_NAME = 'vcrpy'
 
 try:
     from botocore import awsrequest  # NOQA
@@ -66,27 +67,24 @@ def test_boto3_awsrequest_stubs(tmpdir):
 
 
 def test_boto3_without_vcr():
-    username = 'user'
-    response = IAM_CLIENT.get_user(UserName=username)
+    response = IAM_CLIENT.get_user(UserName=IAM_USER_NAME)
 
-    assert response['User']['UserName'] == username
+    assert response['User']['UserName'] == IAM_USER_NAME
 
 
 def test_boto_medium_difficulty(tmpdir):
-    username = 'user'
 
     with vcr.use_cassette(str(tmpdir.join('boto3-medium.yml'))):
-        response = IAM_CLIENT.get_user(UserName=username)
-        assert response['User']['UserName'] == username
+        response = IAM_CLIENT.get_user(UserName=IAM_USER_NAME)
+        assert response['User']['UserName'] == IAM_USER_NAME
 
     with vcr.use_cassette(str(tmpdir.join('boto3-medium.yml'))) as cass:
-        response = IAM_CLIENT.get_user(UserName=username)
-        assert response['User']['UserName'] == username
+        response = IAM_CLIENT.get_user(UserName=IAM_USER_NAME)
+        assert response['User']['UserName'] == IAM_USER_NAME
         assert cass.all_played
 
 
 def test_boto_hardcore_mode(tmpdir):
-    username = 'user'
     with vcr.use_cassette(str(tmpdir.join('boto3-hardcore.yml'))):
         ses = boto3.Session(
             aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
@@ -95,8 +93,8 @@ def test_boto_hardcore_mode(tmpdir):
         )
 
         iam_client = ses.client('iam')
-        response = iam_client.get_user(UserName=username)
-        assert response['User']['UserName'] == username
+        response = iam_client.get_user(UserName=IAM_USER_NAME)
+        assert response['User']['UserName'] == IAM_USER_NAME
 
     with vcr.use_cassette(str(tmpdir.join('boto3-hardcore.yml'))) as cass:
         ses = boto3.Session(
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -11,3 +11,5 @@ pytestdebug.log
 
 fixtures/
 /docs/_build
+
+.idea/
```
adamchainz: your own IDE files shouldn't be added to the gitignore of every project, instead configure a global gitignore

https://help.github.com/articles/ignoring-files/

vcr/patch.py
```diff
@@ -6,21 +6,23 @@
...
+log = logging.getLogger(__name__)
 # Save some of the original types for the purposes of unpatching
 _HTTPConnection = httplib.HTTPConnection
 _HTTPSConnection = httplib.HTTPSConnection
 
 
 # Try to save the original types for boto3
 try:
-    import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+    from botocore.awsrequest import AWSHTTPSConnection, AWSHTTPConnection
```
adamchainz: I think we also need to try the `vendored` imports for users on older boto3 versions

```
try:
    from botocore.awsrequest import AWSHTTPSConnection, AWSHTTPConnection
except ImportError:
    try:
        import botocore.vendored.requests.packages.urllib3.connectionpool as cpool  
    except ImportError:  # pragma: no cover
        pass
    else:
        _Boto3VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
        _cpoolBoto3HTTPConnection = cpool.HTTPConnection        
        _cpoolBoto3HTTPSConnection = cpool.HTTPSConnection
else:
    _Boto3VerifiedHTTPSConnection = AWSHTTPSConnection
    _cpoolBoto3HTTPConnection = AWSHTTPConnection
    _cpoolBoto3HTTPSConnection = AWSHTTPSConnection
```

vcr/patch.py
```diff
@@ -184,13 +186,18 @@ def _requests(self):
...
         return self._urllib3_patchers(cpool, requests_stubs)
 
+    @_build_patchers_from_mock_triples_decorator
     def _boto3(self):
         try:
-            import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+            import botocore.awsrequest as cpool
         except ImportError:  # pragma: no cover
-            return ()
+            yield ()
```
adamchainz: same here, support old and new imports?

vcr/patch.py
```diff
@@ -406,23 +413,24 @@ def reset_patchers():
             yield mock.patch.object(cpool.HTTPConnectionPool, 'ConnectionCls', _cpoolHTTPConnection)
             yield mock.patch.object(cpool.HTTPSConnectionPool, 'ConnectionCls', _cpoolHTTPSConnection)
 
+    # FIXME this is not right probably
     try:
-        import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+        import botocore.awsrequest as cpool
```
adamchainz: same here, old and new imports, plus that FIXME?

.gitignore
1oglop1: hmm good idea, thank you

vcr/patch.py
1oglop1: We can check for the `boto3` version and act based on that information.
Would it be good solution?
But the problem is possibly on the side of tests
which are pinned to the old version `requests27: requests==2.7.0` whereas `vendored` usually uses newer versions

vcr/patch.py
1oglop1: Fix me is there because I wasn't really sure it works this way.
And I need more details on how this works. 
However it behaved correctly with latest boto3 versions

vcr/patch.py
adamchainz: You could check for version, although it's normally better to check for behaviours rather than versions (behaviour could be reversed in a later version), hence my suggestion to use `try/except`.

Tests should probably be updated to use new requests

1oglop1: Btw isn't there any better way then having all these imports on 3 places?
And implement some sort of plugin system? 
https://github.com/Rockhopper-Technologies/pluginlib

tests/integration/test_boto3.py
```diff
@@ -23,6 +47,19 @@ def test_boto_stubs(tmpdir):
...
 
 
+@boto3_vendored
+def test_boto3_awsrequest_stubs(tmpdir):
+    with vcr.use_cassette(str(tmpdir.join('boto3-stubs.yml'))):
+        from botocore.awsrequest import AWSHTTPConnection, AWSHTTPSConnection
+        from vcr.stubs.boto3_stubs import VCRRequestsHTTPConnection, VCRRequestsHTTPSConnection
+        # addopted from test_boto_vendored_stubs, but
+        # Todo this seems to be other way around
+        assert issubclass(AWSHTTPConnection, VCRRequestsHTTPConnection)
```
1oglop1: I wonder if the logic should not be other way around? (also in fn above)
issubclass(VCRRequestsHTTPConnection. AWSHTTPConnection)

1oglop1:
Most tests are now passed but AWS credentials seems to be missing or so, 
https://travis-ci.org/kevin1024/vcrpy/jobs/476016137
I'd welcome comments @adamchainz

jtratner:
Hi @adamchainz and @1oglop1  - we're super interested in getting later version of boto3 to work - is there anything we can do to help get this merged into the main vcrpy? :)

adamchainz:
I'm not a maintainer and I don't have time to review this now, sorry.

1oglop1:
@jtratner if you could review the code and set up test env to make sure that tests will pass that would be great.

I exchanged few emails with @kevin1204 so I might be able to get permissions to this repo. If you are under some time pressure we could move the work to my fork in the meantime this gets settled. Let me know what would be better for you.

kevin1024:
Yup, I’m happy to bring on a new maintainer!

1oglop1:
I should be able to provide cloud test env.

1oglop1:
@jtratner
I'm  looking at current tests, it does `put_object`, I was thinking if is a good idea to set up public bucket instead and just test with `get_object` and `list_bucket` instead. 
So the testing bucket can contain just 1 tiny file. 

I already have such a bucket and I'll work on tests bit later

1oglop1:
update: there seems to be a problem with travis now: :/ https://travis-ci.org/1oglop1/vcrpy/builds/490307330

https://travis-ci.community/t/unable-to-download-python-3-7-archive-on-travis-ci/639

1oglop1:
@kevin1024 I seem to get blocked by this: https://github.com/kevin1024/pytest-httpbin/issues/52 do you have some time to fix it? It does not seem to be much work according to comment from link inside the issue:

> Python 3.6 is a little more forgiving than Python 3.7. Python 3.7 uses OpenSSL's hostname verification algorithms, which interpret the RFCs more strictly. You have to include a SAN field of type IP address. Matching against CN has been deprecated for more than 15 years, see https://bugs.chromium.org/p/chromium/issues/detail?id=308330

kevin1024:
OK, I believe the underlying pytest-httpbin issue should be resolved.

1oglop1:
First results from local tests
`pytest tests/integration/test_boto3.py -v -s`
```
tests/integration/test_boto3.py::test_boto_vendored_stubs SKIPPED
tests/integration/test_boto3.py::test_boto3_awsrequest_stubs PASSED
tests/integration/test_boto3.py::test_boto3_without_vcr PASSED
tests/integration/test_boto3.py::test_boto_medium_difficulty PASSED
tests/integration/test_boto3.py::test_boto_hardcore_mode PASSED
```

However it seems tox might have some problems or I might be missing something
`tox -e py37-boto3`
seems to be affeced by tox-dev/tox/issues/809

vcr/patch.py
```diff
@@ -184,13 +189,27 @@ def _requests(self):
...
+                import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+            except ImportError:  # pragma: no cover
+                pass
+            else:
+                from .stubs import boto3_stubs
+                yield self._urllib3_patchers(cpool, boto3_stubs)
+        else:
+            from .stubs import boto3_stubs
+            log.debug("Patching boto3 cpool with %s", cpool)
+            print('pipip')
```
yardensachs: cleanup?

1oglop1: In progress ;)
I might be able to push at the end of the week or so

fcurella:
@1oglop1 anything I can do to keep the ball rolling? I have time to volunteer if you need any help

1oglop1:
@fcurella That would be nice if you could help!
If you have any idea how to run the tests with travis/tox
here is my testing branch: https://github.com/jans-forks/vcrpy/tree/travis-test
I was actually thinking about suggesting different CI and build the CI matrix with GitlabCI.

You can reach out to me via gitter for more info. My timezone is CEST

fcurella:
I think there's something off with Travis. I've submitted a PR (https://github.com/kevin1024/vcrpy/pull/433)  but it never kicked a build.

vcr/patch.py
pawciobiel: pluginlib is cool but it may be an overkill for this

vcr/patch.py
```diff
@@ -406,23 +413,24 @@ def reset_patchers():
             yield mock.patch.object(cpool.HTTPConnectionPool, 'ConnectionCls', _cpoolHTTPConnection)
             yield mock.patch.object(cpool.HTTPSConnectionPool, 'ConnectionCls', _cpoolHTTPSConnection)
 
+    # FIXME this is not right probably
```
pawciobiel: Why would this be not right?

vcr/stubs/boto3_stubs.py
```diff
@@ -1,11 +1,22 @@
...
+from botocore.awsrequest import AWSHTTPConnection as HTTPConnection
+from botocore.awsrequest import AWSHTTPSConnection as VerifiedHTTPSConnection
+
 from ..stubs import VCRHTTPConnection, VCRHTTPSConnection
 
+
 # urllib3 defines its own HTTPConnection classes, which boto3 goes ahead and assumes
 # you're using.  It includes some polyfills for newer features missing in older pythons.
 
+# try:
```
pawciobiel: Why do we need this bit?

pawciobiel:
I used it with a test for `upload_fileobj` and it worked.

1oglop1:
I finally found out what's breaking boto3 integration tests

`tox.ini` runs `runtests.sh` https://github.com/jans-forks/vcrpy/blob/fix/382-boto3/tox.ini

which does only `REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` py.test $*` -but `REQUESTS_CA_BUNDLE` breaks the boto3 tests

I need to find the way to run boto3 tests withtout setting `REQUESTS_CA_BUNDLE`

pawciobiel:
@1oglop1 Could this be caused by the version of `certifi` package?
I think it may be possible to unset that variable in in tox.ini, eg:
```
[testenv:boto3]
    setenv REQUESTS_CA_BUNDLE = 
```

1oglop1:
I guess that I finally fixed the tests

1oglop1:
@pawciobiel 
good news. most tests are passing, but `urllib3` still gives some errors

https://travis-ci.org/jans-forks/vcrpy/jobs/533145126

pawciobiel:
@1oglop1 try adding 
```
cert_reqs='CERT_NONE',  assert_hostname=False
```
to https://github.com/kevin1024/vcrpy/blob/master/tests/integration/test_urllib3.py#L23

https://stackoverflow.com/a/18062293

1oglop1:
@pawciobiel Thanks for the prompt response it worked like a charm for py3!
https://travis-ci.org/jans-forks/vcrpy/builds/533280216

Now tests are failing for legacy python only!

pawciobiel:
@1oglop1 idk, I find src/urllib3/connection.py:370 confusing :D

I wonder if passing `check_hostname=False` and `assert_hostname=False` or assert_hostname=='127.0.0.1' into https://github.com/kevin1024/vcrpy/blob/master/tests/integration/test_urllib3.py#L18
would help?

svdgraaf:
I can confirm that this branch works with my 100+ unittests 👍🏻 Would be awesome if we could get this merged...

kevin1024:
Happy to merge if we can get tests passing

1oglop1:
@kevin1024 I'm not sure it's worth the effort to support anything related to legacy python since
 at least 1/2 of upstream libraries are dropping its support as well.

https://pythonclock.org

eg:
`boto` - deprecated in favour of `boto3`
`aiohttp` - https://github.com/aio-libs/aiohttp/releases/tag/v3.0.1

other libs https://python3statement.org including `requests`.

Also it might be beneficial to start thinking of modularity, because `requests3` is planning to drop `urllib3` dependency - but this is sound of distant (not so distant) future.

arthurHamon2:
@1oglop1 I just saw your thread and I did not realize that you were trying to solve the same issues with the tests that I have encountered. I did a PR here https://github.com/kevin1024/vcrpy/pull/411 to update the unit tests and make them work with the latest dependencies, including Python 2.7, does it help this PR to be merged ?

jeshan:
> I did a PR here #411 
I think Arthur meant, #441 and not #411 which is the current one.

arthurHamon2:
> > I did a PR here #411
> 
> I think Arthur meant #441 and not #411 which is the current one.

Thanks for noticing this typo, I have updated my original comment.

1oglop1:
@arthurHamon2 I'll try to find some time implement the changes in my branch.

arthurHamon2:
@1oglop1 you are almost there !
You only have a style issue : 
```
./tests/unit/test_migration.py:14:5: F811 redefinition of unused 'Loader' from line 9
```

Any chance you can clean your commit history by removing all unnecessary commits ?
And why do you still working on this PR as this PR https://github.com/kevin1024/vcrpy/pull/448 is ready to be merged ? Did I miss something ?

1oglop1:
@arthurHamon2 Yes you did, all boto3 tests are allowed to fail and are failing

1oglop1:
The root cause is that there is something different between `botocore==1.12.149` and current `botocore==1.12.180` which breaks tests inside `tox` again

1oglop1:
After trial and error, I found the breaking change to be possibly introduced in `botocore==1.12.156`

1oglop1:
After more tests it seems that breaking change in `botocore==1.12.156` is allowing `urllib3==1.25`
because botocore with `urllib==1.24` with latest boto3 still works

```
E           botocore.exceptions.HTTPClientError: An HTTP Client raised and unhandled exception: 'AWSHTTPSConnection' object has no attribute 'assert_hostname'

.tox/py37-boto3/lib/python3.7/site-packages/botocore/httpsession.py:298: HTTPClientError
```

again this is broken only when tests are run inside tox

1oglop1:
@arthurHamon2 If you are interested to dive in,
it seems that one of these changes broke the integration completely not only tests https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#125-2019-04-22

last error I was able to get is:
```
E           botocore.exceptions.HTTPClientError: An HTTP Client raised and unhandled exception: 'AWSHTTPSConnection' object has no attribute 'assert_hostname'
```


or if `boto3_stubs.py`-> `VerifiedHTTPSConnection.assert_hostname=False`
```
E           botocore.exceptions.SSLError: SSL validation failed for https://iam.amazonaws.com/ [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1056)

```

I do not have more time to investigate anymore.

arthurHamon2:
@1oglop1 It seems I have already encountered this error, I can take a look !
It seems that I can't see this error in the Travis CI build, I only have this : 

```
tests/integration/test_boto3.py:11: in <module>
    aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
.tox/boto3/lib/python3.6/os.py:669: in __getitem__
    raise KeyError(key) from None
E   KeyError: 'AWS_ACCESS_KEY_ID'
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
====================== 7 skipped, 1 error in 1.15 seconds ======================
```

How can I test it ? Do you have an `AWS_ACCESS_KEY_ID` that I can use ? Without this, I'm afraid I can't do much ...

1oglop1:
I could either provide you with credentials or give you permission to use my Travis
PM mi on Gitter please

arthurHamon2:
@1oglop1 I can't see how to fix the issue with the newest boto3 version, can we just freeze the botocore version and try to fix the issue in another PR ? We will be able to merge this PR then :)

arthurHamon2:
I'have finally managed to fix boto3 integration and clean the commit history !
@1oglop1 does it sound good to you ?
@neozenith can you review/merge this PR please ?
FYI : the boto3 tests fails on kevin1024 repository but pass on @1oglop1 repo : https://travis-ci.org/jans-forks/vcrpy/builds/558223421 (that's why there are still marked as allowed to fail, I assume the kevin1024's travis did not setup AWS credentials unlike @1oglop1's)

1oglop1:
Nice job!
We discussed possibility of mocking out tests by using localstack. Would you mind looking into it as well?

arthurHamon2:
@1oglop1 let's merge this PR first and I will try to open a new one using `localstack` library, what do you think ?

1oglop1:
I just verified the work of @arthurHamon2 and everything is working correctly.
@neozenith 
I also conducted small test with `localstack` and it seems possible to use this for the project.
However setting up correctly would take more time than test agains real AWS.
Because `localstack` runs in `http` by default and to set up `https` endpoints we'd need to create trusted certificates inside test environment.

I'd say that we should merge current integration and move `localstack` tests into separate issue.

PR #442, opened 2019-06-14
fixes #440 VCR responses don't work with Biopython on Python3 

### LINKED ISSUE #440: Doesn't work correctly with Biopython on Python3
I am using `vcrpy` version `2.0.1` with `biopython` version `1.72` when Biopython tries to iterate over the VCRHttpResponse `'VCRHTTPResponse' object has no attribute 'fp'` exception is thrown.

  **Note: works fine in Python2 and in Python3 too without VCR**

Stack Trace:
```
  File "/home/addgene/addgene-core/src/django/addgene/common/tests/regression/efetch.py", line 149, in test_efetch_vcr
    articles = list(Medline.parse(handle))
  File "/srv/addgene-py3/lib/python3.5/site-packages/Bio/Medline/__init__.py", line 129, in parse
    for line in handle:
  File "/usr/lib/python3.5/http/client.py", line 470, in readinto
    if self.fp is None:
AttributeError: 'VCRHTTPResponse' object has no attribute 'fp'
```

If I add `self.fp = None` in the constructor of `VCRHTTPResponse` class, `fp` error is not thrown.

 Even after fixing the `fp` attribute issue, Iterating over the response results in an empty list.

Regression Test which I wrote for it.
```
from __future__ import unicode_literals

import io
from Bio import Medline

from addgene.test.django import AddgeneTestCase

from vcr.stubs import VCRHTTPResponse

from Bio._py3k import _binary_to_string_handle, _bytes_to_string


class TestVcrResponseForBiopython(AddgeneTestCase):

    def setUp(self):
        self.recorded_response = {
            "status": {
                "message": "OK",
                "code": 200
            },
            "headers": {
                "content-length": ["0"],
                "server": ["gunicorn/18.0"],
                "connection": ["Close"],
                "access-control-allow-credentials": ["true"],
                "date": ["Fri, 24 Oct 2014 18:35:37 GMT"],
                "access-control-allow-origin": ["*"],
                "content-type": ["text/html; charset=utf-8"],
            },
            "body": {
                "string":b"\nPMID- 19416910\nOWN - NLM\n"
            }
        }
        self.vcr_response = VCRHTTPResponse(self.recorded_response)

    def test_efetch_vcr(self):
        handle = _binary_to_string_handle(self.vcr_response)
        articles = list(Medline.parse(handle))
```
`_binary_to_string_handle` is used by the `Bio.Entrez.efetch` method which makes the actual http request (recorded via VCR) and passes the output of `_binary_to_string_handle(VCRHttpResponse)` to the `Medline.parse` method.  

code of `_binary_to_string_handle`:
```
    def _binary_to_string_handle(handle):
        """
        params: handle: VCRHTTPResponse

        Treat a binary (bytes) handle like a text (unicode) handle (PRIVATE).
        """
        try:
            # If this is a network handle from urllib,
            # the HTTP headers may tell us the encoding.
            encoding = handle.headers.get_content_charset()
        except AttributeError:
            encoding = None
        if encoding is None:
            # The W3C recommendation is:
            # When no explicit charset parameter is provided by the sender,
            # media subtypes of the "text" type are defined to have a default
            # charset value of "ISO-8859-1" when received via HTTP.
            # "ISO-8859-1" is also known as 'latin-1'
            # See the following for more detail:
            # https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1
            encoding = 'latin-1'
        wrapped = io.TextIOWrapper(io.BufferedReader(handle), encoding=encoding)
        try:
            # If wrapping an online handle, this is nice to have:
            wrapped.url = handle.url
        except AttributeError:
            pass
        return wrapped
```

Code of Medline.parse:
```
def parse(handle):
    """
    handle: <_io.TextIOWrapper encoding='utf-8'>
    
    Read Medline records one by one from the handle.

    The handle is either is a Medline file, a file-like object, or a list
    of lines describing one or more Medline records.

    Typical usage::

        from Bio import Medline
        with open("mymedlinefile") as handle:
            records = Medline.parse(handle)
            for record in records:
                print(record['TI'])

    """
    # TODO - Turn that into a working doctest
    # These keys point to string values
    textkeys = ("ID", "PMID", "SO", "RF", "NI", "JC", "TA", "IS", "CY", "TT",
                "CA", "IP", "VI", "DP", "YR", "PG", "LID", "DA", "LR", "OWN",
                "STAT", "DCOM", "PUBM", "DEP", "PL", "JID", "SB", "PMC",
                "EDAT", "MHDA", "PST", "AB", "AD", "EA", "TI", "JT")
    handle = iter(handle)

    key = ""
    record = Record()
    for line in handle:
        line = line.rstrip()
        if line[:6] == "      ":  # continuation line
            if key == "MH":
                # Multi-line MESH term, want to append to last entry in list
                record[key][-1] += line[5:]  # including space using line[5:]
            else:
                record[key].append(line[6:])
        elif line:
            key = line[:4].rstrip()
            if key not in record:
                record[key] = []
            record[key].append(line[6:])
        elif record:
            # Join each list of strings into one string.
            for key in record:
                if key in textkeys:
                    record[key] = " ".join(record[key])
            yield record
            record = Record()
    if record:  # catch last one
        for key in record:
            if key in textkeys:
                record[key] = " ".join(record[key])
        yield record
``` 
In this `for line in handle:` has 0 iterations, and `handle` is `io.TextIOWrapper(io.BufferedReader(handle), encoding=encoding)` and it returns an empty list. For normal HttpResponse handle it works as expected and returns the list of articles. 

My guess is something is wrong in the way VCR defines VCRHTTPResponse which doesn't confine with the expected HTTP response norms of Biopython.

Biopython Docs references:  

https://biopython.org/DIST/docs/api/Bio._py3k-module.html

https://biopython.org/DIST/docs/api/Bio.Medline-module.html

rishab121:
fixes https://github.com/kevin1024/vcrpy/issues/440

I have added `fp` pointer to the VCRHTTPResponse constructor and implemented few IOBase methods. I have also added a regression test for the same.

### DIFF
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -1,4 +1,6 @@
 # coding: UTF-8
+import io
+
 from vcr.stubs import VCRHTTPResponse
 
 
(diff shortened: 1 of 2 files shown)


arthurHamon2:
@rishab121 can you please rebase your branch on master in order to run the Travis CI ?

rishab121:
@arthurHamon2 Ready for re-review.

PR #439, opened 2019-06-12
Add verbosity/explanations on CannotOverwriteExistingCassetteException

arthurHamon2:
First of all, thanks @kevin1024 and to all who have contributed to this project, it is awesome and I'm using it for my professional projects for the last 4 years.

I'm working on a project with particularly big requests bodies and query parameters, when I record a cassette and make changes to my code, the cassette fails sometimes but without any information on what requests changes made the cassette to fail. This can be annoying to spot any differences with a large body (and the YAML encoding making it harder as the body is stored as base64, but I saw the community had a workaround : https://github.com/kevin1024/vcrpy/issues/241). My point is we have little information about why we are getting the "CannotOverwriteExistingCassetteException" exception. 

The PR https://github.com/kevin1024/vcrpy/pull/309 and https://github.com/kevin1024/vcrpy/pull/332 are related to this topic but do not improve much as the logs in debug mode is very verbose and we still don't have the two different values printed, only the matchers that are failing.

My idea here is to get the closest request(s) in the cassette to the one being search, based on the number of matchers that have succeeded and get a detailed assertion message about the differences. This PR includes a big refactoring on how a matcher works, but it is compatible with the old way. Basically I use the assert statement to spot the differences instead of a boolean value.

This is an example of the new `CannotOverwriteExistingCassetteException` message :

```
CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('/private/var/folders/0t/wwygv0ys3ll2ydlq636m7_p00000gn/T/pytest-of-arthurhamon/pytest-157/test_default_matcher_does_not_3/test.yml') in your current record mode ('once').
No match for the request (<Request (GET) http://127.0.0.1:52107/get?p1=q1&a=b>) was found.
Similar request found : (<Request (GET) http://127.0.0.1:52107/get?p1=q1&p2=q2>).
Matcher failed : query
--------------- DETAILS ---------------
[('a', 'b'), ('p1', 'q1')] != [('p1', 'q1'), ('p2', 'q2')]
----------------------------------------
```

It gets even better when you tell pytest to re-rewrite the assert statements of the vcr.matchers module (this should be added to the pytest plugin if the PR is merged : https://github.com/ktosiek/pytest-vcr):

`tests/integarations/__init__.py`
```
import pytest

pytest.register_assert_rewrite("vcr.matchers")
```

```
[...]

PR #445, opened 2019-06-26
Add context to thrown CannotOverwriteExistingCassetteException in write_protected mode

neozenith:
# Context

I have attempted to use a `conftest.py` local plugin to extract the information I need to create a diff of the intended request and the recorded requests when a `CannotOverwriteExistingCassetteException` occurs.

https://stackoverflow.com/questions/56745834/python-pytest-pytest-exception-interact-customize-exception-information-from-vcr

I can get:
 - The list of matchers
 - The list of recorded requests from the cassette

What I can't get:
 - The request from the patched VCRHTTPConnection

What this PR proposes is adding the request and cassette onto the exception to then be handled arbitrarily.

# Usage

The idea is to get the values back out via the `args` property of an `Exception`

```python
@pytest.hookimpl(hookwrapper=True)
def pytest_exception_interact(node, call, report):
    excinfo = call.excinfo
    if report.when == "call" and isinstance(excinfo.value, CannotOverwriteExistingCassetteException):
        vcr_request = None
        cassette = None
        for arg in excinfo.value.args:
            if isinstance(arg, Cassette):
                cassette = arg
            if isinstance(arg, Request):
                vcr_request = arg

        if cassette and vcr_request:
            try:
                raise RequestNotFoundCassetteException(f"....") from excinfo.value
            except RequestNotFoundCassetteException as e:
                excinfo._excinfo = (type(e), e)
                report.longrepr = node.repr_failure(excinfo)


```

# Misc:

I tried running the tox tests but kept getting:

```
ERROR: InvocationError for command 'C:\Users\jpeak\projects\vcrpy\runtests.sh' (exited with code 8)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
     - secure: LBSEg/gMj4u4Hrpo3zs6Y/1mTpd2RtcN49mZIFgTdbJ9IhpiNPqcEt647Lz94F9Eses2x2WbNuKqZKZZReY7QLbEzU1m0nN5jlaKrjcG5NR5clNABfFFyhgc0jBikyS4abAG8jc2efeaTrFuQwdoF4sE8YiVrkiVj2X5Xoi6sBk=
   matrix:
     - TOX_SUFFIX="flakes"
-    - TOX_SUFFIX="requests27"
+    - TOX_SUFFIX="requests"
     - TOX_SUFFIX="httplib2"
     - TOX_SUFFIX="boto3"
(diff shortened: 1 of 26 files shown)


arthurHamon2:
Hi @neozenith !

I have exactly the same need and I proposed a PR here : https://github.com/kevin1024/vcrpy/pull/439
I think we are trying to achieve the same goal here : knowing why your cassette failed to be replayed.
Maybe we can merge our diffs, what do you think ?

Issue #447, opened 2019-06-28
v2.1.x release


jasongi-actu:
There hasn't been a release since September 2018. The collections.abc (#416) fix hasn't been deployed yet which will mean vcr.py stops working on 3.8. 3.8 goes live on 2019-10-21, but is already in beta. I think it's a good idea to pre-empt everybody's tests failing when they upgrade by doing a release.

neozenith:
@arthurHamon2 has put a tonne of work in the last few days as well. So that is due to be released. 

The 3.8 launch sounds like a reasonable time line to aim for getting a release out.

arthurHamon2:
@neozenith @jasongi-actu I have already fix this compatibility issue on my fork, let me 1 or 2 days to create a PR and then I will need your reviews in order to merge it, I'll keep you in touch !

neozenith:
As a matter of organisation would it be worth setting up a gitter channel to coordinate organisational activities separate from issues and PRs?

https://gitter.im/

neozenith:
To capture what we want to release for v2.1.x the list as far as I'm aware is:
 - Python 3.8 support
 - Fix TravisCI and test matrix
 - [WIP] #382 `boto3` compatability (Potentially target v2.2?)
 - Improved matcher nearest match messages
     - TODO: Update readthedocs documentation for this 

@arthurHamon2 ... anything I missed? 

Normally I'd run this to get an idea of changes relative to last tagged release.
```bash
git log --oneline $(git describe --tags $(git rev-list --tags --max-count=1))...master
```

@kevin1024 @IvanMalison as the most experienced contributors on this project could I get some guidance on the release process for:
 - PyPI
 - Readthedocs

I'd like to get this documented in the Contributing guide so I have a reference for subsequent releases.

arthurHamon2:
@neozenith I have also merge multiple pending PRs that fix various bugs. I would like to merge as much as possible for the next release. Could you help me posting messages to the pending PR in order to ask the contributors to rebase their branch with master ?

neozenith:
Sounds good. I'll start doing that as part of daily rounds.

kevin1024:
Make sure you update the changelog before release.  For pypi - it’s just build the sdist, build the wheel, and publish.  For RTD - good question.  I think it’s supposed to auto build?  I can add you to pypi, as a maintainer - just let me know your username over there.

neozenith:
Ok thanks. 

[...]

PR #449, opened 2019-07-01
Fix all warnings

arthurHamon2:
Fix all the warnings that appear when running tests.
That fixes the deprecation on collections.abc as well, in order to work with the new Python 3.8 release.


arthurHamon2:
@neozenith can you review these changes ? This include the discussion in this issue on Python 3.8 compatiblity : https://github.com/kevin1024/vcrpy/issues/447

neozenith:
Good use of trying to use libYAML too.

PR #451, opened 2019-07-02
fix IndexError when empty cassette throws CannotOverwriteCassetteException

neozenith:
# Context

 - No cassette recorded as yet
 - pytest default is `--vcr-record=none`
 - Attempt to request in write_protected mode to prove test suite fails

Fails with IndexError instead of CannotOverwriteCassetteException

### DIFF
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -311,7 +311,7 @@ class Cassette(object):
         best_matches.sort(key=lambda t: t[0], reverse=True)
         # Get the first best matches (multiple if equal matches)
         final_best_matches = []
-        previous_nb_success = best_matches[0][0]
+        previous_nb_success = 0 
         for best_match in best_matches:
             nb_success = best_match[0]
             # Do not keep matches that have 0 successes,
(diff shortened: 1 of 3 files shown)


vcr/cassette.py
```diff
@@ -309,7 +309,7 @@ def find_requests_with_most_matches(self, request):
         best_matches.sort(key=lambda t: t[0], reverse=True)
         # Get the first best matches (multiple if equal matches)
         final_best_matches = []
-        previous_nb_success = best_matches[0][0]
```
arthurHamon2: If the cassette is empty, it means that `best_matches` is empty too. Just add a return statement if we are in this case.

vcr/cassette.py
```diff
@@ -309,7 +309,11 @@ def find_requests_with_most_matches(self, request):
         best_matches.sort(key=lambda t: t[0], reverse=True)
         # Get the first best matches (multiple if equal matches)
         final_best_matches = []
-        previous_nb_success = 0 
+
+        if not best_matches:
```
neozenith: @arthurHamon2 simple guard statement instead. Thanks for pointer.

neozenith:
I'll follow up the unit test tomorrow. Thanks for the reviews.

neozenith:
I thought we fixed the boto3 issue? It's still marked as allowed failure?

arthurHamon2:
@neozenith It's still mark as allowed failure as we do not have valid AWS credentials on this repository. The idea by @1oglop1 is to mock the AWS behavior with a library called `localstack`, but it will be done in another PR.
I'll merge this PR then !

PR #435, opened 2019-05-22
Drop support for EOL Python 3.4

hugovk:
Python 3.4 is EOL and no longer receiving security updates (or any updates) from the core Python team since 2019-03-16.

Source: https://en.wikipedia.org/wiki/CPython#Version_history

Here's the pip installs for vcrpy from PyPI for April 2019:

| category | percent | downloads |
|----------|--------:|----------:|
|      3.6 |  34.62% |    54,167 |
|      3.7 |  28.32% |    44,318 |
|      2.7 |  23.90% |    37,405 |
|      3.5 |  10.45% |    16,359 |
|      3.4 |   1.65% |     2,575 |
| null     |   0.61% |       956 |
|      3.8 |   0.44% |       684 |
|      3.3 |   0.01% |        15 |
|      2.6 |   0.00% |         2 |
| Total    |         |   156,481 |

Source: `pypistats python_minor vcrpy --last-month # pip install pypistats`

It also reduces the CI matrix from 42 to 36 jobs (Total time 58 min 7 sec -> 47 min 11 sec).

This also fixes a Flake8 error.

PR #455, opened 2019-07-08
Fix Python version for extra requirement YARL

browniebroke:
This was excluded when Python 3.4 support was dropped in #435

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ install_requires = [
     'six>=1.5',
     'contextlib2; python_version=="2.7"',
     'mock; python_version=="2.7"',
-    'yarl; python_version>"3.5"',
+    'yarl; python_version>="3.5"',
 ]
 
 excluded_packages = ["tests*"]
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ install_requires = [
     'six>=1.5',
     'contextlib2; python_version=="2.7"',
     'mock; python_version=="2.7"',
+    'yarl; python_version>="3.5"',
 ]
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -28,7 +28,7 @@ def run_tests(self):
     'six>=1.5',
     'contextlib2; python_version=="2.7"',
     'mock; python_version=="2.7"',
-    'yarl; python_version>"3.5"',
+    'yarl; python_version>="3.5"',
```
lamenezes: do we really need yarl as a vcrpy requirement? I ask because it's only used by aiohttp and it's already a requirement there [1]. @browniebroke would you please test if it's ok to remove it instead of changing the minimum version?

[1] https://github.com/aio-libs/aiohttp/blob/master/setup.py#L84

browniebroke: There is a place in the codebase where `yarl` is imported and when this happens, usually the module should be listed as a dependency:

https://github.com/kevin1024/vcrpy/blob/a17624a4649308c2f5f9255391fad8ee4c0a351f/vcr/stubs/aiohttp_stubs/__init__.py#L10

Could this stub be applied if `aiohttp` isn't installed?

browniebroke: I'm trying to remove it now, let's see if the build passes...

lamenezes: @browniebroke I missed the line where yarl is explicitly used in the aiohttp stub. Therefore I agree with your initial statement: we should change the minimum version for yarl.

browniebroke: No worries, I'll revert my last commit then :)

PR #457, opened 2019-07-15
Fix log format bug

stj:
'{}' is not a valid in log formatting.

This fixes an issue I committed a17624a.

### DIFF
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -1,4 +1,5 @@
 import contextlib
+import logging
 
 import pytest
 asyncio = pytest.importorskip("asyncio")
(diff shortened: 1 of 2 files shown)


arthurHamon2:
Nice catch ! Could you commit only the file `vcr/stubs/aiohttp_stubs/__init__.py` ? The other changes seems to be related to style.

stj:
Sorry about that. My editor is set up to run black on save. I should have checked the changes before committing. It's fixed now.

arthurHamon2:
@stj looks like you have break the tests, can you take a look to fix them ?

stj:
@arthurHamon2 fixed that.

PR #467, opened 2019-08-07
Format project with black

neozenith:
Format with line length 110 to match flake8

Toying with the idea of migrating this project to use black as a code formatter.

Tasks:
 - [x] Add Tox environment that runs black check
 - [x] Add Travis job that handles checking 

Key files changed for review:
 - tox.ini
 - .travis.yml
 - pyproject.toml
 - readme.rst
 - docs/changelog.rst

All the .py files too


tox.ini
```diff
@@ -25,6 +25,13 @@ commands =
     pyflakes ./docs/conf.py
 deps = flake8
 
+[testenv:format]
+skipsdist = True
+commands =
+    black --version
+    black --check .
```
hugovk: `--diff` will also print out the difference, which can save some time if the PR author doesn't have Black installed:

```suggestion
    black --check --diff .
```

tox.ini
```diff
@@ -25,6 +25,13 @@ commands =
     pyflakes ./docs/conf.py
 deps = flake8
 
+[testenv:format]
```
hugovk: I suggest including this in the `flakes` job (and possibly renaming it `lint`) as the test matrix is already a pretty hefty 36 jobs.

And fixing Black often fixes Flake8 too.

neozenith: Great idea!

PR #479, opened 2019-08-28
Clarify documentation of custom request matchers

gward:
* The choice is not what the function *returns*, but what it *does*
  (raising an exception is not returning a value).

* Recommend "assert" more strongly, and additionally recommend an
  error message with the assert.

* Back up that recommendation in the sample code.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -97,12 +97,12 @@ Create your own method with the following signature
 
     def my_matcher(r1, r2):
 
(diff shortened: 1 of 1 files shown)


neozenith:
https://vcrpy.readthedocs.io/en/latest/advanced.html

neozenith:
Also... https://vcrpy.readthedocs.io/en/latest/changelog.html?highlight=gward#changelog
;)

PR #483, opened 2019-09-29
Add `pytest-recording` to the documentation as an alternative Pytest plugin

### LINKED ISSUE #476: Alternative pytest plugin
Hello folks!
Thank you for this library, it helps us a lot :)
I want to propose to add a mention of an alternative pytest plugin to VCR.py docs - [pytest-recording](https://github.com/kiwicom/pytest-recording)
It happened to us that some `pytest-vcr` limitations were blocking us and I decided to rewrite it in a slightly different way. 

### Differences

**Automatic cassettes naming**

`pytest-recording` names cassettes in the way to avoid possible collisions:
- `{cassette_lib}/{test_module}/{test_name}.{ext}`
- `{cassette_lib}/{test_module}/{test_class}.{test_name}.{ext}`

It helps when there are multiple test modules that have test functions with the same name. In this case, `pytest-recording` will create two cassettes:

```
tests
├── cassettes
│  ├── test_a
│  │  └── test_function.yaml
│  └── test_b
│     └── test_function.yaml
├── test_a.py
└── test_b.py
```

on the other hand, `pytest-vcr` will either fail on the second test (`--vcr-record-mode` not set) or will put requests/responses in the same file (`--vcr-record-mode=all`), which seems a bit counter-intuitive to me, because I expect isolated cassettes per test function.

```
tests
├── cassettes
│  └── test_function.yaml
├── test_a.py
└── test_b.py
```

Also, when test item has chars like `/` in its name (e.g. via `pytest.mark.parametrize`) then the name is sanitized and no subdirectories are created (there is a pending PR for `pytest-vcr` that fixes it https://github.com/ktosiek/pytest-vcr/pull/24)

**Custom cassette name via pytest-mark**

`pytest.mark.vcr` from `pytest-recording` passes all args & kwargs directly to `use_cassette` and it is possible to specify a custom cassette name like this:

```
@pytest.mark.vcr("custom.yaml")
def test_something()
```

In `pytest-vcr` the cassette name is taken only from the `vcr_cassette_name` fixture

Ref: https://github.com/ktosiek/pytest-vcr/issues/23

**Composable `vcr` marks**

It is possible to specify marks on multiple levels - the resulting config will be a merge of all marks applied to the test function.

**Default recording mode**

`pytest-recording` uses `recording-mode=none` by default to prevent accidental network access in the marked tests

**Multiple cassettes per test**

It is possible to specify multiple cassettes per test function which could decrease cassettes duplication (in our case we were working with 3rd party currency rates API and those requests were almost everywhere), and only the closest one is writable.

**Extras**:
- network blocking for `socket.connect*` and `pycurl`, but most probably it will be moved into 
a separate plugin
- WIP cassettes mutation as a fault-injection for tests

Hope, that this plugin could be mentioned along with `pytest-vcr` in the docs :) Also, I'd be happy to hear your feedback on the library and provide support to address any comments / adjustments / feature requests :)

Cheers

Stranger6667:
Closes #476

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,7 @@ Changelog
   - Use latest pypy3 in Travis (@hugovk)
   - Improve documentation about custom matchers (@gward)
   - Fix exception when body is empty (@keithprickett)
+  - Add `pytest-recording` to the documentation as an alternative Pytest plugin (@Stranger6667)
 -  2.1.0 - Add a `rewind` method to reset a cassette (thanks @khamidou)
    New error message with more details on why the cassette failed to play a request (thanks @arthurHamon2, @neozenith)
    Handle connect tunnel URI (thanks @jeking3)
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -101,3 +101,6 @@ Pytest Integration
 
 A Pytest plugin is available here : `pytest-vcr
 <https://github.com/ktosiek/pytest-vcr>`__.
+
+Alternative plugin, that also provides network access blocking: `pytest-recording
+<https://github.com/kiwicom/pytest-recording>`__.


arthurHamon2:
@Stranger6667 can you solve the conflict ? I'll merge your PR after that :)

Stranger6667:
Done! Is it ok to have a merge commit? If not, then I can do rebase

arthurHamon2:
Please rebase as the merge commit will only make the git history unclear.

Stranger6667:
Done!

arthurHamon2:
Thx :)

Issue #485, opened 2019-10-17
Yarl dependency incompatible with Python 3.5


jdevera:
From setup.py:

```python
    'yarl; python_version>="3.5"',
```

But in yarl's setup:

```python
if sys.version_info < (3, 6):
    raise RuntimeError("yarl 1.4+ requires Python 3.6+")
```

PR #488, opened 2019-11-02
Fix header matcher for boto3 (fixes #474)

### LINKED ISSUE #474: TypeError: a bytes-like object is required, not 'str' when processing header object
```
headers = {'X-Amz-Target': b'<target>', 'Content-Type': b'<content-type>', 'User-Agent': b'<user-agent>', Signature=b'<sig>', 'Content-Length': '<len>'}

    def checker(headers):
>       return value in headers.get(header, '').lower()
E       TypeError: a bytes-like object is required, not 'str'

.venv/lib/python3.6/site-packages/vcr/matchers.py:56: TypeError
```

The code on line 56 in matchers fails unexpectedly. The content type return as a bytes like object   but the code still fails. When we convert the bytes object to a string it works. Replacing the current function at https://github.com/kevin1024/vcrpy/blob/master/vcr/matchers.py#L54-L57:
```
def _header_checker(value, header='Content-Type'):
    def checker(headers):
        return value in headers.get(header, '').lower()
    return checker
```
with  
```
def _header_checker(value, header='Content-Type'):
    def checker(headers):
        return value in headers.get(header, '').decode('utf-8').lower()
    return checker
```
fixes this issue especially with boto3 requests. 

Would there be any risk with making this change?

simahawk:
tested, works smoothly with this set of headers that were failing before:
```
headers = {'X-Amz-Content-SHA256': b'UNSIGNED-PAYLOAD', 'Cache-Control': b'max-age=31536000, public', 'X-Amz-Date': b'20191102T143910Z', 'User-Agent': b'Boto3/1.9.102 Python/3.5.3 Linux/4.15.0-54-generic Botocore/1.12.253 Resource', 'Content-MD5': b'GQqjEXsRqrPyxfTl99nkAg==', 'Content-Type': b'text/plain', 'Expect': b'100-continue', 'Content-Length': '21'}
```
I'll see if I can provide a fixture for tests.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,7 @@ Changelog
   - Fix exception when body is empty (@keithprickett)
   - Add `pytest-recording` to the documentation as an alternative Pytest plugin (@Stranger6667)
   - Fix yarl and python3.5 version issue (@neozenith)
+  - Fix header matcher for boto3 - fixes #474 (@simahawk)
 -  2.1.0 - Add a `rewind` method to reset a cassette (thanks @khamidou)
    New error message with more details on why the cassette failed to play a request (thanks @arthurHamon2, @neozenith)
    Handle connect tunnel URI (thanks @jeking3)
--- a/tests/unit/test_matchers.py
+++ b/tests/unit/test_matchers.py
@@ -54,6 +54,16 @@ req2_body = (
     b"<member><name>a</name><value><string>1</string></value></member>"
     b"</struct></value></data></array></value></param></params></methodCall>"
 )
+boto3_bytes_headers = {
+    "X-Amz-Content-SHA256": b"UNSIGNED-PAYLOAD",
+    "Cache-Control": b"max-age=31536000, public",
+    "X-Amz-Date": b"20191102T143910Z",
+    "User-Agent": b"Boto3/1.9.102 Python/3.5.3 Linux/4.15.0-54-generic Botocore/1.12.253 Resource",
+    "Content-MD5": b"GQqjEXsRqrPyxfTl99nkAg==",
+    "Content-Type": b"text/plain",
+    "Expect": b"100-continue",
+    "Content-Length": "21",
+}
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


simahawk:
travis red on some combinations as `master`. Apparently not related to this PR.

neozenith:
Changes look good. Thanks for the PR!

You’re right, there is something on master CI we haven’t got around to addressing that’s throwing some red. 

When I get some spare time I’ll have a look to clear that so we can push this through. 

If you could add the associated test that’d be appreciated. 

And.... add in `docs/Changelog.rst` a line to claim credit.

neozenith:
Turns out the CI red is due to a Yarl version issue raised in #485

neozenith:
Fixed in #489, please rebase with master and let's have another crack at getting this pushed through. Due to release to PyPI the accumulated patches too.

simahawk:
@neozenith rebased, tests updated

simahawk:
forgot to run black. Any reason not to use https://pre-commit.com/ here?

neozenith:
Thanks for the quick turn around!

That's weird, the travis integration stopped showing on this PR.

I have little experience with using pre-commit. If you'd like to set it up to catch things like black earlier then PR's are welcome.

I've personally never had a reliable experience getting a team to setup git hooks. So I rely on CI to remind them of code standards. Although it looks like pre-commit might do the step I need them to do when they install the dev libraries for testing anyhow.

I don't have the bandwidth so I'll have to pass on that for now but great suggestion.

Thanks again for contributing the patch!

simahawk:
contrib: very welcome. OTOH: thanks to you and anyone else involved in keeping alive such a good project :wink: 
pre-commit: it's very handy. Allows you to avoid useless builds and reviews and fixups just because of linting or other non good practices.
The nice thing is that you can set up as many checks as you want. In the odoo community we are using this kind of conf https://github.com/OCA/storage/blob/12.0/.pre-commit-config.yaml
I can draft a PR w/ the same config then you can play with it according to your needs.

Issue #475, opened 2019-08-16
aiohttp stub throws away all recorded responses to the same endpoint except for the last one


HugoArts:
Hi,

we are using vcrpy to test our api client, built on aiohttp. When upgrading to the 2.1.0 release we noticed our tests started failing. In the failing test, we are making multiple calls to the same URL in succession. It seems that when playing back the casette, vcrpy throws away all but the last response.

I have created a minimal script to demonstrate the behavior. It seems to be specific to the aiohttp stub:

```python3
#! /usr/bin/env python3

import asyncio
import aiohttp
import vcr

@vcr.use_cassette("test_casette.yaml")
async def main():
    async with aiohttp.ClientSession() as session:
        async with session.get('https://postman-echo.com/get') as resp:
            print('FIRST RESPONSE:')
            print(resp.headers)
        async with session.get('https://postman-echo.com/get') as resp:
            print('SECOND RESPONSE:')
            print(resp.headers)

loop = asyncio.get_event_loop()
loop.run_until_complete(main())
```

When running the script twice in succession, we can observe the difference in output between the first run (with real requests) and the second (requests recorded by vcr):

```shell
(venv)[hugo@hugos-mbp vcr_demo]$ python demo.py
FIRST RESPONSE:
<CIMultiDictProxy('Content-Encoding': 'gzip', 'Content-Type': 'application/json; charset=utf-8', 'Date': 'Fri, 16 Aug 2019 10:47:11 GMT', 'Etag': 'W/"e4-Us7L+i6yExRm9KG9mNvMvSP9DQ8"', 'Server': 'nginx', 'Set-Cookie': 'sails.sid=s%3AI6gIdE3cbaI1pza-0CWg621ScWcCNBAV.b8XF%2FDL42icPhHSBA8Cdl50FV%2BthHFk5ztcMdxySJ5M; Path=/; HttpOnly', 'Vary': 'Accept-Encoding', 'Content-Length': '171', 'Connection': 'keep-alive')>
SECOND RESPONSE:
<CIMultiDictProxy('Content-Encoding': 'gzip', 'Content-Type': 'application/json; charset=utf-8', 'Date': 'Fri, 16 Aug 2019 10:47:11 GMT', 'Etag': 'W/"14e-BBSrIh8nOQJ4wopc8Pj851uldcU"', 'Server': 'nginx', 'Vary': 'Accept-Encoding', 'Content-Length': '271', 'Connection': 'keep-alive')>
[...]

Issue #463, opened 2019-07-31
Differences between requests and aiohttp cassettes


dmfigol:
Thank you for the library.
My open-source library supports sync and async at the same time, so I'd love to use the same cassette for both tests. However, there are some challenges.
Here are the cassettes fragments for requests and aiohttp:
`requests`:
```yaml
interactions:
- request:
    ...
  response:
    body:
      string: ...
      headers:
        Access-Control-Expose-Headers:
        - Link
        Cache-Control:
        - no-cache
        Content-Type:
        - application/json;charset=UTF-8
        Date:
        - Wed, 31 Jul 2019 17:32:09 GMT
        Link:
        - <url>
        Server:
        - Redacted
        Strict-Transport-Security:
        - max-age=63072000; includeSubDomains; preload
        Transfer-Encoding:
        - chunked
        Vary:
        - Accept-Encoding
        Via:
        - 1.1 linkerd
        content-length:
        - '38937'
    status:
      code: 200
      message: OK
    url: https://api.ciscospark.com/v1/rooms
version: 1
```
`aiohttp`:
```yaml
interactions:
- request:
    ...
  response:
    body:
      string: ...
    headers:
      Access-Control-Expose-Headers: Link
      Cache-Control: no-cache
      Content-Encoding: gzip
      Content-Type: application/json;charset=UTF-8
      Date: Wed, 31 Jul 2019 17:32:12 GMT
      Link: <url>
      Server: Redacted
      Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
      Transfer-Encoding: chunked
      Vary: Accept-Encoding
      Via: 1.1 linkerd
    status:
      code: 200
      message: OK
    url: <url>
```
We can see two main differences:  
1) in `requests` cassette response headers are serialized as `key -> list of strings`, in `aiohttp` cassette they are serialized as `key -> str`
2) in `aiohttp` cassette there is a response url key which equals to the request url, `requests` cassette does not have that

Also, because of 2), `requests` cassette played in `aiohttp` code will crash:
```python
cls = <class 'yarl.URL'>, val = None

    def __new__(cls, val="", *, encoded=False, strict=None):
        if strict is not None:  # pragma: no cover
            warnings.warn("strict parameter is ignored")
        if type(val) is cls:
            return val
        if type(val) is str:
            val = urlsplit(val)
        elif type(val) is SplitResult:
            if not encoded:
                raise ValueError("Cannot apply decoding to SplitResult")
[...]

PR #495, opened 2019-12-01
aiohttp: fix multiple requests being replayed per request and add support for `request_info` on mocked responses

### LINKED ISSUE #481: aiohttp: replaying doesn't work with two exact requests in replay mode "once"
Hi again maintainers of vcrpy :)

I found this interesting problem today where in my test, I make 2 HTTP identical requests (one in the test and one in app code — perhaps this is a terrible idea and let me know if it is) and both are recorded, but when trying to replay the test it fails on the second request being made.

After debugging this for a while, it looks like what happens is we update `play_counts` for both instances of the responses being played for the first request, even though we should only track one being sent back. The code in question is here:

https://github.com/kevin1024/vcrpy/blob/baadf913ef3ed67fa599c3af18d5f9fcb005cf41/vcr/cassette.py#L267-L270

So `play_states` will be `Counter({0: 1, 1: 1})`, which feels incorrect to me since requests and responses should be one-to-one. Does that sound reasonable?

When looking up `request in self`:
 https://github.com/kevin1024/vcrpy/blob/baadf913ef3ed67fa599c3af18d5f9fcb005cf41/vcr/cassette.py#L260

It fails because of

https://github.com/kevin1024/vcrpy/blob/baadf913ef3ed67fa599c3af18d5f9fcb005cf41/vcr/cassette.py#L355-L360

I think what's happening is something specific to `aiohttp` specifically around these lines:

https://github.com/kevin1024/vcrpy/blob/baadf913ef3ed67fa599c3af18d5f9fcb005cf41/vcr/stubs/aiohttp_stubs/__init__.py#L77-L80

Perhaps it shouldn't be a `while`? It seems that it's necessary for this bug that was recently resolved: https://github.com/kevin1024/vcrpy/pull/456. 

It's been a while since I've looked at this codebase, so I'm not sure what would be best to remedy this and I haven't spent time thinking about a solution since I only just now figured out the problem. @lamenezes @arthurHamon2 it seems like y'all are familiar with this — what do you think we should do here? Or am I doing something completely wrong here and this is working as intended? How can I help, if this is a legitimate bug?

Here's my cassette data:
```yaml
interactions:
- request:
    body: '{"query": "\n    query {\n        coachAssignments {\n            uuid\n        }\n    }\n    "}'
    headers:
      Content-Type:
      - application/json
      authorization:
      - DUMMY
      x-datadog-parent-id:
      - DUMMY
      x-datadog-trace-id:
      - DUMMY
    method: POST
    uri: http://localhost:8000/graphql
  response:
    body:
      string: '{"data":{"coachAssignments":[{"uuid":"59829c6c-c904-49ea-8157-1d1ac6a2412a"}]}}'
    headers:
      Connection: keep-alive
      Content-Length: '79'
      Content-Type: application/json
      Date: Wed, 28 Aug 2019 23:00:18 GMT
      Server: nginx
      Set-Cookie: csrftoken=redacted;
        expires=Wed, 26 Aug 2020 23:00:18 GMT; Max-Age=31449600; Path=/; SameSite=Lax
      Vary: Cookie, Origin
      X-Frame-Options: SAMEORIGIN
    status:
      code: 200
      message: OK
    url: http://localhost:8000/graphql
- request:
    body: '{"query": "\n    query {\n        coachAssignments {\n            uuid\n        }\n    }\n    "}'
    headers:
      Content-Type:
      - application/json
      authorization:
      - DUMMY
      x-datadog-parent-id:
      - DUMMY
      x-datadog-trace-id:
      - DUMMY
    method: POST
    uri: http://localhost:8000/graphql
  response:
    body:
      string: '{"data":{"coachAssignments":[{"uuid":"59829c6c-c904-49ea-8157-1d1ac6a2412a"}]}}'
    headers:
      Connection: keep-alive
      Content-Length: '79'
      Content-Type: application/json
      Date: Wed, 28 Aug 2019 23:00:18 GMT
      Server: nginx
      Set-Cookie: csrftoken=redacted;
        expires=Wed, 26 Aug 2020 23:00:18 GMT; Max-Age=31449600; Path=/; SameSite=Lax
      Vary: Cookie, Origin
      X-Frame-Options: SAMEORIGIN
    status:
      code: 200
      message: OK
    url: http://localhost:8000/graphql
version: 1
```

nickdirienzo:
This PR resolves #481 and #475, while also improving #456.

It was discovered that, for `aiohttp` specifically, when there were multiple requests recorded that looked the same, they would all be played. Looking into this further, it was because the URL we recording to was the original request's URL instead of the implicit request's that followed the redirects.

This led to the change you see here where we're recording the individual request-response chains (similar to that of other HTTP clients). We also change the explicit `while cassette.can_play_response_for(vcr_request)` to one that checks the response status codes and only performs a similar check if the code is a 3xx one.

While implementing that, I wanted to be extra sure that we had similar looking responses, so I started checking the `request_info` object. Originally, each of the mock responses do not have those objects available (I'm not sure if this is intended or not), so this PR also adds support for that. `test_redirect` was amended to include a specific check that we're returning the same `request_info` between the real one and the recorded one.

Speaking of tests, thanks to @vEpiphyte for writing `test_double_requests`, which made this issue repeatable and give me confidence that this works.

There are two concerns I have around this PR:
1. This change breaks cassettes for `aiohttp` clients where redirects are performed because how we're using the cassette is fundamentally different. It feels weird to have a major bump for this, but if we're following SemVer, that may be right --- how do you all do versioning? Happy to include the version bump in this PR if that's preferred.
2. I'm pretty skeptical of my usage of `cassette.find_requests_with_most_matches` (https://github.com/kevin1024/vcrpy/compare/master...nickdirienzo:issue-481?expand=1#diff-4636a4db24aa9f9206350b6a8e171ae1R104). While the request _should_ exist, it may not and then the assumed look up will fail hard. We do need to get the recorded request in this block, otherwise we can't have the correct `request_info`. Do y'all have ideas on how to do this more safely?

Some references that I needed along the way:
- `aiohttp` follows redirects by default: https://docs.aiohttp.org/en/stable/client_advanced.html#redirection-history
- `RequestInfo`: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.RequestInfo

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,9 @@
 Changelog
 ---------
+-  3.0.0 (UNRELEASED)
+  - Fix multiple requests being replayed per single request in aiohttp stub (@nickdirienzo)
+  - Add support for `request_info` on mocked responses in aiohttp stub (@nickdirienzo)
+  - ...
 -  2.1.x (UNRELEASED)
   - ....
 -  2.1.1 
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -262,3 +262,41 @@ def test_redirect(aiohttp_client, tmpdir):
         assert len(cassette_response.history) == len(response.history)
         assert len(cassette) == 3
         assert cassette.play_count == 3
+
+
+def test_double_requests(tmpdir):
+    """We should capture, record, and replay all requests and response chains,
+        even if there are duplicate ones.
+
+        We should replay in the order we saw them.
+    """
+    url = "https://httpbin.org/get"
+
+    with vcr.use_cassette(str(tmpdir.join("text.yaml"))):
+        _, response_text1 = get(url, output="text")
+        _, response_text2 = get(url, output="text")
+
+    with vcr.use_cassette(str(tmpdir.join("text.yaml"))) as cassette:
+        resp, cassette_response_text = get(url, output="text")
+        assert resp.status == 200
+        assert cassette_response_text == response_text1
+        # This check fails because we increment the play_count
+        # to 2 in the while loop!
+        assert cassette.play_count == 1
+
+        # Now make the second test to url
+        resp, cassette_response_text = get(url, output="text")
+
+        # This check fails because we have a 599, not a 200 returned.
+        assert resp.status == 200
+        # assert resp.status == 599  # <----- sad path check :(
+
+        # This check fails because we get a error message back, not the
+        # actual response text
+        assert cassette_response_text == response_text2
+
+        # sad_path_message = 'No match for the request <Request (GET) https://httpbin.org/get> was found'
+        # assert cassette_response_text.startswith(sad_path_message)  # <--- Sad path check :(
+
+        # The final play_count is correct
+        assert cassette.play_count == 2
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -263,6 +263,15 @@ def test_redirect(aiohttp_client, tmpdir):
         assert len(cassette) == 3
         assert cassette.play_count == 3
 
+    # Assert that the real response and the cassette response have a similar
+    # looking request_info.
+    assert cassette_response.request_info.url == response.request_info.url
+    assert cassette_response.request_info.method == response.request_info.method
+    assert {k: v for k, v in cassette_response.request_info.headers.items()} == {
+        k: v for k, v in response.request_info.headers.items()
+    }
+    assert cassette_response.request_info.real_url == response.request_info.real_url
+
 
 def test_double_requests(tmpdir):
     """We should capture, record, and replay all requests and response chains,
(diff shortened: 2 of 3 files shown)


nickdirienzo:
What do you think @neozenith? How can I help make this easy for you and the other maintainers to get this over the finish line? Anyone else I should tag? :)

neozenith:
Hey @nickdirienzo thank you so much for diving into this issue! 

It'd make me very happy to get this over the line. If it is a breaking change though, we should bump it to v3.x.x

That said we are due to cull python2 support with PR https://github.com/kevin1024/vcrpy/pull/443 as well.

I might add this to the `v3.0.x` milestone https://github.com/kevin1024/vcrpy/milestone/4

I'll have to circle back to reviewing this later this week as I have limited bandwidth.

nickdirienzo:
)You're welcome!

Bumping to 3.x.x sounds right. I'm mostly worried about users that may have been following redirects in their tests so they would have to re-record those tests. However, I'm not entirely sure those tests are correct with the current implementation, so maybe I'm over-worrying about this.

What's the timeline for the v3 release? (A bit of an aside: I can try to also find time to hack on #463. However, I'm not sure if that's a must-have for the release or not.)

And no worries. Take your time on reviewing this one. Appreciate any and all feedback!

neozenith:
I'll aim to get on top of the review in the next week. We've been holding off the cull of python2 but that cock is ticking ...

As soon as a PR is ready and good to go I'd rather release it than sit on our hands to collect a group of PRs together since there are so few PRs it makes no sense to wait.

Also means we will increment a few major changes pretty quickly but it all goes into the changelog.

Speaking of which, you should add a line to `docs/changelog.rst` if you haven't already.

nickdirienzo:
Thanks for the context! Yep, 2020 is around the corner 😅 

That sounds good to me. I'll update the changelog.

neozenith:
Hey @nickdirienzo, sorry... I will get to this early this week. Got busy.

nickdirienzo:
Hey Josh! All good. I totally understand. Not in a rush here. I wasn't able
to push up that change until this weekend anyway.

nickdirienzo:
Oh sweet — thanks @neozenith! Let me know when the 3.x release is and I can retest this before we release, if that'd be helpful.

neozenith:
I'm preparing a release now.

PR #498, opened 2019-12-13
v3.0.0 RC

neozenith:
Updated release notes and `setup.py` version number

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,8 +12,9 @@ All help in providing PRs to close out bug issues is appreciated. Even if that i
   - Add Python 3.8 TravisCI support
 -  3.0.0
   - This release is a breaking change as it changes how aiohttp follows redirects and your cassettes may need to be re-recorded with this update.
-  - Fix multiple requests being replayed per single request in aiohttp stub (@nickdirienzo)
-  - Add support for `request_info` on mocked responses in aiohttp stub (@nickdirienzo)
+  - Fix multiple requests being replayed per single request in aiohttp stub #495 (@nickdirienzo)
+  - Add support for `request_info` on mocked responses in aiohttp stub #495 (@nickdirienzo)
+  - doc: fixed variable name (a -> cass) in an example for rewind #492 (@yarikoptic)
 -  2.1.1 
   - Format code with black (@neozenith)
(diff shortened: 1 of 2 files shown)


neozenith:
@nickdirienzo I have to head out for a few hours but this will be the version published to PyPI when I get back to my desk later tonight.

I wanted to take the opportunity in the changelog to let others know a v4.0 which drops Python 2 and adds Python3.8 support is coming.

Feel free to use this in the mean time for pre-release:

```bash
pip install https://github.com/kevin1024/vcrpy/archive/master.zip
# OR
pip install https://github.com/neozenith/vcrpy/archive/master.zip
```

nickdirienzo:
Thanks, Josh! I'm hoping I can test this out later tonight and report
back to you.

nickdirienzo:
I too have to step out for a couple hours, so don't let my checking block the release, unless you feel like it should (in which case, I can definitely look this weekend). I should've tested the library with my commit included with the tests I had in question, but work got really busy really quickly, heh. 😅

nickdirienzo:
Hm, for some reason, the test cases I'm running against I seem to not get matches for one of the cases I was expecting to cover with this. I have some other more pressing things to get to right now, but I can probably take a look this weekend to see if the change caused this as a regression or if I'm just not filtering correctly. 

What would you feel comfortable doing here for the release? It feels like we've made a net positive change here and tests do pass, but I'm not sure what's best for library users.

I'm sorry I didn't test my test suite against my change before submitting the MR (I was doing it all on my personal laptop to keep things a bit separated...).

I can share a redacted YAML file if you'd like, or I can open up a new issue. Whatever you think is best here!

neozenith:
I’m happy to hold tight until you figure out if it is VCRpy library or your library. 

Let’s triage first. A few more days won’t hurt. 

Otherwise Monday I’ll pull the trigger as over all it is a benefit and we are protecting users that lock based on version that it is a major change. And communicated in the change log so they can roll back. So release risk is minimal.

nickdirienzo:
Hmm, okay. I figured out what was happening and good news is the fix works for my use case!

I was running tests in a Docker container which was using the pre-3.0 release. 🤦‍♂

Issue #296, opened 2017-02-15
Details url for older version of VCR is outdated


ustroetz:
The link (http://git.io/mHhLBg) reported when a cassette was generated in an older version of VCR is no longer working. 

If you provide me the new one I send a PR to fix it.

https://github.com/kevin1024/vcrpy/blob/ecbc192fc445d3bd91aba43d1a34dd6d9ccb1976/vcr/serialize.py#L30

neozenith:
A lot of changes have happened to VCRpy since this ticket was opened. As this ticket has become stale and we have a lot of old tickets that need to be groomed, I'm closing this for now to make it easier to see what is still relevant.

However if it is still needed, please feel free to re-open or create a new ticket.

Thanks! 🙏

PR #506, opened 2019-12-24
WIP: Unhide a ValueError for deserialization errors.

### LINKED ISSUE #501: deserialize call chain hdes useful exception information
There are two issues with cassette loading and exception information being lost.

1. The ``deserialize`` method here https://github.com/kevin1024/vcrpy/blob/v3.0.0/vcr/serialize.py#L34 can hide yaml errors ( see #497)
2. The ``_load`` method here https://github.com/kevin1024/vcrpy/blob/v3.0.0/vcr/cassette.py#L338 passes on ``ValueError``.   This is problematic because the default deserialize errors raise ValueError messages.  This hides the potential warning about version change.

In both cases, there is no indication to the user that the cassette failed to load, and vcrpy can then possibly re-record a cassette, overwriting what may be on disk.

I would suggest that we add log.exception or log.error statements there so that users can know that an error has occurred.

vEpiphyte:
Closes #501 

Removes reference to a dead link - OBE's #296

### DIFF
--- a/tests/fixtures/migration/old_cassette.yaml
+++ b/tests/fixtures/migration/old_cassette.yaml
@@ -1,3 +1,6 @@
+# This cassette is problematic because it was recorded
+# with a Python 2.x environment and would NEVER deserialize
+# properly in a 3.x environment.
 - request: !!python/object:vcr.request.Request
     body: null
     headers: !!python/object/apply:__builtin__.frozenset
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ pip-wheel-metadata/
 
 fixtures/
 /docs/_build
+
+.idea/
--- a/runtests.sh
+++ b/runtests.sh
@@ -4,4 +4,4 @@
 # If you are getting an INVOCATION ERROR for this script then there is 
 # a good chance you are running on Windows.
 # You can and should use WSL for running tox on Windows when it calls bash scripts.
-REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` py.test $*
+REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` python -m pytest $*
--- a/tests/unit/test_persist.py
+++ b/tests/unit/test_persist.py
@@ -1,5 +1,6 @@
 import pytest
 
+from vcr.errors import LegacyCassetteError
 from vcr.persisters.filesystem import FilesystemPersister
 from vcr.serializers import jsonserializer, yamlserializer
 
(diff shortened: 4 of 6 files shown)


vEpiphyte:
@neozenith Opened this for discussion

tests/fixtures/migration/old_cassette.yaml
```diff
@@ -1,3 +1,6 @@
+# This cassette is problematic because it was recorded
```
vEpiphyte: This cassette fails to deserialize at all in 3.7...it may be a good test for "bad yaml" if someone was moving from VCR<2.1 on python 2.7 to python 3 and vcr>=2.1 it doesn't produce a yaml loads error, just a format error.

If we want to keep this as valid yaml that will load we need to make the following patch:

```
diff --git a/tests/fixtures/migration/old_cassette.yaml b/tests/fixtures/migration/old_cassette.yaml
index 68d6b49..bd9c286 100644
--- a/tests/fixtures/migration/old_cassette.yaml
+++ b/tests/fixtures/migration/old_cassette.yaml
@@ -1,9 +1,6 @@
-# This cassette is problematic because it was recorded
-# with a Python 2.x environment and would NEVER deserialize
-# properly in a 3.x environment.
 - request: !!python/object:vcr.request.Request
     body: null
-    headers: !!python/object/apply:__builtin__.frozenset
+    headers: !!python/object/apply:builtins.frozenset
     - - !!python/tuple [accept-encoding, 'gzip, deflate, compress']
       - !!python/tuple [user-agent, python-requests/2.2.1 CPython/2.6.1 Darwin/10.8.0]
       - !!python/tuple [accept, '*/*']
@@ -13,9 +10,9 @@
     port: 80
     protocol: http
   response:
-    body: {string: !!python/unicode "{\n  \"origin\": \"217.122.164.194\"\n}"}
-    headers: [!!python/unicode "access-control-allow-origin: *\r\n", !!python/unicode "content-type:
-        application/json\r\n", !!python/unicode "date: Mon, 21 Apr 2014 23:06:09 GMT\r\n",
-      !!python/unicode "server: gunicorn/0.17.4\r\n", !!python/unicode "content-length:
-        32\r\n", !!python/unicode "connection: keep-alive\r\n"]
+    body: {string: !!python/str "{\n  \"origin\": \"217.122.164.194\"\n}"}
+    headers: [!!python/str "access-control-allow-origin: *\r\n", !!python/str "content-type:
+        application/json\r\n", !!python/str "date: Mon, 21 Apr 2014 23:06:09 GMT\r\n",
+      !!python/str "server: gunicorn/0.17.4\r\n", !!python/str "content-length:
+        32\r\n", !!python/str "connection: keep-alive\r\n"]
     status: {code: 200, message: OK}
```

neozenith:
Thanks for following this up! I’ll have a look at it between now and New Years.

neozenith:
Actually I won’t have bandwidth to look at PRs until mid January when I’m finished my annual leave.

vEpiphyte:
@neozenith Have you been able to look at this?

kevin1024:
Closing this WIP PR due to extended inactivity. If this issue still exists, please open a new issue or PR. Thanks for the contribution!

Issue #502, opened 2019-12-17
aiohttp streaming fails on initial cassette recording


mikemoritz:
When using the aiohttp Streaming API and recording a new cassette the actual data read will fail, although the cassette is successfully recorded and subsequent tests pass.

Minimal example:
```
    # Will fail on initial record, but succeed after that
    with vcr.use_cassette('./vcrpytest1.yaml'):
        async with aiohttp.ClientSession() as session:
            async with session.get('https://httpbin.org/range/50') as response:
                byts = b''
                async for data in response.content.iter_chunked(1024*1024):
                    byts += data
                assert len(byts) == 50

    # Fails everytime
    with vcr.use_cassette('./vcrpytest1.yaml', record_mode='all'):
        async with aiohttp.ClientSession() as session:
            async with session.get('https://httpbin.org/range/50') as response:
                byts = b''
                async for data in response.content.iter_chunked(1024*1024):
                    byts += data
                assert len(byts) == 50
```

Run with Python 3.7.4:
aiohttp==3.6.2
vcrpy==3.0.0

I also created a gist that runs through a few different scenarios:  https://gist.github.com/mikemoritz/01f7f222f9353846db16a6d07d98d806#file-vcrpytest-py

neozenith:
I'm not familiar with `aiohttp` so will have to rely on others to contribute a PR to address this.

Thank you for such a clear, well documented and repeatable issue. That means a lot!

neozenith:
@nickdirienzo any ideas or suggestions?

nickdirienzo:
Interesting! Off the top of my head, I'm not sure. I have some questions that might help — and I might be able to look into this tomorrow or next week actually.

@mikemoritz 
- Do you have a stack trace with what's failing?
- Has this worked in previous versions of vcrpy?
- Have you ran this with the env var `PYTHONASYNCIODEBUG=1` so `asyncio` code runs in debug mode? 

No worries if you don't know these off the top of your head, but this is where I would start. Thanks!

mikemoritz:
It doesn't "fail" per se, but rather returns 0 bytes.  However, I think I found the offender:

When recording the cassette the body is read:
```
async def record_response(cassette, vcr_request, response):
    """Record a VCR request-response chain to the cassette."""

    try:
        body = {"string": (await response.read())}

    # ...
```
https://github.com/kevin1024/vcrpy/blob/master/vcr/stubs/aiohttp_stubs/__init__.py#L119

[...]

PR #509, opened 2019-12-31
Port aiohttp unread_data to support stream reads

mikemoritz:
- Addresses #502 
- The original test actually identified the issue, so for this change it only needed to returned to the correct output
- A user will only encounter "unread" data on initial recording if they use the Streaming API, which fails today, so it leaves the primary aiohttp unchanged

### DIFF
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -117,7 +117,8 @@ def test_stream(tmpdir, scheme):
     url = scheme + "://httpbin.org/get"
 
     with vcr.use_cassette(str(tmpdir.join("stream.yaml"))):
-        resp, body = get(url, output="raw")  # Do not use stream here, as the stream is exhausted by vcr
+        resp, body = get(url, output="stream")
+        assert len(body) > 0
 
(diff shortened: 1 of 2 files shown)


mikemoritz:
FYI this change appears to be running into the Black/E203 conflict, and I opted for the Black style: https://github.com/psf/black#slices

martindurant:
+1 for finding and fixing this!
I just spent ages trying to understand why my streaming responses were empty when using vcrpy. 
Any chance of pushing it over the line?

kevin1024:
Gotta say this one is over my head, but there seems to be some interest.  Any way we can get CI to pass?

martindurant:
See my patches in https://github.com/dask/gcsfs/blob/master/gcsfs/tests/conftest.py , they may help you

vEpiphyte:
@mikemoritz were you ever able to check the CI failure on this?

kevin1024:
Closing this PR due to extended inactivity. There was interest in this fix but CI issues were never resolved. If aiohttp streaming issues persist, please open a fresh PR. Thanks for the contribution!

PR #511, opened 2020-01-13
Do not redirect on 304 for aiohttp

royjs:
Currently VCR tries to redirect on 304 (not modified) status codes. It causes a crash because the 'location' header is not set on 304s.

I added an exception for 304s in the redirection code.

### DIFF
--- a/vcr/stubs/aiohttp_stubs/__init__.py
+++ b/vcr/stubs/aiohttp_stubs/__init__.py
@@ -92,6 +92,9 @@ def play_responses(cassette, vcr_request):
     # If we're following redirects, continue playing until we reach
     # our final destination.
     while 300 <= response.status <= 399:
+        if "location" not in response.headers:
+            break
+
         next_url = URL(response.url).with_path(response.headers["location"])
 
         # Make a stub VCR request that we can then use to look up the recorded
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -273,6 +273,23 @@ def test_redirect(aiohttp_client, tmpdir):
     assert cassette_response.request_info.real_url == response.request_info.real_url
 
 
+def test_not_modified(aiohttp_client, tmpdir):
+    """It doesn't try to redirect on 304"""
+    url = "https://httpbin.org/status/304"
+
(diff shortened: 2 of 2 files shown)


vcr/stubs/aiohttp_stubs/__init__.py
```diff
@@ -91,7 +91,7 @@ def play_responses(cassette, vcr_request):
 
     # If we're following redirects, continue playing until we reach
     # our final destination.
-    while 300 <= response.status <= 399:
+    while 300 <= response.status <= 399 and response.status != 304:
         next_url = URL(response.url).with_path(response.headers["location"])
```
jamescurtin: It seems like there are other status codes that also don't contain a location header (_e.g._ `300` and `308`).

Perhaps instead of excluding `304`s in the `while` loop, you could check for the existence of `response.headers["location"]` before setting `next_url` and `break` if not set?

royjs: Yeah this way it covers all cases. I made the suggested change

neozenith:
Hi, I just wanted to update active PRs and issues to let you know I will no longer be an active maintainer on this project.

I recently changed jobs where I no longer use VCRpy or Python at all so I don't have the bandwidth anymore. I won't be monitoring issues or PRs either.

I will have to delegate to other maintainers or ask for some contributors to step up.

I know the Azure CLI uses VCRpy so it would be nice if they could spare some sprint time to maintain this project as it was invaluable in our CI when I was using it.

Kind regards,
Josh

jairhenrique:
@jamescurtin @royjs what needs to be done on this pull request to be merged?

royjs:
AFAIK this is ready to be merged

jairhenrique:
@kevin1024 can you help here?

IvanMalison:
Is the failure here https://travis-ci.org/github/kevin1024/vcrpy/jobs/703316572

related to this PR?

royjs:
It looks like the call to `https://httpbin.org/redirect/2` is currently returning a 404
See https://github.com/postmanlabs/httpbin/issues/617

chlowell:
@IvanMalison will this be backported to 3.0?

PR #546, opened 2020-07-27
aiohttp: Add support to allow_redirects option

pauloromeira:
Adds support to the `allow_redirects` options, which defaults to `True`, but in many cases we want to handle redirects manually:
> allow_redirects ([bool](https://docs.python.org/3/library/functions.html#bool)) – If set to False, do not follow redirects. True by default (optional).

https://docs.aiohttp.org/en/stable/client_reference.html


pauloromeira:
Hi @jairhenrique and @kevin1024, is this going to be merged? :sweat_smile:

pauloromeira:
> @pauloromeira can you rebase with `master`. Tests are now passing.

Thanks, @jairhenrique. Done!

PR #552, opened 2020-08-18
Handle 307 and 308 redirect method in httpx

kestutisbalt:
Request method should be preserved whenever response status is 307 or 308


vcr/stubs/httpx_stubs.py
```diff
@@ -108,7 +108,8 @@ def _play_responses(cassette, request, vcr_request, client, kwargs):
         if not next_url:
             break
 
-        vcr_request = VcrRequest("GET", next_url, None, dict(response.headers))
+        method = request.method if response.status_code != 303 else "GET"
```
2m, line 111: Its tricky to know what would be the best logic here because it is not strictly mandated by the HTTP spec.

However according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections the only status code that *always* changes to `GET` method is `303`. So this LGTM!

vcr/stubs/httpx_stubs.py
```diff
@@ -108,7 +108,11 @@ def _play_responses(cassette, request, vcr_request, client, kwargs):
         if not next_url:
             break
 
-        vcr_request = VcrRequest("GET", next_url, None, dict(response.headers))
+        # Even though 302 shouldn't change request's method, browsers like Chromium
```
IvanMalison, line 111: why does what browsers do here matter?

Issue #534, opened 2020-06-09
Change urlib2 to urilib.request in the usage page of docs


AthulMuralidhar:
- As we all know python 2 has reached its end of life this year (2020) and there are some libraries that changed names etc.
- In the usage page of vcrpy we see that the examples still use `urlib2` that has since changed. The link is here: https://vcrpy.readthedocs.io/en/latest/usage.html
- We need to change it to `urlib.request`as per python docs: https://docs.python.org/3.5/library/urllib.request.html#module-urllib.request

1oglop1:
you are more than welcome to PR this change I believe someone will be able to review it ;)

PR #554, opened 2020-08-19
Change urllib2 to ulrlib in accordance with Py 3.5

AthulMuralidhar:
- closes: #534 
- the change is required because the urllib2 is now moved to urllib in python 3.5

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -4,10 +4,10 @@ Usage
 .. code:: python
 
     import vcr
-    import urllib2
+    import urllib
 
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -4,10 +4,10 @@ Usage
 .. code:: python
 
     import vcr
-    import urllib2
```
graingert, line 7: hmm, we should probably lint our docs

https://pypi.org/project/flake8-rst/

AthulMuralidhar: should i do the linting?

PR #562, opened 2020-10-05
Fix tests

jairhenrique:
- Fix travis matrix. Current configuration don't run all tests (httpx is an example).
- Fix httpx integration for new version
- Fix tornado tests


jairhenrique:
@kevin1024 could you review and bump a new version if ok?

kevin1024:
released!

PR #566, opened 2020-10-13
Added assertion message for only the new request

cgorshing:
I'd like to have something displayed as an error message when the body fails to match. Just having the request alone is beneficial enough for me (in case there might be issues with showing both requests).


kevin1024:
Did you try turning on the debug logging?  This should have more information when the request fails to match:

https://github.com/kevin1024/vcrpy/blob/master/docs/debugging.rst

kevin1024:
If so - is there any way to add this additional information, but only with debug logging turned on?

Issue #398, opened 2018-09-30
filter_post_data_parameters not working with aiohttp


cinemascop89:
When trying to use the `filter_post_data_parameters` argument with the aiohttp client, vcrpy crashes with the folowing traceback:

```
Traceback (most recent call last):
  File "test.py", line 23, in <module>
    loop.run_until_complete(main())
  File "/usr/lib64/python3.6/asyncio/base_events.py", line 468, in run_until_complete
    return future.result()
  File "test.py", line 20, in main
    await do_request(client)
  File "test.py", line 12, in do_request
    async with client.post('https://httpbin.org/post', data=data) as response:
  File "/home/ego/devel/experiments/vcrpy_aiohttp/env/lib/python3.6/site-packages/aiohttp/client.py", line 855, in __aenter__
    self._resp = await self._coro
  File "/home/ego/devel/experiments/vcrpy_aiohttp/env/lib/python3.6/site-packages/vcr/stubs/aiohttp_stubs/__init__.py", line 57, in new_request
    if cassette.can_play_response_for(vcr_request):
  File "/home/ego/devel/experiments/vcrpy_aiohttp/env/lib/python3.6/site-packages/vcr/cassette.py", line 254, in can_play_response_for
    request = self._before_record_request(request)
  File "/home/ego/devel/experiments/vcrpy_aiohttp/env/lib/python3.6/site-packages/vcr/config.py", line 253, in before_record_request
    request = function(request)
  File "/home/ego/devel/experiments/vcrpy_aiohttp/env/lib/python3.6/site-packages/vcr/filters.py", line 100, in replace_post_data_parameters
    splits = [p.partition(b'=') for p in request.body.split(b'&')]
AttributeError: 'dict' object has no attribute 'split'
```

Bellow is the code used to produce this error:

```python
from pprint import pprint
import asyncio

import vcr
import aiohttp

async def do_request(client):
    data = {
        'user': 'foo',
        'password': 'bar',
    }
    async with client.post('https://httpbin.org/post', data=data) as response:
        print('status', response.status)
        pprint(await response.json())


async def main():
    with vcr.use_cassette('cassettes/test.yaml', filter_post_data_parameters=['password']):
        async with aiohttp.ClientSession() as client:
            await do_request(client)

loop = asyncio.get_event_loop()
loop.run_until_complete(main())

```

This was detected using vcrpy 2.0.1 and aiohttp 3.4.4

neozenith:
[...]

PR #572, opened 2021-01-25
Fix for the issue filter_post_data_parameters not working with aiohttp #398

vprakashplanview:
A test_replace_post_data_parameters_with_body_type_as_dict is added under /vcrpy/tests/unit/test_filters.py

### DIFF
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -153,6 +153,19 @@ def test_replace_post_data_parameters_empty_body():
     assert request.body is None
 
 
+def test_replace_post_data_parameters_with_body_type_as_dict():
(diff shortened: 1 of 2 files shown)


scop:
I implemented a fix for this before noticing this PR. However I don't think the implementation here is correct -- I believe it would not work appropriately when `filter_post_data_parameters` or `remove_post_data_parameters` delegates to it, nor it behaves as the API docs say with regards to replacements possibly being callables: https://vcrpy.readthedocs.io/en/latest/advanced.html#advanced-use-of-filter-headers-filter-query-parameters-and-filter-post-data-parameters

My version of the fix in #582 should behave appropriately with regards to these issues, and has some more unit test to showcase it.

PR #574, opened 2021-01-27
Fix not working for binary response in httpx

yhlam:
The `httpx_reponse.content.decode("utf-8", "ignore")` would corrupt binary content that can't decoded to unicode. We should serialize the binary content directly in that case.

A better way to fix this is probably is set the content to `response["body"]["string"]` because `vcr.serializers.compat.convert_to_unicode()` and `vcr.serializers.compat.convert_to_bytes()` handle the bytes / unicode conversion of that field. But that will change the cassette format for httpx. I guess you guys can consider that when implementing #463.

### DIFF
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -25,11 +25,20 @@ def _transform_headers(httpx_reponse):
 
 
 def _to_serialized_response(httpx_reponse):
+    content = httpx_reponse.content
+    try:
+        if content is not None and not isinstance(content, str):
+            content = content.decode("utf-8")
+    except (TypeError, UnicodeDecodeError, AttributeError):
+        # Sometimes the string actually is binary or StringIO object,
+        # so if you can't decode it, just give up.
+        pass
+
     return {
         "status_code": httpx_reponse.status_code,
         "http_version": httpx_reponse.http_version,
         "headers": _transform_headers(httpx_reponse),
-        "content": httpx_reponse.content.decode("utf-8", "ignore"),
+        "content": content,
     }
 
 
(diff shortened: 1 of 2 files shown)


vcr/stubs/httpx_stubs.py
```diff
@@ -25,11 +25,20 @@ def _transform_headers(httpx_reponse):
...
 
 def _to_serialized_response(httpx_reponse):
+    content = httpx_reponse.content
+    try:
+        if content is not None and not isinstance(content, str):
+            content = content.decode("utf-8")
+    except (TypeError, UnicodeDecodeError, AttributeError):
+        # Sometimes the string actually is binary or StringIO object,
+        # so if you can't decode it, just give up.
+        pass
```
romulocollopy, line 35: hey @yhlam , I think we could completely ignore this decode, as I did in my PR #583 

As you can see on [httpx](https://github.com/encode/httpx/blob/2d571046e15d3f4c4cbe5a68173cdf12e99cb3ca/httpx/_types.py#L77), the content attribute can be a `bytes` type.

romulocollopy (commented):
As far as I tested, `httpx.Response.content` can deal with bytes type. So it wouldn't be necessary to try to decode

vcr/stubs/httpx_stubs.py
yhlam: Yes, `httpx` can take `bytes`. That's why the `content` is encoded back to `bytes` in `_from_serialized_response` when constructing `httpx.Response`. The decode is for serializing the VCR response, not for `httpx`. If you take out the decode, I believe you will get `!!binary` in the serialized yaml even for plain text. I guess reason of decoding the data into unicode is trying to have human readable yaml file. This unicode / bytes conversion happens in other libraries as well, but it is done in for the [body](https://github.com/kevin1024/vcrpy/blob/c79a06f639dd628536c9868044e78df1012985f9/vcr/serializers/compat.py#L54) field only. Therefore, I suggest the better fix should be having a consistent serialization schema for all libraries, which is discussed in #463.

parkerhancock:

jairhenrique:
Fixed in https://github.com/kevin1024/vcrpy/pull/784

PR #576, opened 2021-02-14
persisters/filesystem_cached: Add file-based persister with cache

gioele:
This persister keeps a cache of cassettes, to avoid the overhead of constantly serializing and deserializing unchanged cassettes.

This is especially useful when no changes to the cassette are expected, for example when running testsuites with `mode=none`.

In a testsuite with hundreds of calls backed from YAML cassettes with 10k responses, the use of this persister saves various minutes.


IvanMalison (changes requested):
This isn't imported or tested anywhere

PR #581, opened 2021-04-15
fix(aiohttp): record body of first request in redirect chain

scop:
The first could be e.g. a POST with one, only the remaining ones are
expected to be GETs without one.

### DIFF
--- a/vcr/stubs/aiohttp_stubs.py
+++ b/vcr/stubs/aiohttp_stubs.py
@@ -175,14 +175,14 @@ async def record_responses(cassette, vcr_request, response):
     to the final destination.
     """
 
-    for past_response in response.history:
+    for i, past_response in enumerate(response.history):
         aiohttp_request = past_response.request_info
 
-        # No data because it's following a redirect.
         past_request = Request(
             aiohttp_request.method,
             str(aiohttp_request.url),
-            None,
+            # Record body of first request, rest are following a redirect.
+            None if i else vcr_request.body,
             _serialize_headers(aiohttp_request.headers),
         )
         await record_response(cassette, past_request, past_response)


scop:
Rebased.

IvanMalison:
@scop I'm not really familiar with the aiohttp code. How sure are you that this it the right thing?

@kevin1024 @scop Do you have any interest in helping out with maintainership?

I don't think anyone is very actively maintaining right now, and I really have not used vcrpy in years. I think we'd all really appreciate if you could step in.

scop:
I am very confident that this is the right thing to do. If not done, the request body of a POST that results in a redirect is recorded as `null`.

Notably, at least for example for `requests`, the body _is_ recorded.

Small reproducer:
```shellsession
$ cat test.py 
import asyncio

import aiohttp
import requests
import vcr

with vcr.use_cassette("redirect_requests.yaml"):
    response = requests.post("https://httpbingo.org/redirect-to",
                             params={"url": "https://vcrpy.readthedocs.io"},
                             data={"hello": "world"})

async def do_aiohttp():
        async with session.post("https://httpbingo.org/redirect-to",
                                params={"url": "https://vcrpy.readthedocs.io"},
                                data={"hello": "world"}) as response:
            text = response.text
with vcr.use_cassette("redirect_aiohttp.yaml"):
    asyncio.get_event_loop().run_until_complete(do_aiohttp())
$ python ./test.py
$ head -n 3 redirect_requests.yaml 
interactions:
- request:
    body: hello=world
$ head -n 3 redirect_aiohttp.yaml 
interactions:
- request:
    body: null
```

With regards to maintainership, I think there's a lot of value having vcrpy around, and it should be used a lot more than it currently is. In a sense I am interested, but I already have so much on my plate already that I cannot commit to being around to help out consistently, and I don't want to raise false hopes. My own vcrpy usage is currently a bit sporadic, too, but it's possible that'll change some in the foreseeable future. Anyway I'll keep this in mind.

scop:
Anything more I can do to help getting this merged?

kevin1024:
I gotcha.  Thanks for your contribution.

scop:
Thanks!

PR #583, opened 2021-04-22
enables sending non-ascii characters in httpx requests

romulocollopy:
The decode("utf-8") call in _make_vcr_request causes the request preparation to fail when non-ascii data is present in the binary.

Since The `vcr.request.Request` object is capable of checking if the body has a `read()` method and treat it according to since `7caf2973`, we can remove this call


jairhenrique:
@romulocollopy can you rebase your branch with master!

jairhenrique (approved):

PR #591, opened 2021-05-25
Fix compressed response from httpx

geraxe:
In case of httpx responce.body decompressed automatically by httpx. If we not remove "content-encoding" header here, vcr.filters.decode_response will try to decompress already decompressed body.


the-allanc:
This might be fixed in my pull request - #649.

parkerhancock:
Also fixred in #784

PR #594, opened 2021-07-05
Add the `before_handle_response` option

### LINKED ISSUE #544: before_record_response returns inconsistent data depending on recording or not
It's all in the title and seems to be related to #356
My tests depend on some sensitive data to be at the value I replaced it with.
So when I run the tests I have sometimes the real data (which I don't want) and sometimes the fake data actually in the cassette.

This is an inconsistent behaviour and, depending on what the data is used for afterwards, could lead to sensitive information being leaked when recording cassettes. For example, if the test logs, integration servers, etc.

akaihola:
Fixes #544


kevin1024:
Closing this draft PR due to extended inactivity. If there's still interest in this feature, a fresh PR would be welcome. Thanks for the contribution\!

PR #595, opened 2021-07-05
Add the `alter_live_response` option

### LINKED ISSUE #544: before_record_response returns inconsistent data depending on recording or not

akaihola:
Fixes #544.

### The problem
Currently callbacks defined using the `before_record_response` can modify responses, but this creates an inconsistency where e.g. unit test cases see unmodified responses when cassettes are being recorded, and modified ones when they are later played back.

### The solution
The `alter_live_response = True` option instructs VCRpy to apply response modifications also to the live response when recording cassettes:
```python
def vcr_config():
    return {
        "before_record_response": [my_callback],
        "alter_live_response": True,
    }
```

This is an alternative to the `before_handle_response` option implemented in #594.

- [ ] unit tests
- [ ] documentation
- [ ] change log

### DIFF
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -180,6 +180,7 @@ class Cassette:
         match_on=(uri, method),
         before_record_request=None,
         before_record_response=None,
+        alter_live_response=False,
         custom_patches=(),
         inject=False,
         allow_playback_repeats=False,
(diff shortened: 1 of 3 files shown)


akaihola:
I'll abandon this approach in favor of #594.

Issue #468, opened 2019-08-07
fails to re-establish a connection after keep-alive timout


evgeni:
vcrpy 2.0.1, requests 2.22.0

We are using requests with sessions to call an API endpoint to trigger a job and then call another endpoint in a loop to get the job status and thus wait until the job is done. The server is an Apache httpd with `KeepAliveTimeout 5` and a Rails app.

The code is working fine outside of VCR, but as soon as we enable VCR recording, the "wait" part of the code fails (`requests.exceptions.ConnectionError: ('Connection aborted.', BadStatusLine("''",))` on Python 2.7, `requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))` on Python 3.7)

Example code:

```python
import time

import urllib3
urllib3.disable_warnings()
import requests

import vcr

BASE = 'https://foreman.example.com'
STATUS = '{}/api/v2/status'.format(BASE)
DELETE = '{}/katello/api/organizations/10/subscriptions/delete_manifest'.format(BASE)


def wait(task):
    while task['state'] != 'stopped':
        print("sleeping for {}".format(task['id']))
        time.sleep(10)
        r = s.get("{}/foreman_tasks/api/tasks/{}".format(BASE, task['id']))
        task = r.json()

def delete():
    print("delete")
    r = s.post(DELETE, json={})
    print(r.headers)
    wait(r.json())

with vcr.use_cassette('katello.yaml'):
    s = requests.Session()
    s.verify = False
    s.auth = ('admin', 'changeme')

    delete()
```

output:
```
python test_katello_vcr.py
delete
[...]

PR #605, opened 2021-08-31
Fix patching of is_connection_dropped from urllib3 (issue #468)

peterisr:
Fix patching of is_connection_dropped from urllib3

With old patching method, urllib3 never detected TCP connections that
were closed by the server side. For example, persistent HTTP connection
that were closed by the server (e.g., due to timeout) were not
recognized as closed. Any following requests that attempted to reuse
the same, closed connection caused the following failure:

    urllib3.exceptions.ProtocolError: ('Connection aborted.',
       RemoteDisconnected('Remote end closed connection without response'
    ))

Fixes: https://github.com/kevin1024/vcrpy/issues/468

### DIFF
--- a/tests/integration/test_urllib3.py
+++ b/tests/integration/test_urllib3.py
@@ -2,6 +2,8 @@
 
 # coding=utf-8
 
+import http.server
+import multiprocessing
 import pytest
 import pytest_httpbin
 import vcr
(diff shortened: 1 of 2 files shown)


larstiq:
We also carry this as local patch, would be nice to see the issue fixed.

Issue #255, opened 2016-05-16
AttributeError: 'NoneType' object has no attribute 'decode' when 'body' in match_on, fixed with 'raw_body'


jhh3000:
When using vcr with the following configuration with 'body', there are errors with loading the JSON response -- it could be due to a blank response or blank payload. When I use 'raw_body' it works perfectly. Here's the stack trace (relevant sections):

```bash
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/requests/sessions.py", line 473, in get
    return self.request('GET', url, **kwargs)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/requests/sessions.py", line 461, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/requests/sessions.py", line 573, in send
    r = adapter.send(request, **kwargs)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/requests/adapters.py", line 370, in send
    timeout=timeout
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 518, in urlopen
    body=body, headers=headers)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 353, in _make_request
    httplib_response = conn.getresponse(buffering=True)
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/vcr/stubs/__init__.py", line 219, in getresponse
    if self.cassette.can_play_response_for(self._vcr_request):
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/vcr/cassette.py", line 232, in can_play_response_for
    return request and request in self and \
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/vcr/cassette.py", line 303, in __contains__
    for index, response in self._responses(request):
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/vcr/cassette.py", line 227, in _responses
    if requests_match(request, stored_request, self._match_on):
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/vcr/matchers.py", line 97, in requests_match
    matches = [(m(r1, r2), m) for m in matchers]
  File "/home/vagrant/venv/local/lib/python2.7/site-packages/vcr/matchers.py", line 80, in body
    return transformer(read_body(r1)) == transformer(read_body(r2))
[...]

PR #608, opened 2021-09-28
Fix for #255 for other content types

lenwhite:
The previous fix for #255 only worked with json responses. The error in #255 (`AttributeError: 'NoneType' object has no attribute 'decode' when 'body' in match_on, fixed with 'raw_body'`) still occurs with e.g. form-encoded data. 

Adding this check should fix that.

### DIFF
--- a/vcr/matchers.py
+++ b/vcr/matchers.py
@@ -70,15 +70,25 @@ def _transform_json(body):
         return json.loads(body.decode("utf-8"))
 
 
+def _transform_qs(body):
+    if body:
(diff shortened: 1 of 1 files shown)


jairhenrique:
This issue was fixed on #278

fdintino:
This was not fixed by #278. The author of the PR linked to the issue associated with #278, noting that the problem was only addressed for a JSON body, but that it persists for form-data. I just encountered this bug.

cedmt:
@fdintino is right, I wish this could be merged. I have the same issue with null body on the latest version.

PR #609, opened 2021-09-30
Call urlparse less frequently

mbrunthaler:
Currently, `urlparse` is invoked for all properties of uri in `vcr.request.Request` which has an impact on performance especially with lengthy cassettes. I've changed this to use a property such that `urlparse` is only invoked when `uri` is changed.


jairhenrique (approved):

jairhenrique:
Fixed on #655

Issue #571, opened 2021-01-18
Ability to not follow redirects for aiohttp


LilSpazJoekp:
In the code there is a comment indicating that there is a way to not follow redirects here: https://github.com/kevin1024/vcrpy/blob/c79a06f639dd628536c9868044e78df1012985f9/vcr/stubs/aiohttp_stubs.py#L124-L125 However, it appears that the code does not check to see if it is supposed to follow redirects or not. Would it be possible to add a way to not follow redirects?

PR #617, opened 2021-11-04
Support the allow_redirects option of aiohttp when playing responses

scotloach:
This is https://github.com/kevin1024/vcrpy/issues/571
In aiohttp the request method takes an optional "allow_redirects" parameter which can be set to False to disable redirects.


jairhenrique:
@scotloach can you write some tests for it?

PR #620, opened 2021-11-12
Enable brotli decompression if it is available

### LINKED ISSUE #599: support "conent-encoding: br"
Currently `vcrpy` doesn't decompress response in Brotli compression format, but Brotli is well supported in HTTP clients like `requests`, `httpx`.

https://github.com/kevin1024/vcrpy/blob/c79a06f639dd628536c9868044e78df1012985f9/vcr/filters.py#L127-L162

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#directives

immerrr:
This PR enables brotli decompression of responses if it is available (via `brotli`, `brotlipy` or `brotlicffi`)

I have added all three brotli variations to the test suite, but I'm at a bit of a loss, because `httpbin` brings in `brotlipy` unconditionally, so there is no real way to test:
- how vcrpy will behave without any brotli implementation
- how vcrpy will behave with a different brotli implementation (in theory, all three support the same `brotli.decompress(data)` API)

Happy to hear any feedback, but also feel free to make any adjustments that feel right (I have enabled edits from maintainers).

Fixes #599


immerrr:
@jairhenrique would you be able to look at this at your convenience?

jairhenrique (approved):

vcr/filters.py
```diff
@@ -6,6 +6,20 @@
...
+    try:
+        import brotlicffi as brotli
+    except ImportError:
+        brotli = None
+
+
+AVAILABLE_DECOMPRESSORS = {"gzip", "deflate"}
+if brotli is not None:
+    AVAILABLE_DECOMPRESSORS.add("br")
+
```
CharString, line 22: It is easier to expand with future or custom decompressors if `AVAILABLE_DECOMPRESSORS` is a `Dict[[str], Callable[[str], str]]`
```python
AVAILABLE_DECOMPRESSORS = {
    "brotli": brotli.decompress,
    "deflate": zlib.decompress,
    "gzip": lambda body: zlib.decompress(body, zlib.MAX_WBITS | 16),
}
```

Then `decompress_body` can just be
```python
def decompress_body(body, encoding):
    return AVAILABLE_DECOMPRESSORS[encoding](body)
```

Adding a new scheme will be as easy as adding a function to the dict.

CharString (changes requested):
Compression is an ongoing field of research. This suggestion will make it easier to adopt new algorithms.

finswimmer:
Hey,

what's the reason that this hasn't been merged until now? :thinking: 

fin swimmer

CharString (approved):

PR #644, opened 2022-05-05
Replace `assert` with `raise AssertionError` (fixes support for PYTHONOPTIMIZE=1)

neliseiska:
Assert seems to behave badly in some cases and should not be used for
handling runtime errors.
    
See for example:
https://medium.com/@jadhavmanoj/python-what-is-raise-and-assert-statement-c3908697bc62
https://github.com/emre/notes/blob/master/python/when-to-use-assert.md

### DIFF
--- a/vcr/matchers.py
+++ b/vcr/matchers.py
@@ -9,35 +9,43 @@ log = logging.getLogger(__name__)
 
 
 def method(r1, r2):
-    assert r1.method == r2.method, "{} != {}".format(r1.method, r2.method)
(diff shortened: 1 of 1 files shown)


hartwork (approved):
Hi @neliseiska,

I confirm that (1) without this pull request `export PYTHONOPTIMIZE=1` makes the current code fail and (2) with this pull requests those failing tests turn green. Hence I will merge this pull request. Thanks for your contribution to VCR.py! :pray: 

Best, Sebastian

PR #647, opened 2022-05-11
Add support for matching request body multipart form data

mparent61:
Request bodies of content type `multipart/form-data` can have arbitrary "boundary" separators, but contain identical data. This can cause VCRPy's "body" matcher to fail on otherwise identical requests.

This adds support for parsing the encoded body data, essentially stripping out the "boundary" values, and comparing the contents.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import os
 import re
 import sys
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 
 long_description = open("README.rst", "r").read()
@@ -47,6 +47,7 @@ install_requires = [
     "wrapt",
     "six>=1.5",
     "yarl",
+    "requests_toolbelt",
 ]
 
 setup(
--- a/tests/unit/test_matchers.py
+++ b/tests/unit/test_matchers.py
@@ -3,8 +3,7 @@ from unittest import mock
 
 import pytest
 
-from vcr import matchers
-from vcr import request
+from vcr import matchers, request
 
 # the dict contains requests with corresponding to its key difference
 # with 'base' request.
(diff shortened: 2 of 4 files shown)


mezhaka:
@kevin1024 Any plans for integrating this?

@mparent61 Is there any workaround right now? I have a multpart/form-data and I cannot use the cassete I record (which is actually a sending and recieving a small file).

mparent61:
@mezhaka - I've been using this workaround --
```
import re

import pytest
from requests_toolbelt.multipart import decoder
from vcr.matchers import _get_transformer, _identity
from vcr.util import read_body

def decode_multipart_request_body(request):
    return decoder.MultipartDecoder(
        content=request.body, content_type=request.headers["content-type"]
    )


def is_multipart_form_data_request(request):
    return re.match(
        r"multipart/form-data(;|$)", request.headers.get("content-type", "").lower()
    )

# This is a workaround for request body "multipart form" handling
def request_body_matcher(r1, r2) -> None:
    if is_multipart_form_data_request(r1) and is_multipart_form_data_request(r2):
        decoded1 = decode_multipart_request_body(r1)
        decoded2 = decode_multipart_request_body(r2)
        assert decoded1.encoding == decoded2.encoding
        assert len(decoded1.parts) == len(decoded2.parts)
        for part1, part2 in zip(decoded1.parts, decoded2.parts):
            assert part1.headers == part2.headers
            assert (
                part1.content == part2.content
            ), f"Multipart data doesn't match for part with content type: {part1.headers}"
    else:
        # Copied directly from VCRPy's `body` matcher. For some reason PyTest's custom ASSERT handlers (with
        # much better diagnostic info) does not work when the assert fires inside VCRPy.
        transformer = _get_transformer(r1)
        r2_transformer = _get_transformer(r2)
        if transformer != r2_transformer:
            transformer = _identity
        body1 = read_body(r1)
        body2 = read_body(r2)
        assert transformer(body1) == transformer(body2)

```

and then to setup - 
```
@pytest.fixture(scope="module")
def vcr(vcr):
    # Override VCRPy's default body matcher to support Multipart form data, and provide better ASSERT messages.
    vcr.register_matcher("body", request_body_matcher)
    return vcr
```

Issue #550, opened 2020-08-04
httpx cassetes not compatible with vcr format


geraxe:
httpx not worked from box 
if we create simple test 

```
import pytest
import httpx


@pytest.mark.vcr()
def test_iana():
    response = httpx.get('http://www.iana.org/domains/reserved')
    assert b'Example domains' in response.content

```
and run it 
`pytest --record-mode=once tests/test_vcr.py`

we have error with cassette file format
```
>           new_body = decompress_body(response["body"]["string"], encoding)
E           KeyError: 'body'

vcr/filters.py:158: KeyError
```
because cassette file format for httpx uses `content` parameter, but all other stubs use `body.string` parameters to save and filter result

and if we will use existing cassettes with httpx it won`t work for the same reason

geraxe:
#591

finswimmer:
Hey,

interesting that this issue is open for so long. We are replacing `requests` by `httpx` in more and more projects. So it would be great to get this issue fixed :smiley: 

Thanks a lot for this great project :+1:

kevin1024:
v8.0.0 completely rewrote httpx support (PR #943) to patch httpcore instead of httpx. The new implementation uses the standard cassette format with `body.string` (same as other HTTP libraries) and includes backward compatibility to read old httpx cassettes that used the `content` format. Could you please test with v8.0.0 and confirm cassette compatibility is resolved?

PR #649, opened 2022-06-12
Make httpx_stubs generate cassettes consistent with other stubs.

the-allanc:
Encountered this myself, but this was reported in #550.

I had generated some cassettes with _httplib2_, and then wanted to use them for _httpx_ - but they were entirely incompatible. This pull request resolves that.

But the cassette structure defined by _httpx_stubs_ was incompatible with other parts of the code - like the code to decode compressed responses. As such, I've changed the code such that it uses a set of response fields which are consistent with elsewhere in the codebase.

I've kept compatibility with the custom format that _httpx_stubs_ was using, so these cassettes can still be read - but no longer generated.

This pull request also includes two cassettes using the custom and regular formats, to show they can both be processed. I've based them on gzipped content, as this tests one of the complexities around how httpx was being used compared to other libraries (it seems content can only be retrieved in its decoded form, rather than the raw content given).

### DIFF
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -7,6 +7,8 @@ import httpx
 
 from vcr.errors import CannotOverwriteExistingCassetteException
 from vcr.request import Request as VcrRequest
+from vcr.filters import decode_response
+from vcr.serializers.compat import convert_body_to_bytes
 
 _httpx_signature = inspect.signature(httpx.Client.request)
 
@@ -62,17 +64,30 @@ def _from_serialized_headers(headers):
 @patch("httpx.Response.close", MagicMock())
 @patch("httpx.Response.read", MagicMock())
 def _from_serialized_response(request, serialized_response, history=None):
-    content = serialized_response.get("content")
-    if isinstance(content, str):
-        content = content.encode("utf-8")
+
+    # HTTPX cassette format.
+    if "status_code" in serialized_response:
+        serialized_response = decode_response(convert_body_to_bytes({
+            'headers': serialized_response['headers'],
+            'body': {'string': serialized_response['content']},
+            'status': {'code': serialized_response['status_code']},
+        }))
+        # We don't store the reason phrase in this format.
+        extensions = None
+
+    # Cassette format that all other stubs use.
+    else:
+        extensions = {"reason_phrase": serialized_response["status"]["message"].encode()}
+
     response = httpx.Response(
-        status_code=serialized_response.get("status_code"),
+        status_code=serialized_response["status"]["code"],
         request=request,
-        headers=_from_serialized_headers(serialized_response.get("headers")),
-        content=content,
+        headers=_from_serialized_headers(serialized_response["headers"]),
+        content=serialized_response["body"]["string"],
         history=history or [],
+        extensions=extensions,
     )
-    response._content = content
+
     return response
 
 
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -7,8 +7,8 @@ from unittest.mock import MagicMock, patch
 import httpx
 
 from vcr.errors import CannotOverwriteExistingCassetteException
-from vcr.request import Request as VcrRequest
 from vcr.filters import decode_response
+from vcr.request import Request as VcrRequest
 from vcr.serializers.compat import convert_body_to_bytes
 
 _httpx_signature = inspect.signature(httpx.Client.request)
@@ -37,7 +37,6 @@ def _transform_headers(httpx_response):
 
 
 async def _to_serialized_response(resp, aread):
-
     # The content shouldn't already have been read in by HTTPX.
     assert not hasattr(resp, "_decoder")
 
@@ -49,7 +48,7 @@ async def _to_serialized_response(resp, aread):
             resp.read()
 
     result = {
-        "status": dict(code=resp.status_code, message=resp.reason_phrase),
+        "status": {"code": resp.status_code, "message": resp.reason_phrase},
         "headers": _transform_headers(resp),
         "body": {"string": resp.content},
     }
@@ -59,6 +58,7 @@ async def _to_serialized_response(resp, aread):
     del resp._decoder
     resp._content = resp._get_content_decoder().decode(resp.content)
     return result
+
 
 def _from_serialized_headers(headers):
     """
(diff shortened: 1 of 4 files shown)


the-allanc:
I fixed the linting issues (using Black) and have force pushed the changes - can we get this pipeline run again?

vcr/stubs/httpx_stubs.py
```diff
@@ -27,17 +27,22 @@ def _transform_headers(httpx_response):
...
-def _to_serialized_response(httpx_response):
+def _to_serialized_response(resp):
     return {
-        "status_code": httpx_response.status_code,
-        "http_version": httpx_response.http_version,
-        "headers": _transform_headers(httpx_response),
-        "content": httpx_response.content.decode("utf-8", "ignore"),
+        "status": dict(code=resp.status_code, message=resp.reason_phrase),
+        "headers": _transform_headers(resp),
+        "body": {"string": resp.content},
```
hans2520, line 45: Question -- does this really generate the body content as a yaml binary string as in your[test example](https://github.com/kevin1024/vcrpy/pull/649/files#diff-e9c5392f3e558174781dc10a0f66b79455ef38846fdef971c9bb95cd70dd7d19) above? I believe that is just for reading back a cassette that has this type of content, but a new cassette generated with this stub will not be encoded the same way the others will be (or at least with requests library).

This is effectively what I needed to be able to generate the body content in the same yaml binary format:

```
from ..serializers.yamlserializer import serialize
...

def _to_serialized_response(resp):
    return {
        "status": dict(code=resp.status_code, message=resp.reason_phrase),
        "headers": _transform_headers(resp),
        "body": {"string": serialize(resp.content).encode("unicode_escape")},
    }
```

Can you confirm? And if not outright add this into PR, can we look at having an option we can set so this will be possible?

the-allanc: Perhaps I misunderstand.

Just to compare requests and httpx, I ran this in a Python prompt based on my branch:
```
with vcr.use_cassette('httpx_and_requests.yaml'):
    requests.get('https://httpbin.org/json')
    httpx.get('https://httpbin.org/json')
    requests.get('https://httpbin.org/gzip')
    httpx.get('https://httpbin.org/gzip')
```

Using that, I get [this output generated](https://gist.github.com/the-allanc/fcb9ea8c260050211cb198b198b1798a) - for the JSON resource, both httpx and requests generate a regular string (without the `!!binary` marker). For the GZIP resource, requests creates a binary string, but httpx does not.

I'm able to load the requests generated by the httpx stub with the requests library, and vice versa. If I retrieve binary content (like a JPEG image) using httpx, and capture that in a cassette, then I get the `!!binary` marker.

The only reason that the gzip resource differs between requests and httpx is because requests seems to allow you to get the raw data stream (which hasn't been decoded - i.e. un-gzipped), whereas I had trouble accomplishing that with httpx - so we store the decoded (un-gzipped) text and remove the Content-Encoding response header.

hans2520: I get your test, and I don't think you're misunderstanding anything.

Attempting to make httpx _not_ decompress is its own can of worms and isn't possible without modifying source. The upshot is that we can't generate cassettes with httpx that have binary encoded yaml like we did with requests -- and we considered that a feature, for various reasons.

So at least allowing the option on whether to always store content as binary encoded would be greatly appreciated!

the-allanc: @hans2520 OK, I've spent more time on this one...

After stepping into the httpx source code, I figured out that if you stream the response, and then manipulate the headers before httpx tries to process it, then you can get access to the raw data stream. So that's what I did.

I've made changes to try to prevent httpx reading a response, then reading the content without decoding, and then finally amending the response to still make it usable to regular clients that will want the content decoded normally.

httpx does still automatically read the content of responses on some occasions (particularly on redirects), so we still have to support the serialisation of decoded responses - so that code still exists in that form. I've added a new "client" class for _test_httpx_ which will automatically read responses to show that we handle decoded responses just as well as their non-decoded equivalents.

Hope you find this set of changes much more to your liking... :smile:

hans2520: Yes indeed, you are correct about manipulating the httpx headers (I don't think streaming was a requirement actually, but you would have needed to patch two different locations in my test... which was over a week ago, sorry can't remember exactly. But for the streaming -- if it works, why not).

I haven't had a chance your modification just yet, but just glancing at the commit you made and inferring from your comment, the only thing I'm wondering about is if the format of the response will actually be "raw" content (i.e. unicode character sequences) vs the yaml binary format. That's what happened in my testing when I modified the httpx source so as to not decode the response, and why I ultimately preferred not doing that but making the suggestion above to use the yaml serializer to get yaml binary format. 

If that's true, the one nice thing about yaml binary format is plays with margins nicely. But the main reason we were wanting it is so that in the end we would not have to regenerate a bunch of cassette formats we had in source control, theoretically if only the client changed but the response stayed the same (and it was still yaml binary encoded), none of that responses would change...

the-allanc: The format of the cassettes that are generated for httpx seem to be the same as the ones generated by requests (at least with my latest set of changes).

the-allanc:
Any chance to get this merged in?

07pepa:
i suggest use pydantic to validate casetes each time you load them ... will be much more secured 

aka some validation needs to be done othervise incompatibility will creap in

07pepa:
also please test uploads across bord to not left it out

i have  PR for it but i am not shure if it is right 

btw i just spinned fastapi server to record and later replay casete 

pr is here in upload cassete
https://github.com/kevin1024/vcrpy/pull/657

the-allanc:
I've updated my changes so that there are no longer merge conflicts (since my original changes were over a year old). Perhaps this could be reconsidered to be merged in?

jairhenrique:
@the-allanc can you check this https://github.com/kevin1024/vcrpy/actions/runs/5733507649/job/15553772124?pr=649

the-allanc:
> @the-allanc can you check this https://github.com/kevin1024/vcrpy/actions/runs/5733507649/job/15553772124?pr=649

I've fixed the typo in the comment.

the-allanc:
Everything has been resolved here.

jairhenrique:
LGTM.
@hartwork can you help with this review?

hartwork:
@jairhenrique sorry, I lack bandwidth for this, I need to cut down my time on VCR.py for now.

the-allanc:
I'll flag this one up again - I've already rebased the original MR to make it compatible, and I can see there are other MRs which are attempting to address issues around _httpx_ (I think some of which have already been addressed in this MR), so I would like this one to get put in place before I then have to rebase it further.

estahn:
@kevin1024 Can we get this merged?

svdgraaf:
Yeah, this would be a great feature to add. Thanks for the work @the-allanc!

the-allanc:
This is now the third version of this pull request. I've had to rebase it twice because other changes have been pulled into the main branch before this one - including one which made significant changes to how the HTTPX stub was implemented, meaning the code was incompatible with what was done before.

Could I please finally get this merged in - the changes make it possible to:
1. Use HTTPX with cassettes generated through other HTTP clients.
2. That new cassettes generated via HTTPX can be manipulated with the standard set of filters, serializers and other utilities that vcrpy provides.
3. Is still able to read cassettes previously generated by HTTPX in older versions of vcrpy.

the-allanc:
@jairhenrique As you merged in the other recent pull requests relating to HTTPX (#784 and #788), maybe you're in the best place to decide on this one?

jairhenrique (approved):

PR #655, opened 2022-09-15
Improve performance by caching urlparse

ADR-007:
`urlparse` took most of the time in my tests. So, those simple changes speed up tests several times in my case.


deronnax:
@kevin1024 I noticed it's a change that comes back often, it was already requested in https://github.com/kevin1024/vcrpy/pull/609. Let's get it merged?

jairhenrique (approved):

Issue #656, opened 2022-09-23
[BUG] HTTPX Binary upload data unsuported


07pepa:
I am trying to migrate one library from requests to http and it seems httpx portion have some bugs

folowing casete cant be used with httpx
https://github.com/billdeitrick/pypco/blob/master/tests/cassettes/TestPublicRequestFunctions.test_upload_and_use_file.yaml

because it tries to decode the binary string (that is nonsense)

```
Error
Traceback (most recent call last):
  File "pypco\pypco\pco.py", line 251, in request_response
    response = self._do_url_managed_request(method, url, payload, upload, **params)
  File "pypco\pypco\pco.py", line 223, in _do_url_managed_request
    return self._do_ratelimit_managed_request(method, url, payload, upload, **params)
  File "pypco\pypco\pco.py", line 184, in _do_ratelimit_managed_request
    response = self._do_timeout_managed_request(method, url, payload, upload, **params)
  File "pypco\pypco\pco.py", line 145, in _do_timeout_managed_request
    return self._do_request(method, url, payload, upload, **params)
  File "pypco\pypco\pco.py", line 110, in _do_request
    response = self.session.request(
  File ".virtualenvs\pypco-EZZlDlNc\lib\site-packages\httpx\_client.py", line 815, in request
    return self.send(request, auth=auth, follow_redirects=follow_redirects)
  File ".virtualenvs\pypco-EZZlDlNc\lib\site-packages\vcr\stubs\httpx_stubs.py", line 168, in _inner_send
    return _sync_vcr_send(cassette, real_send, *args, **kwargs)
  File ".virtualenvs\pypco-EZZlDlNc\lib\site-packages\vcr\stubs\httpx_stubs.py", line 155, in _sync_vcr_send
    vcr_request, response = _shared_vcr_send(cassette, real_send, *args, **kwargs)
  File ".virtualenvs\pypco-EZZlDlNc\lib\site-packages\vcr\stubs\httpx_stubs.py", line 81, in _shared_vcr_send
    vcr_request = _make_vcr_request(real_request, **kwargs)
  File ".virtualenvs\pypco-EZZlDlNc\lib\site-packages\vcr\stubs\httpx_stubs.py", line 72, in _make_vcr_request
    body = httpx_request.read().decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 137: invalid start byte
```

llybin:
[...]

PR #657, opened 2022-09-24
Fix httpx upload

07pepa:
- removed unneeded decoding of body in httpx_stubs.py
- added asset to upload
- added equivalent cassette for upload
- added testing to cover uploading to vcr with httpx
- fixes https://github.com/kevin1024/vcrpy/issues/656


parkerhancock:
Fixed in #784

parkerhancock:

PR #659, opened 2022-09-27
disallowing writes only when fully matched request is in cassette

miki725:
`cassette.filter_request` only filters some request attributes like headers but does not take into account any of the `match_on` configurations. As a result if a similar request has been seen before but does not fully match as per `match_on`, writes will be disallowed. `cassette.__contains__` however fully takes into account all `match_on` matchers and therefore will only guard against recording to the cassette when the same request has been previously recorded


miki725:
if this looks reasonable I can work on adjusting/adding tests as necessary but would appreciate any feedback first

PR #667, opened 2022-10-26
add python 3.11 support

evgeni:


evgeni:
(this should be plain formality, the code seems to work on 3.11)

armingerten:
Unfortunately, the code is not working for me because I am unable to install some of the transitive dependencies.
This is depending on https://github.com/aio-libs/yarl/issues/781 which is depending on https://github.com/aio-libs/multidict/issues/778 .

evgeni:
But doesn't the unavailability of wheels only mean you need to build those from source?

CI seems to be happy to do so.

PR #668, opened 2022-10-30
Record on exception (continuation of PR #237)

### LINKED ISSUE #205: Don't save requests from decorated tests if decorated test fails

Terseus:
Fixes #205 
Closes #237

This is a continuation of the PR #237.
Given that the author @dan-passaro didn't answered in 3 years, I considered a good idea to take over their branch and make it able to be merged.

@neozenith since you reviewed #237 I'll await for your comments.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -404,3 +404,25 @@ the Cassette ``allow_playback_repeats`` option.
         for x in range(10):
             response = urllib2.urlopen('http://www.zombo.com/').read()
         assert cass.all_played
+
+Discards Cassette on Errors
+---------------------------
+
+By default VCR will save the cassette file even when there is any error inside
+the enclosing context/test.
+
+If you want to save the cassette only when the test succeedes, set the Cassette
+``record_on_exception`` option to ``False``.
+
+.. code:: python
+
+    try:
+        my_vcr = VCR(record_on_exception=False)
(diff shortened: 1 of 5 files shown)


neozenith:
I’ll delegate review to @jairhenrique as they seem to be the most recent active contributor. 

Unfortunately I have other priorities come up and I’m no longer involved in this project sorry.

jairhenrique:
@Terseus please, update the docs with this new param.

Terseus:
Of course, I'll look into it.

Also will check what's happening with the pipeline, it worked on my machine™.

Terseus:
Hi @jairhenrique

It turns out, it didn't worked in my machine after all :)
I had a bug in my setup and the tests for pypy3 were skipped.

After that found out that the fail was because a bug introduced by the `record_on_exception` implementation in Pypy, it should be fixed now.
Also I'd added a small section on how to use it in the "Advanced Features" page.

jairhenrique:
@Terseus I fix the CI error. Can you rebase with master again!?

Terseus:
@jairhenrique Sure, done.

jairhenrique:
@Terseus thanks!

PR #680, opened 2023-01-22
Allow for changing patcher builder class

amosjyng:
I have a project that uses its own `CassettePatcherBuilder` class to record interactions of a different sort than vcrpy does. However, right now this requires [a bit of a hack](https://github.com/amosjyng/vcr-langchain/blob/df45d3f784f93144f1ba8ee9517f7b55de134218/vcr_langchain/cassette.py#L18) to get around the hardcoded usage of `CassettePatcherBuilder` in the `_patch_generator` function. This is not great for future forward compatibility since my project will not be up to date on any future changes to `_patch_generator`, such as the most recent one where the function no longer saves cassettes because that is handled by `__exit__` instead.

Would it be possible to make the `CassettePatcherBuilder` class more configurable?

### DIFF
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -58,12 +58,13 @@ class CassetteContextDecorator:
     def __init__(self, cls, args_getter):
         self.cls = cls
         self._args_getter = args_getter
+        self._patcher_builder_cls = CassettePatcherBuilder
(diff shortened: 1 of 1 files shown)


amosjyng:
NVM, I ended up doing things differently :P

PR #681, opened 2023-01-22
For custom cassette persisters no longer catch ValueError

amosjyng:
I had to spend some time digging before I realized that serialization/deserialization errors were being swallowed here

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -136,7 +136,7 @@ Create your own persistence class, see the example below:
 
 Your custom persister must implement both ``load_cassette`` and ``save_cassette``
 methods.  The ``load_cassette`` method must return a deserialized cassette or raise
-``ValueError`` if no cassette is found.
+``CassetteNotFoundError`` if no cassette is found.
 
 Once the persister class is defined, register with VCR like so...
 
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -136,7 +136,8 @@ Create your own persistence class, see the example below:
 
 Your custom persister must implement both ``load_cassette`` and ``save_cassette``
 methods.  The ``load_cassette`` method must return a deserialized cassette or raise
-``CassetteNotFoundError`` if no cassette is found.
+either ``CassetteNotFoundError`` if no cassette is found, or ``CassetteDecodeError``
+if the cassette cannot be successfully deserialized.
 
 Once the persister class is defined, register with VCR like so...
 
--- a/tests/integration/test_register_persister.py
+++ b/tests/integration/test_register_persister.py
@@ -5,9 +5,11 @@
 import os
 from urllib.request import urlopen
 
+import pytest
+
 # Internal imports
 import vcr
-from vcr.persisters.filesystem import FilesystemPersister
+from vcr.persisters.filesystem import CassetteDecodeError, CassetteNotFoundError, FilesystemPersister
 
 
 class CustomFilesystemPersister:
@@ -23,6 +25,19 @@ class CustomFilesystemPersister:
     def save_cassette(cassette_path, cassette_dict, serializer):
         cassette_path += ".test"
         FilesystemPersister.save_cassette(cassette_path, cassette_dict, serializer)
+
+
+class BadPersister(FilesystemPersister):
+    """A bad persister that raises different errors."""
+
+    @staticmethod
+    def load_cassette(cassette_path, serializer):
+        if "nonexistent" in cassette_path:
+            raise CassetteNotFoundError()
+        elif "encoding" in cassette_path:
+            raise CassetteDecodeError()
+        else:
+            raise ValueError("buggy persister")
 
 
 def test_save_cassette_with_custom_persister(tmpdir, httpbin):
@@ -53,3 +68,22 @@ def test_load_cassette_with_custom_persister(tmpdir, httpbin):
     with my_vcr.use_cassette(test_fixture, serializer="json"):
         response = urlopen(httpbin.url).read()
         assert b"difficult sometimes" in response
+
+
+def test_load_cassette_with_custom_persister(tmpdir, httpbin):
+    """
+    Ensure expected errors from persister are swallowed while unexpected ones
+    are passed up the stack.
+    """
+    my_vcr = vcr.VCR()
+    my_vcr.register_persister(BadPersister)
+
+    with my_vcr.use_cassette("bad/nonexistent") as cass:
+        assert len(cass) == 0
+
+    with my_vcr.use_cassette("bad/encoding") as cass:
+        assert len(cass) == 0
+
+    with pytest.raises(ValueError):
+        with my_vcr.use_cassette("bad/buggy") as cass:
+            pass
--- a/tests/integration/test_register_persister.py
+++ b/tests/integration/test_register_persister.py
@@ -70,10 +70,10 @@ def test_load_cassette_with_custom_persister(tmpdir, httpbin):
         assert b"difficult sometimes" in response
 
 
-def test_load_cassette_with_custom_persister(tmpdir, httpbin):
+def test_load_cassette_persister_exception_handling(tmpdir, httpbin):
     """
     Ensure expected errors from persister are swallowed while unexpected ones
-    are passed up the stack.
+    are passed up the call stack.
     """
     my_vcr = vcr.VCR()
     my_vcr.register_persister(BadPersister)
--- a/tests/unit/test_cassettes.py
+++ b/tests/unit/test_cassettes.py
@@ -34,12 +34,12 @@ def test_cassette_load_nonexistent():
     assert len(a_cassette) == 0
 
 
-def test_cassette_load_invalid_unicode(tmpdir):
-    a_file = tmpdir.join("invalid_unicode.yml")
+def test_cassette_load_invalid_encoding(tmpdir):
+    a_file = tmpdir.join("invalid_encoding.yml")
     with open(a_file, "wb") as fd:
         fd.write(b"\xda")
-    with pytest.raises(UnicodeDecodeError):
-        Cassette.load(path=a_file)
+    a_cassette = Cassette.load(path=str(a_file))
+    assert len(a_cassette) == 0
 
 
 def test_cassette_not_played():
--- a/tests/unit/test_cassettes.py
+++ b/tests/unit/test_cassettes.py
@@ -29,6 +29,19 @@ def test_cassette_load(tmpdir):
     assert len(a_cassette) == 1
 
 
+def test_cassette_load_nonexistent():
+    a_cassette = Cassette.load(path="something/nonexistent.yml")
+    assert len(a_cassette) == 0
+
+
+def test_cassette_load_invalid_unicode(tmpdir):
+    a_file = tmpdir.join("invalid_unicode.yml")
+    with open(a_file, "wb") as fd:
+        fd.write(b"\xda")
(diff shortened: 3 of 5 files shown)


hartwork (commented):
@amosjyng I see how this change could help on an _abstract_ level. Could you elaborate what _concrete_ practical problem this is fixing for you? What made you want to change this code? Thank in advance! :pray:

amosjyng:
IIRC I was trying to save different keys to the recorded file, but `deserialize(data, serializer)` was throwing a ValueError because the keys were not as expected. vcrpy would continually treat the session recording as if it weren't there at all, and it took a little while to debug that it was due to this mismatch between how the session was serialized vs how it was deserialized. Hence, this PR as a suggestion to make such development easier in the future.

hartwork:
Hi @amosjyng,

thanks for elaborating!

I just found that the documentation at https://vcrpy.readthedocs.io/en/latest/advanced.html#register-your-own-cassette-persister says that…

> [t]he `load_cassette` method must return a deserialized cassette or raise `ValueError` if no cassette is found.

I think that means that changing `except ValueError:` would break documented API. If that understanding is correct it would mean that we either (a) cannot change it or (b) need a major version bump to clearly communicate this change.

So we probably need to clarify that first — @jairhenrique  @kevin1024 what do you think, does this break the API?

Best, Sebastian

---
Notes to self for later:
- rebase onto latest `master` to get the CI involved
- class name `ExpectedCassetteError` not ideal?
- add a simple test using a mock custom cassette persistor

kevin1024:
1. Yes, this would be a breaking change to a documented API and would require a major version bump
2. Yeah, this probably would be a quality-of-life improvement for those using vcrpy.

My guess is that most folks have *not* built their own persister, so it should only affect a minority of vcrpy users.  Has anyone built their own persister?  Honestly IIRC I just thought it was a cool idea for someone to store cassettes in a database or s3 or something but I haven't heard of anyone using it.

hartwork:
@kevin1024 to be sure I read (2) right: You're pro fixing this and the major version bump?

kevin1024:
Yeah,  but this single change alone probably isn’t worth releasing a version bump in and of itself.  Maybe we can batch it up with some other breaking changes like dropping older models python versions and/or old http library versions and do a major bump sometime in the future. What do you think?

hartwork:
> Yeah, but this single change alone probably isn’t worth releasing a version bump in and of itself. Maybe we can batch it up with some other breaking changes like dropping older models python versions and/or old http library versions and do a major bump sometime in the future. What do you think?

@kevin1024 if we go the collect-more-breakage route, we either need a second master branch or we lose instant releasability on the master branch. The latter is a clear non-option to me and former is probably overkill and may hold this for later for an unclear amount of time. Given that the breakage is easy to adjust to, will likely not affect most users (as you mentioned), and will likely show directly where used in the test suite, I would vote for the simple approach and just treat this like a normal pull request except with bumping the version to 5.0.0, and continue as usual. If you like, we can merge #695 dropping Python 3.7 right after and call that the release, if you feel it's in the breaking bucket (— personally, I don't see it as breaking but it can be argued). What do you think?

kevin1024:
Excellent points, all!  I agree with your plan.  The only thing this PR needs still is an update to the documentation to mention the new API (different exception)

hartwork:
> Excellent points, all! I agree with your plan.

@kevin1024 cool!

> The only thing this PR needs still is an update to the documentation to mention the new API (different exception)

Yes!

I believe in total (integrating my earlier todo-notes from https://github.com/kevin1024/vcrpy/pull/681#issuecomment-1581749936 above) we need:
- a rebase onto latest `master` (e.g. to bring the CI in)
- probably two(?) new exception classes to cleanly distinct the two likely major/common cases for custom persisters, e.g.
  - `CassetteNotFoundError`, potentially inheriting from built-in `FileNotFoundError`
  - `CassetteDecodeError`, potentially inheriting from `ValueError`
- a simple test per new exception
- doc fixes at https://github.com/kevin1024/vcrpy/blob/f4316d2daea94a3baaad2aaac443c97548843e59/docs/advanced.rst?plain=1#L139

@amosjyng @kevin1024 what do you think?

PS: I can help out with any of these parts as needed.

amosjyng:
Sounds great to me! There's no rush on my end, we can wait for whenever the next major version is due. I just opened this PR as a suggestion to hopefully save others in the future some time :)

hartwork:
Hi @amosjyng I think it's likely that (driven by #695) there will be a natural next release before ~2023-07-01. Let's get started on the `ValueError` thing then so that we're done by then, I'd be happy to review the next iteration.

hartwork:
@amosjyng any news on the implementation?

amosjyng:
Sorry, I thought you guys were talking to each other about making those changes, not to me :P

I've split it into `CassetteNotFoundError` and `CassetteDecodeError` as requested. I've added tests for a custom persister that throws those errors, as well as a test for the default persister throwing a `CassetteNotFoundError`. I'm not sure how to get the default persister to throw the `CassetteDecodeError` as it's not clear to me how you'd trigger a `UnicodeEncodeError` while reading a file; should we perhaps be checking for `UnicodeDecodeError` as well?

docs/advanced.rst
```diff
@@ -136,7 +136,7 @@ Create your own persistence class, see the example below:
 
 Your custom persister must implement both ``load_cassette`` and ``save_cassette``
 methods.  The ``load_cassette`` method must return a deserialized cassette or raise
-``ValueError`` if no cassette is found.
+``CassetteNotFoundError`` if no cassette is found.
```
hartwork, line 139: This should mention `CassetteDecodeError` also somehow or no one will know about that exception with reading the internal code.

tests/integration/test_register_persister.py
```diff
@@ -53,3 +68,22 @@ def test_load_cassette_with_custom_persister(tmpdir, httpbin):
     with my_vcr.use_cassette(test_fixture, serializer="json"):
         response = urlopen(httpbin.url).read()
         assert b"difficult sometimes" in response
+
+
+def test_load_cassette_with_custom_persister(tmpdir, httpbin):
+    """
+    Ensure expected errors from persister are swallowed while unexpected ones
+    are passed up the stack.
```
hartwork, line 76: Could you make "stack" be "call stack"? I think it would make it a tiny bit more clear.

vcr/persisters/filesystem.py
```diff
@@ -5,17 +5,25 @@
...
         cassette_path = Path(cassette_path)  # if cassette path is already Path this is no operation
         if not cassette_path.is_file():
-            raise ValueError("Cassette not found.")
+            raise CassetteNotFoundError()
         try:
             with cassette_path.open() as f:
                 data = f.read()
         except UnicodeEncodeError as err:
-            raise ValueError("Can't read Cassette, Encoding is broken") from err
+            raise CassetteDecodeError("Can't read Cassette, Encoding is broken") from err
```
hartwork, line 26: I believe `except UnicodeEncodeError as err:` needs to be `except UnicodeDecodeError as err:`, instead.

(Even if loading were to decode and re-encoded certain bytes, I believe it's unlikely that that re-encoding would fail, if decoding worked the first time. So probably we don't need to catch both but only `UnicodeDecodeError`. What do you think?)

tests/integration/test_register_persister.py
```diff
@@ -53,3 +68,22 @@ def test_load_cassette_with_custom_persister(tmpdir, httpbin):
     with my_vcr.use_cassette(test_fixture, serializer="json"):
         response = urlopen(httpbin.url).read()
         assert b"difficult sometimes" in response
+
+
+def test_load_cassette_with_custom_persister(tmpdir, httpbin):
```
hartwork, line 73: How about a rename to `test_load_cassette_persister_exception_handling`? It think something like that would be a bit more clear and in line with the docstring.

tests/unit/test_cassettes.py
```diff
@@ -29,6 +29,19 @@ def test_cassette_load(tmpdir):
...
+    a_cassette = Cassette.load(path="something/nonexistent.yml")
+    assert len(a_cassette) == 0
+
+
+def test_cassette_load_invalid_unicode(tmpdir):
+    a_file = tmpdir.join("invalid_unicode.yml")
+    with open(a_file, "wb") as fd:
+        fd.write(b"\xda")
+    with pytest.raises(UnicodeDecodeError):
+        Cassette.load(path=a_file)
```
hartwork, line 42: With `load_cassette` fixed to handle `UnicodeDecodeError` I believe this test should then be:

```python
assert Cassette.load(path=a_file) is None
```

(with no more `pytest.raises`)

tests/unit/test_cassettes.py
```diff
@@ -29,6 +29,19 @@ def test_cassette_load(tmpdir):
...
 
+def test_cassette_load_nonexistent():
+    a_cassette = Cassette.load(path="something/nonexistent.yml")
+    assert len(a_cassette) == 0
+
+
+def test_cassette_load_invalid_unicode(tmpdir):
+    a_file = tmpdir.join("invalid_unicode.yml")
+    with open(a_file, "wb") as fd:
+        fd.write(b"\xda")
```
hartwork, line 40: Just a detail: I would prefer wording "invalid encoding" (or "malformed encoding") over "invalid Unicode" here, to be more precise.

hartwork (changes requested):
@amosjyng thanks for this, looking pretty good already, nice! :+1: 

I was wondering whether `UnicodeEncodeError` should actually be `UnicodeDecodeError` instead also — by now I think it's indeed mixed up. Please see my comments on details below:

docs/advanced.rst
amosjyng: Done! Feel free to change the wording further.

vcr/persisters/filesystem.py
amosjyng: Agreed, done!

tests/integration/test_register_persister.py
amosjyng: Done!

tests/integration/test_register_persister.py
amosjyng: Done! I just realized I forgot to rename the function after copy-pasting, so this fixes the name collision as well.

tests/unit/test_cassettes.py
amosjyng: A Cassette is still returned, but with zero records, same as the case where the file doesn't exist. I've modified it to be

```python
    a_cassette = Cassette.load(path=str(a_file))
    assert len(a_cassette) == 0
```

to keep in line with the asserts from `test_cassette_load` and `test_cassette_load_nonexistent`

tests/unit/test_cassettes.py
amosjyng: Good point, done!

hartwork (approved):
@amosjyng thanks for the adjustments, looks great! :+1: 

I have some hope that @kevin1024 is with me on releasing 4.4.0 shortly **before** the merge (i.e. issue #724). Let's see what he thinks; approving the pull request now…

hartwork:
@amosjyng you probably saw https://github.com/kevin1024/vcrpy/releases/tag/v5.0.0 but… your documentation changes are also live at https://vcrpy.readthedocs.io/en/latest/advanced.html#register-your-own-cassette-persister now. Thanks again! :pray:

amosjyng:
Woohooo, that's cool to see! Thanks @hartwork ! :)

Issue #174, opened 2015-07-16
mutating requests in before_record_request mutates the actual request returned to the code


graingert:

kevin1024:
Does it also affect the request sent to the server?  Should only affect the response that is recorded.

Do you mean if you grab the cassette?  `with vcr.use_cassett('foo.yml') as cass` and then do `cass.request`?

graingert:
Yes it effects the  request to the server. You're missing a deepcopy

IvanMalison:
This is by design. I'm pretty sure that @kevin1024 is correct that before_record_request is only applied after the real request is made (https://gist.github.com/99942d7e60364ae3c11a the call to append is where before_record_request happens).

What IS true is that if you were to pass a body object to a some library and then mutate that object in before_record_request, it would come back as mutated to whomever is making the request. I can see an argument that the request/response should be deep copied before this callbacks are applied.

graingert:
Yes that's what I mean

> What IS true is that if you were to pass a body object to a some library and then mutate that object in before_record_request, it would come back as mutated to whomever is making the request. I can see an argument that the request/response should be deep copied before this callbacks are applied.

IvanMalison:
@graingert I'd welcome a fix that does a deepcopy if that is what you want. I'm probably going to do a release tomorrow after I've given people some time to look at what I did for #177. LMK if you want this to make it in to that...

IvanMalison:
I actually don't think that this is an issue...
request copies headers, and the body is required to be a string by the point it gets there. Since it doesn't modify the request it makes to the server I don't see where there is a problem here.

graingert:
This definitely caught me out. Perhaps an immutable dict or namedtuple would be better than deep copy?

neozenith:
A lot of changes have happened to VCRpy since this ticket was opened. As this ticket has become stale, would you mind closing it if it is no longer needed / relevant? 

If I haven't heard anything in a week I'll mark it as closed as we have a lot of old tickets that need to be groomed to make it easier to see what is still relevant.

However if it is still needed, please feel free to re-open or create a new ticket.

Thanks! 🙏

abramclark:
I just ran into this as well. I tried to prevent all headers from being recorded with this simple code:

`
def before_record(req):
    req.headers.clear()
    return req

VCR = vcr.VCR(before_record_request=before_record)
`

[...]

PR #687, opened 2023-04-28
Fix for #174

abramclark:
Naive initial fix for
https://github.com/kevin1024/vcrpy/issues/174

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,8 +5,7 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - "*"
+  workflow_dispatch:
 
 jobs:
   build:
--- a/tests/integration/test_filter.py
+++ b/tests/integration/test_filter.py
@@ -145,7 +145,7 @@ def test_decompress_regular(tmpdir, httpbin):
 
 
 def test_before_record_request_corruption(tmpdir, httpbin):
-    """Modifying request in before_record_request should not effect outgoing request"""
+    """Modifying request in before_record_request should not affect outgoing request"""
     def before_record(request):
         request.headers.clear()
         request.body = b''
--- a/tests/integration/test_filter.py
+++ b/tests/integration/test_filter.py
@@ -142,3 +142,23 @@ def test_decompress_regular(tmpdir, httpbin):
         resp = urlopen(url).read()
         assert_cassette_has_one_response(cass)
     assert_is_json(resp)
+
+
+def test_before_record_request_corruption(tmpdir, httpbin):
+    """Modifying request in before_record_request should not effect outgoing request"""
+    def before_record(request):
+        request.headers.clear()
+        request.body = b''
+        return request
+
+    req = Request(
+        httpbin.url + "/post",
+        data=urlencode({'test': 'exists'}).encode(),
+        headers={'X-Test': 'exists'},
+    )
+    cass_file = str(tmpdir.join("modified_response.yaml"))
+    with vcr.use_cassette(cass_file, before_record_request=before_record):
+        resp = json.loads(urlopen(req).read())
+
+    assert(resp['headers']['X-Test'] == 'exists')
+    assert(resp['form']['test'] == 'exists')
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -266,6 +266,9 @@ class Cassette:
                 yield index, response
 
     def can_play_response_for(self, request):
+        # Deepcopy is here because mutation of `request` will corrupt the
+        # real request.
+        request = copy.deepcopy(request)
         request = self._before_record_request(request)
         return request and request in self and self.record_mode != RecordMode.ALL and self.rewound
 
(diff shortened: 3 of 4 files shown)


kevin1024:
Seems like a good fix to me.  Could you add a test?

abramclark:
I tried to write a test. Here's what I have so far ```
def test_before_record_request_corruption(tmpdir, httpbin):
    """Modifying request in before_record_request should not effect outgoing request"""
    def before_record(request):
        request.headers.clear()
        request.body.clear()
        return request

    req = Request(
        httpbin.url + "/post",
        data=urlencode({'test': 'exists'}).encode(),
        headers={'X-Test': 'exists'},
    )
    cass_file = str(tmpdir.join("modified_response.yaml"))
    with vcr.use_cassette(cass_file, before_record_request=before_record):
        resp = json.loads(urlopen(req).read())

    assert(resp['headers']['X-Test'] == 'exists')
    assert(resp['form']['test'] == 'exists')
```
It was pretty painful using urllib. Why not use requests? It's already a dependency in testing.

Unfortunately I can't figure out how to run the test. Using tox I get this error: ```
$> venv310/bin/pip install tox
$> venv310/bin/tox
...............
py310-httpx019: failed with ./runtests.sh (resolves to ./runtests.sh) is not allowed, use allowlist_externals to allow it
```

And when I try to use pytest directly I get this error: ```
$> venv310/bin/pip install pytest pytest-httpbin
$> venv310/bin/pytest tests/integration/test_filter.py
.................
  File ".../vcrpy/venv310/lib/python3.10/site-packages/httpbin/core.py", line 22, in <module>
    from werkzeug.wrappers import BaseResponse
ImportError: cannot import name 'BaseResponse'
```
So it looks like pytest-httpbin is importing the wrong version of werkzeug / using deprecated API. I've tried Linux and MacOS and Python 3.7, 3.9, and 3.10 and everything shows the same errors.

**Comment by abramclark** [comment]
Ahah I can skip all the pyenv stuff which I still haven't figured out, and simply use:

**Comment by abramclark** [comment]
This is looking good locally:
https://github.com/kevin1024/vcrpy/compare/master...abramclark:vcrpy:fix-174?expand=1

**Comment by abramclark** [comment]
Yep looks like the build is broken, same thing I was thinking. This isn't related to my diff, I assume

**Comment by abramclark** [comment]
This is ready to land right?

File: tests/integration/test_filter.py [hunk 1]
```diff
@@ -142,3 +142,23 @@ def test_decompress_regular(tmpdir, httpbin):
         resp = urlopen(url).read()
         assert_cassette_has_one_response(cass)
     assert_is_json(resp)
+
+
+def test_before_record_request_corruption(tmpdir, httpbin):
+    """Modifying request in before_record_request should not effect outgoing request"""
```
hartwork, line 148: I believe "effect" should be "affect" with "a" to be correct?

vcr/cassette.py
```diff
@@ -266,6 +266,9 @@ def _responses(self, request):
                 yield index, response
 
     def can_play_response_for(self, request):
+        # Deepcopy is here because mutation of `request` will corrupt the
+        # real request.
+        request = copy.deepcopy(request)
```
hartwork, line 271: There is an idea at https://github.com/kevin1024/vcrpy/issues/174#issuecomment-1242432327 to change the `request = copy.copy(request)` in…
https://github.com/kevin1024/vcrpy/blob/1a3bc67c7c9dcede68c1b072fff2676d221b230b/vcr/config.py#L221-L227
…to `deepcopy` instead, and it seems cleaner if also fixing what you're intending to fix. Have you tried? What do you think?

PS: Disclaimer: I only know part of the picture here, this will need review by someone else on top.

hartwork (commented):
Please see my two comments below:

hartwork:
@abramclark PS: It would be great to have this PR rebase into latest master to get the merge commit out of here. Thanks!

vcr/cassette.py
abramclark: Are you saying if I changed the `copy` to `deepcopy` in `before_record_request` I could remove the change in `can_play_response`?

abramclark: Well it passes my tests, so I will update with this. Seems a lot cleaner.

tests/integration/test_filter.py
abramclark: doh

vcr/cassette.py
hartwork: @abramclark the changes to `can_play_response_for` are still included. Did you forget to drop them? Does the test still pass without?

abramclark:
>  It would be great to have this PR rebased into latest `master`

I just rebased to master, do I also need to rebase to edit out the merge?

.github/workflows/main.yml
abramclark, line 1: This shouldn't be in the diff?

hartwork: @abramclark I think it's fair to say the rebase went wrong. `git reflog` should have reference to prior state to hard reset to, if you like to the rebase start over.

abramclark:
Abandoning due to git snafu, creating new PR

hartwork:
@abramclark please re-use the pull request, just force-push into it, same branche name

PR #689, opened 2023-05-03
Update urllib3 handlers

### LINKED ISSUE #688: vcrpy throws an error if latest urllib3 is installed
vcrpy won't import with latest urllib.

```
root@ef153dd6180d:/# python --version
Python 3.11.3
root@ef153dd6180d:/# pip install urllib3 vcrpy
Collecting urllib3
  Downloading urllib3-2.0.0-py3-none-any.whl (123 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 123.2/123.2 kB 10.0 MB/s eta 0:00:00
Collecting vcrpy
  Downloading vcrpy-4.2.1-py2.py3-none-any.whl (40 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 40.8/40.8 kB 45.4 MB/s eta 0:00:00
Collecting PyYAML
  Downloading PyYAML-6.0-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (757 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 757.9/757.9 kB 51.1 MB/s eta 0:00:00
Collecting wrapt
  Downloading wrapt-1.15.0-cp311-cp311-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl (78 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 78.9/78.9 kB 80.8 MB/s eta 0:00:00
Collecting six>=1.5
  Downloading six-1.16.0-py2.py3-none-any.whl (11 kB)
Collecting yarl
  Downloading yarl-1.9.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (282 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 282.8/282.8 kB 161.6 MB/s eta 0:00:00
Collecting multidict>=4.0
  Downloading multidict-6.0.4-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (117 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 117.4/117.4 kB 136.7 MB/s eta 0:00:00
Collecting idna>=2.0
  Downloading idna-3.4-py3-none-any.whl (61 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 61.5/61.5 kB 82.9 MB/s eta 0:00:00
Installing collected packages: wrapt, urllib3, six, PyYAML, multidict, idna, yarl, vcrpy
Successfully installed PyYAML-6.0 idna-3.4 multidict-6.0.4 six-1.16.0 urllib3-2.0.0 vcrpy-4.2.1 wrapt-1.15.0 yarl-1.9.2
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv

[notice] A new release of pip available: 22.3.1 -> 23.1.2
[notice] To update, run: pip install --upgrade pip
root@ef153dd6180d:/# python
Python 3.11.3 (main, Apr 12 2023, 14:31:14) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import vcr
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/vcr/__init__.py", line 2, in <module>
    from .config import VCR
  File "/usr/local/lib/python3.11/site-packages/vcr/config.py", line 11, in <module>
    from .cassette import Cassette
  File "/usr/local/lib/python3.11/site-packages/vcr/cassette.py", line 12, in <module>
    from .patch import CassettePatcherBuilder
  File "/usr/local/lib/python3.11/site-packages/vcr/patch.py", line 41, in <module>
    _VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'urllib3.connectionpool' has no attribute 'VerifiedHTTPSConnection'

```

vEpiphyte:
update urllib3 handlers for https://github.com/urllib3/urllib3/pull/1805 to close #688

This does have a problem of "what version of the urllib3 library is installed ?" since vcrpy does not have minimum version enforcement, so the PR as-is would not work with urllib3 < 1.25.9

### DIFF
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -18,39 +18,39 @@ try:
     from botocore.awsrequest import AWSHTTPConnection, AWSHTTPSConnection
 except ImportError:
     try:
-        import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+        import botocore.vendored.requests.packages.urllib3.connection as u_connection
     except ImportError:  # pragma: no cover
         pass
     else:
-        _Boto3VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
-        _cpoolBoto3HTTPConnection = cpool.HTTPConnection
-        _cpoolBoto3HTTPSConnection = cpool.HTTPSConnection
+        _Boto3VerifiedHTTPSConnection = u_connection.HTTPSConnection
+        _cpoolBoto3HTTPConnection = u_connection.HTTPConnection
+        _cpoolBoto3HTTPSConnection = u_connection.HTTPSConnection
 else:
     _Boto3VerifiedHTTPSConnection = AWSHTTPSConnection
     _cpoolBoto3HTTPConnection = AWSHTTPConnection
     _cpoolBoto3HTTPSConnection = AWSHTTPSConnection
 
-cpool = None
+u_connection = None
 # Try to save the original types for urllib3
 try:
-    import urllib3.connectionpool as cpool
+    import urllib3.connection as u_connection
 except ImportError:  # pragma: no cover
     pass
 else:
-    _VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
-    _cpoolHTTPConnection = cpool.HTTPConnection
-    _cpoolHTTPSConnection = cpool.HTTPSConnection
+    _VerifiedHTTPSConnection = u_connection.HTTPSConnection
+    _cpoolHTTPConnection = u_connection.HTTPConnection
+    _cpoolHTTPSConnection = u_connection.HTTPSConnection
 
 # Try to save the original types for requests
 try:
-    if not cpool:
-        import requests.packages.urllib3.connectionpool as cpool
+    if not u_connection:
+        import requests.packages.urllib3.connection as u_connection
 except ImportError:  # pragma: no cover
     pass
 else:
-    _VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
-    _cpoolHTTPConnection = cpool.HTTPConnection
-    _cpoolHTTPSConnection = cpool.HTTPSConnection
+    _VerifiedHTTPSConnection = u_connection.HTTPSConnection
+    _cpoolHTTPConnection = u_connection.HTTPConnection
+    _cpoolHTTPSConnection = u_connection.HTTPSConnection
 
 # Try to save the original types for httplib2
 try:


vEpiphyte:
I have not run the unit tests for this change.

hartwork (commented):
> so the PR as-is would not work with urllib3 < 1.25.9

@vEpiphyte given https://github.com/kevin1024/vcrpy/issues/688#issuecomment-1533416925 I'm more optimistic about that, but it I think we need a version that supports both urllib3 1.x and 2.x.

If I'm reading CI logs right, we'll need a dedicated pull request to fix tox in here before this to have a chance to be green?

hartwork:
I have a green CI now at #690 but it should be noted that some of the tests calling out to httpbin.org are flakey in the sense that the site sometimes returns status 502 and HTML output rather than JSON. I think midterm the tests should probably move away from talking to httpbin.org.

For now I think a good plan would be:
- merge #690 for a start
- rebase #689 here onto latest `master` after
- multiply everything in `tox.ini` that currently says "urllib3" times two:
  - "urllib3-1" with `urllib3<2`
  - "urllib3-2" with `urllib3>=2`

@vEpiphyte How much of that would you want to do yourself in here?
I could also try it out myself in a new pull request with your commit included via cherry-pick.

Are there other ideas?

alisakr:
@vEpiphyte You also need to update 2 other files, vcr/stubs/boto3_stubs.py and vcr/stubs/urllib3_stubs.py

alisakr:
@vEpiphyte @hartwork The solution, which works for urllib3 >=2 AND urllib3<2, is to make the same change here  to two other files, https://github.com/kevin1024/vcrpy/blob/master/vcr/stubs/boto3_stubs.py and https://github.com/kevin1024/vcrpy/blob/master/vcr/stubs/urllib3_stubs.py

hartwork:
@alisakr thanks for the feedback! I'll wait for a reply from @vEpiphyte a bit more so we can coordinate a good way forward.

alisakr:
@hartwork No problem! In the mean time, I see you landed your pull request to temporarily limit the urllib3 max version. https://github.com/kevin1024/vcrpy/pull/690 Should this be released now? 
Last release was august https://github.com/kevin1024/vcrpy/releases

hartwork:
@alisakr I don't have PyPI write permissions unfortunately and the quick-fix only helped CI, not users. Let's get both 1.x and 2.x supported and then ask someone with PyPI access to do a release right after. What do you think?

therve:
I had a look and outside of the imports (the fix is unfortunately much more complex than what's done in the PR), there is also at least one change in the response breaking vcrpy: https://github.com/urllib3/urllib3/pull/2649. If you want to support both urllib3 versions that might end up be messy,

hartwork:
@therve thanks for the pointer! I think if we had a pull request that was complete in terms of support for urllib3 >=2, that would greatly help evaluating with the cost of supporting both at the same time. I imagine with some luck we could just look at it it, and "see" the answer.

vEpiphyte:
Hi Everyone - I'm sorry for not giving more feedback on this. My immediate fire was put out by restricting requests (and avoiding a transitive urllib3 dependency change ), so I haven't had the time to get back to this.

I think we ( as vcrpy users / collaborators / maintainers / interested parties ) should make a decision regarded supported versions of the third party libraries like requests / boto3 ( see #693 ), since that would inform the correct way to fix the patching that is going on. I would like to say "let us just support the un-vendored urllib3 library which changed here urllib3/urllib3/pull/1805 " but I don't want to make that call in a vacuum simply because I was the one who put up a PR.

I believe that would make it much easier to maintain vcrpy moving forward and can be clearly documented. A developer who has to use a older version of one of those libraries which relies on a vendored or older version of urllib3 can always use a older version of vcrpy.

jairhenrique:
@vEpiphyte I agree with you. The ecosystem needs to evolve and maintaining deprecated libraries in new versions is a difficult job that is often not necessary. Software updates are healthy software!

Can you help us by removing support for old versions?

pquentin:
@jairhenrique Thanks! Do you agree that the minimum supported version should be those from #693?

One difficulty is that unlike other libraries we can’t tell pip about those supported versions, so what should we do if an older version of requests is actually in use? It seems to me that raising an exception like a RuntimeError telling users to downgrade vcrpy is the right thing to do.

hartwork:
> It seems to me that raising an exception like a RuntimeError telling users to downgrade vcrpy is the right thing to do.

Please take my vote for recommending an _upgrade_ to the other side or to least mention both options.

jairhenrique:
> @jairhenrique Thanks! Do you agree that the minimum supported version should be those from #693?
> 
> One difficulty is that unlike other libraries we can’t tell pip about those supported versions, so what should we do if an older version of requests is actually in use? It seems to me that raising an exception like a RuntimeError telling users to downgrade vcrpy is the right thing to do.

Yes.

Downgrade vcr or upgrade project libraries.

therve:
> @therve thanks for the pointer! I think if we had a pull request that was complete in terms of support for urllib3 >=2, that would greatly help evaluating with the cost of supporting both at the same time. I imagine with some luck we could just look at it it, and "see" the answer.

Unfortunately I couldn't get something working. I'm somewhat getting lost in the patching about which method should return which object.  [Code is here if you're curious](https://github.com/kevin1024/vcrpy/compare/master...therve:vcrpy:urllib3-2?expand=1).

vEpiphyte:
It sounds the consensus is "support the raw urllib3 library with the changes from 1.25.9" should be the way forward?

shifqu:
I have created a PR (#697) which [should provide backward compatibility as well as support urllib3>2](https://github.com/kevin1024/vcrpy/issues/688#issuecomment-1545837572).

vEpiphyte:
Closing this in favor of #698 which is a much cleaner approach than what I had shotgunned together. Good discussion here though :)

PR #690, opened 2023-05-04
Quick-fix the CI for recent tox and recent urllib3

hartwork:
Related to #688, meant as a new base to #689

CC  @vEpiphyte @deronnax


deronnax:
it fixes the tests but it does not fix the problem itself.

PR #692, opened 2023-05-06
Fix Docker hanging (fixes #519)

### LINKED ISSUE #519: Using vcr with docker API makes tests hang indefinitely
Here's my example test file:  

```python
import docker
import vcr


#@vcr.use_cassette()
def test_docker_images():
    env = docker.from_env()
    images = env.images.list()

    assert images
```

The test works fine, but if I'll uncomment the `@vcr.use_cassette()`, it will hang forever. 

Environment (Ubuntu, python 3.7):
```
attrs==19.3.0
certifi==2020.4.5.1
chardet==3.0.4
docker==4.2.0
idna==2.9
importlib-metadata==1.6.0
more-itertools==8.2.0
multidict==4.7.5
packaging==20.3
pluggy==0.13.1
py==1.8.1
pyparsing==2.4.7
pytest==5.4.1
PyYAML==5.3.1
requests==2.23.0
six==1.14.0
urllib3==1.25.9
vcrpy==4.0.2
wcwidth==0.1.9
websocket-client==0.57.0
wrapt==1.12.1
yarl==1.4.2
zipp==3.1.0
```

The examples from the documentation are working fine.

gravelBridge:
Fixes #519

### DIFF
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -234,7 +234,7 @@ class CassettePatcherBuilder:
             # used connections (which might actually be of a different
             # class) around. This while loop will terminate because
             # eventually the pool will run out of connections.
-            while not connection.__module__.startswith("docker") and snot isinstance(connection, connection_class):
+            while not connection.__module__.startswith("docker") and not isinstance(connection, connection_class):
                 connection = get_conn(pool, timeout)
             return connection
 
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -234,7 +234,7 @@ class CassettePatcherBuilder:
             # used connections (which might actually be of a different
             # class) around. This while loop will terminate because
             # eventually the pool will run out of connections.
-            while not isinstance(connection, connection_class):
+            while not connection.__module__.startswith("docker") and snot isinstance(connection, connection_class):
                 connection = get_conn(pool, timeout)
             return connection
 


hartwork (changes requested):
Hi @gravelBridge,

this seems to be about #519. A fix that makes vcr.py aware of Docker (or any other specific "user") is not ideal. I haven't looked into this myself more yet, but can you think of ways fixing this that are less specific to Docker? Thanks!

dschonholtz:
@hartwork Hey I'm one of the maintainers over at AutoGPT. This fixes our issue and we are using a Fork of vcrpy that has this change in it currently. https://github.com/Significant-Gravitas/vcrpy
We would really prefer to use the official package though.

Could you explain your concern above please?
I don't understand what you mean by: `lack of integration with docker.transport.unixconn.UnixHTTPConnection`
This works great for our use case.

hartwork:
Hi @dschonholtz,

what I mean is that this pull requests fixes the 100% CPU use hanging symptom, but the call to the docker daemon is not taking over —not recorded, not replayed — and as a result:
- the code does not create a cassette file
- talks to the host docker daemon (which may even be dangerous depending on test code and host system state)
- stop working when the docker daemon is not running.

That's pretty bad in my book.

If you want to see it yourself, try this for file `test_foo.py` and run `pytest test_foo.py`. It will create file `2.yml` but there will be no `1.yml`:
```python
import docker
import vcr
import urllib

@vcr.use_cassette('1.yml')
def test_docker_images():
    env = docker.from_env()
    images = env.images()
    assert images

@vcr.use_cassette('2.yml')
def test_urllib():
    response = urllib.request.urlopen('http://www.iana.org/domains/reserved').read()
    assert b'Example domains' in response
```

Now that that's confirmed not work, I will close the pull request as mistaken. If you see a misunderstanding here, please let me know and help me understand it, and we can always re-open. Thanks!

Pwuts:
@hartwork could you point me to why this PR would cause the problem you indicate? I'm not familiar with this part of the VCRpy library code, but we have been missing out on patches for the past 2 months and now also v5.0.0 so I'd like to see if we can fix it (properly).

hartwork:
@Pwuts please see my comment right above with details: https://github.com/kevin1024/vcrpy/pull/692#issuecomment-1553473284

Pwuts:
I read it before commenting, and am just not sure why there are issues with `docker-py` specifically. Is it because Docker uses a custom request class? But then if only standard request classes are used, why is there a class instance check in the loop? And how would monkey-patching `docker-py` fix this issue?

hartwork:
@Pwuts my impression is that docker-py uses a custom connection class [`UnixHTTPConnection`](https://github.com/docker/docker-py/blob/84414e343e526cf93f285284dd2c2c40f703e4a9/docker/transport/unixconn.py#L14) implementing HTTP over a Unix file socket and so VCR.py doesn't not support it before someone adds specific support for it hijacking that kind of transport, which this pull request didn't do.

PR #695, opened 2023-05-10
Drop support for Python 3.7 (after 2023-06-27)

jairhenrique:


tox.ini
```diff
@@ -67,7 +66,7 @@ commands =
     sphinx-build -W -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 # We use Python 3.7. Tox sometimes tries to autodetect it based on the name of
```
vEpiphyte, line 67: comment needs to be updated

hartwork:
End of life of Python 3.7 is Jun 27th. We should probably drop support before then, so I'll add that info to the title of the pull request.

tox.ini
```diff
@@ -89,20 +88,19 @@ deps =
     httplib2: httplib2
     urllib3: urllib3<2
     boto3: boto3
-    boto3: urllib3
+    boto3: urllib3<2
```
hartwork, line 91: This change seems unrelated. If it is, it should either be a separate pull request or just be dropped please. Thank you!

hartwork (commented):
@jairhenrique I understand that six should no longer be needed without Python 2 but I don't understand `.test_case` with regard to its original mission and the two related changes here. Could you help me (and potentially other reviewers) understand these changes? Thanks in advance!

jairhenrique:
> @jairhenrique I came across a test that looks like <3.6 only. Does this look like something that should be dropped in here as well?

I will drop!

alisakr:
@jairhenrique Would it be possible to delay this? I imagine a lot of companies will be continuing to use python 3.7 past end of life.

jairhenrique:
@alisakr Since the core language is no longer supported, I believe it is a waste of effort for libraries to maintain support for a deprecated version.

Python 3.8 has been live since October 2019. I believe enough time has passed to be able to move up a step in versions.

Keeping dependencies up to date is part of the job.

alisakr:
@jairhenrique agreed but sadly not practice (I'm trying to make it so). In any case, I think it's okay if tests don't run on python 3.7. in python 3.7 as long as there's nothing that explicitly prevents it from running on python 3.7. Is that the case?

jairhenrique:
@alisakr The version that is released with changes from this PR will not work with Python 3.7. You will be able to use an earlier version of the library.

All new fixes and improvements will prioritize live Python versions.

hartwork:
I fully support @jairhenrique in this, well said.

tests/unit/test_vcr.py
```diff
@@ -335,7 +335,7 @@ def function_with_return():
     assert ret == function_with_return()
 
 
-class TestVCRClass(VCR().test_case()):
+class TestVCRClass(VCR()):
```
hartwork, line 338: @jairhenrique this change seems broken to me (all else looks great). Could you elaborate on the idea behind this change? I would expect some form of inlining of the original auto-wrapping with `VCR.use_cassette` but this new version now derives from an instance of `VCR` rather than a class. Please help me understand the idea. Thanks!

hartwork (changes requested):

tests/unit/test_vcr.py
jairhenrique: @hartwork I just erased things without thinking too much. As the tests are passing I thought everything was fine. I'll go back to the implementation and look at it calmly.

hartwork: @jairhenrique thanks! Maybe we even need to keep `def test_case` around (and get the six bits out of it or not). This could be of interest: https://stackoverflow.com/a/18513858/11626624 . I haven't fully wrapped my head around it yet.

hartwork: @jairhenrique a few more thoughts if you don't mind:
- We could bring back `VCR.test_case` and six in for now and treat it as a separate issue.
- `VCR.test_case` could be argued to be part of public API. It's maybe the only bit in here that could be argued to break the API.
- Using six' `with_metaclass` may be nicer than inlining it (if keeping `VCR.test_case` intact is the goal).

See you

hartwork: @jairhenrique I was able to verify the regression in practice just mow:

#### Before on `master`

```console
# tox -e py310-requests-urllib3-1 -- -v tests/unit/test_vcr.py::TestVCRClass | grep ::test
tests/unit/test_vcr.py::TestVCRClass::test_one PASSED                    [ 33%]
tests/unit/test_vcr.py::TestVCRClass::test_two PASSED                    [ 66%]
tests/unit/test_vcr.py::TestVCRClass::test_dynamically_added PASSED      [100%]
```

#### After on `drop37`

```console 
# tox -e py310-requests-urllib3-1 -- -v tests/unit/test_vcr.py::TestVCRClass | grep ::test
<no output>
```

Probably the main question now is whether we want to bring the dependency on six back for [`six.with_metaclass`](https://github.com/benjaminp/six/blob/3b7efbcca41857da03fb01f004ccc425ab82dfbf/six.py#L856-L877) or include a copy of that function in VCR.py (e.g. named `_with_metaclass`). The license would allow it, but personally I'm in favor of using `six.with_metaclass` for this. What do you think?

jairhenrique: @hartwork I revert this changes.

hartwork (approved):

hartwork (approved):
@jairhenrique thanks again for your work and patience here and and the multiple rebases resolving conflicts! :+1: The end-of-life date is tomorrow, we're close enough, let's get your work merged!

hartwork:
@alisakr in case you missed it, we made release 4.4.0 today right before dropping Python 3.7 support for upcoming 5.0.0. That should allow you to pull in one last round of bug fixes into a Python 3.7 environment — that was part of the idea behind #724 — while also being clear that the only path forward with VCR.py or anything Python is with Python >=3.8 from now on.

PR #697, opened 2023-05-12
fix: use urllib3.connection where needed.

shifqu:
Since urllib3 v2 the re-export of connection.HTTPConnection in urllib3.connectionpool was removed.

In this commit we use urllib3.connection where needed. Some references to connectionpool.HTTPConnection are still there for backward compatibility.

Closes #688

PR #698, opened 2023-05-13
Drop support for botocore <1.11.0 and requests <2.16.2 (fixes #693)

### LINKED ISSUE #693: What are the minimum supported versions of boto3 and requests?
When trying to add support for urllib3 2.0, @hartwork ran into [various issues](https://github.com/hartwork/image-bootstrap/pull/347#issuecomment-1539769660), one of them being that "there is monkey patching of all sorts of bundled urllib3 copies happenning". Indeed, if we look at how urllib3 is imported, we can see three different things:

```
from urllib3.connectionpool import HTTPConnection, VerifiedHTTPSConnection
from requests.packages.urllib3.connectionpool import HTTPConnection, VerifiedHTTPSConnection
from botocore.vendored.requests.packages.urllib3.connectionpool import HTTPConnection, VerifiedHTTPSConnection (only in tests)
```

Do we still need this? requests 2.16.0 unvendored urllib3 back in 2017 and botocore 1.11.0 [unvendored requests](https://github.com/boto/botocore/pull/1495) in 2018.

If you look at requests usage today, you can see that:

 * requests 2.16+ represents 93% of downloads: https://pepy.tech/project/requests?versions=2.16.*&versions=2.17.*&versions=2.18.*&versions=2.19.*&versions=2.20.*&versions=2.21.*&versions=2.22.*&versions=2.23.*&versions=2.24.*&versions=2.25.*&versions=2.26.*&versions=2.27.*&versions=2.28.*&versions=2.29.*&versions=2.30.*
 * botocore 1.11+ represents 94% of downloads: https://pepy.tech/project/botocore?versions=1.11.*&versions=1.12.*&versions=1.13.*&versions=1.14.*&versions=1.15.*&versions=1.16.*&versions=1.17.*&versions=1.18.*&versions=1.19.*&versions=1.2*

hartwork:
Fixes #693

CC @jairhenrique @pquentin @vEpiphyte

### DIFF
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -44,14 +44,16 @@ else:
 
 # Try to save the original types for requests
 try:
-    if not cpool:
-        import requests.packages.urllib3.connectionpool as cpool
+    import requests
 except ImportError:  # pragma: no cover
     pass
 else:
-    _VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
-    _cpoolHTTPConnection = cpool.HTTPConnection
-    _cpoolHTTPSConnection = cpool.HTTPSConnection
+    if requests.__build__ < 0x021602:
+        raise RuntimeError(
+            "vcrpy >=4.2.2 and requests <2.16.2 are not compatible"
+            "; please upgrade requests (or downgrade vcrpy)"
+        )
+
 
 # Try to save the original types for httplib2
 try:
@@ -406,36 +408,6 @@ class ConnectionRemover:
 def reset_patchers():
     yield mock.patch.object(httplib, "HTTPConnection", _HTTPConnection)
     yield mock.patch.object(httplib, "HTTPSConnection", _HTTPSConnection)
-
-    try:
-        import requests
-
-        if requests.__build__ < 0x021603:
-            # Avoid double unmock if requests 2.16.3
-            # First, this is pointless, requests.packages.urllib3 *IS* urllib3 (see packages.py)
-            # Second, this is unmocking twice the same classes with different namespaces
-            # and is creating weird issues and bugs:
-            # > AssertionError: assert <class 'urllib3.connection.HTTPConnection'>
-            # >   is <class 'requests.packages.urllib3.connection.HTTPConnection'>
-            # This assert should work!!!
-            # Note that this also means that now, requests.packages is never imported
-            # if requests 2.16.3 or greater is used with VCRPy.
-            import requests.packages.urllib3.connectionpool as cpool
-        else:
-            raise ImportError("Skip requests not vendored anymore")
-    except ImportError:  # pragma: no cover
-        pass
-    else:
-        # unpatch requests v1.x
-        yield mock.patch.object(cpool, "VerifiedHTTPSConnection", _VerifiedHTTPSConnection)
-        yield mock.patch.object(cpool, "HTTPConnection", _cpoolHTTPConnection)
-        # unpatch requests v2.x
-        if hasattr(cpool.HTTPConnectionPool, "ConnectionCls"):
-            yield mock.patch.object(cpool.HTTPConnectionPool, "ConnectionCls", _cpoolHTTPConnection)
-            yield mock.patch.object(cpool.HTTPSConnectionPool, "ConnectionCls", _cpoolHTTPSConnection)
-
-        if hasattr(cpool, "HTTPSConnection"):
-            yield mock.patch.object(cpool, "HTTPSConnection", _cpoolHTTPSConnection)
 
     try:
         import urllib3.connectionpool as cpool
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -16,15 +16,16 @@ _HTTPSConnection = httplib.HTTPSConnection
 # Try to save the original types for boto3
 try:
     from botocore.awsrequest import AWSHTTPConnection, AWSHTTPSConnection
-except ImportError:
+except ImportError as e:
     try:
-        import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+        import botocore.vendored.requests  # noqa: F401
     except ImportError:  # pragma: no cover
         pass
     else:
-        _Boto3VerifiedHTTPSConnection = cpool.VerifiedHTTPSConnection
-        _cpoolBoto3HTTPConnection = cpool.HTTPConnection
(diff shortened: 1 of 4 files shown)


Jvillegasd:
approve this!

jairhenrique (approved):

vcr/patch.py
```diff
@@ -18,13 +18,14 @@
...
     try:
-        import botocore.vendored.requests.packages.urllib3.connectionpool as cpool
+        import botocore.vendored.requests  # noqa: F401
+
+        raise RuntimeError(
+            "vcrpy >=4.2.2 and botocore <1.11.0 are not compatible"
+            "; please upgrade botocore (or downgrade vcrpy)"
+        )
     except ImportError:  # pragma: no cover
         pass
```
pquentin, line 28: Can you also remove this try/except and the censored import? If the initial import failed raising the RuntimeError should be enough?

Additionally, should we use `raise  RuntimeError(...) from e`, e being the initial ImportError?

vcr/patch.py
```diff
@@ -43,14 +44,16 @@
 
 # Try to save the original types for requests
 try:
```
pquentin, line 46: Do you still need this try/except block?

pquentin (changes requested):
Thanks!

Have you looked in tests too? There's this line somewhere in the tests I think:

```
from botocore.vendored.requests.packages.urllib3.connectionpool import HTTPConnection, VerifiedHTTPSConnection
```

hartwork:
> Thanks!
> 
> Have you looked in tests too? There's this line somewhere in the tests I think:
> 
> ```
> from botocore.vendored.requests.packages.urllib3.connectionpool import HTTPConnection, VerifiedHTTPSConnection
> ```

@pquentin good point, done, pushed.

vcr/patch.py
hartwork: > Can you also remove this try/except and the censored import? If the initial import failed raising the RuntimeError should be enough?

@pquentin I'm not sure I get the idea. If we can cut anything away here, I'll need your help to see it. Could you elaborate?


> Additionally, should we use `raise RuntimeError(...) from e`, e being the initial ImportError?

To my understand that `from e` is happening automatically and would need suppressing via `from None` if we wanted the opposite. To see it for yourself in practice, you could run these two programs:

Nested traceback output:
```python
try:
    raise ValueError('first')
except:
    raise ValueError('second')
```

Proof of set `__context__`:
```python
try:
    try:
        raise ValueError('first')
    except:
        raise ValueError('second')
except Exception as e:
    print(repr(e.__context__))
```

What do you think?

vcr/patch.py
hartwork: @pquentin I believe so, at least vcrpy can be installed without any requests around:

```console
# cd "$(mktemp -d)"
# python3.8 -m venv venv
# source venv/bin/activate
# pip install vcrpy
[..]
# pip freeze | grep requests
[no output]
```

What do you think?

vcr/patch.py
hartwork: @pquentin PS: quick update, I found one small difference when adding `from e1` to the above sample — the message in between the two tracebracks changes slightly:

- a) Default: `During handling of the above exception, another exception occurred:`
- b) `from e1`: `The above exception was the direct cause of the following exception:`

For myself, the difference is of little value. Do you see value in the difference?

pquentin: My message was not clear, because my phone autocorrected from vendored to censored. Anyway, I understand how we need the try/except because requests is an optional dependency.

Do we need `import botocore.vendored.requests` though?

> @pquentin PS: quick update, I found one small difference when adding from e1 to the above sample — the message in between the two tracebracks changes slightly: [...]

Yes, I think this is valuable, because in case a) you don't know if the exception handling failed or if the new exception was intentional.

vcr/patch.py
pquentin: Good point, thanks.

vcr/patch.py
hartwork: > 



> My message was not clear, because my phone autocorrected from vendored to censored. Anyway, I understand how we need the try/except because requests is an optional dependency.

@pquentin cool!

> Do we need `import botocore.vendored.requests` though?

I believe we do, because how else would we know that we are dealing with `botocore <1.11.0` without that import? In theory we could (a) use other means of detection or (b) not (be able to) throw the runtime error or (c) keep as is. What do you think?

> Yes, I think this is valuable, because in case a) you don't know if the exception handling failed or if the new exception was intentional.

I have added it now just for you :smile:

pquentin: Ah yes because botocore is optional so you need to handle the case where it is not here at all! Makes sense.

Maybe contextlib.suppress would have been slightly clearer and would have avoided the nocover pragma.

Anyway, thanks, you can consider this comment resolved! Thanks for taking the time to explain things.

hartwork: @pquentin while I like `contextlib.suppress` in general, to work here the `raise RuntimeError` would need to go into the `with` body and and I think that would make it a bit mind bending. but with that in mind I now have refactored it to pull the `raise RuntimeError` into a new `else` branch to hopefully be more obvious and more consistent to the rest of it. I hope that's good with you. Thanks for the review!

vEpiphyte:
@hartwork This looks much easier to maintain going forward :) Thank you.

randomir:
Hi @hartwork / @jairhenrique, it's great to see the urllib3 bug fixed -- when do you plan to release the new version? Thanks!

pquentin:
@randomir this does not fix urllib3 2.0 compatibility yet - it's an intermediate step to make adding support easier!

randomir:
@pquentin, oh, I see. I got the wrong impression then from [this comment](https://github.com/kevin1024/vcrpy/pull/689#issue-1694928839), and haven't bothered double-checking. Anyway, do you have an approximate ETA for the urllib3 issue fix? It's quite annoying, but I'm hesitant to patch it with an upper-bound on urllib3... Thanks!

hartwork:
@randomir we'll need some version of #697 next. We'll hopefully know more in a couple of days.

CC @shifqu

vEpiphyte:
sorry for confusion - it seemed like shifqu's work was more complete and in the correct direction than what I had put together in a slapdash attempt.

randomir:
Thank you, @hartwork, I'll keep my :eyes: open. :smile:

PR #699, opened 2023-05-16
Support both urllib3 v1 and v2 at the same time (fixes #688)

hartwork:
Hi! :wave:

This builds on the work by @shifqu in #697 literally. Fixes #688, happy to adjust.

I would like to note two things:
- If anyone has ideas about the urllib3 v2 SSL connection abort protocol errors that seems to happen with Python 3.7/3.8/3.9 but not with Python 3.10/3.11 — that would rock! :pray: 
- If some form of this pull request ends up being merged, please do not squash the commits together. Thanks! :pray: 

Looking forward to your review! :beers: 

CC @jairhenrique @kevin1024 @pquentin @shifqu @vEpiphyte

PR #701, opened 2023-05-17
Allow triggering CI manually

hartwork:
I have needed this myself for #699 [before](https://github.com/kevin1024/vcrpy/actions/runs/4996820803) and #700 seems to be an indicator of the same thing. @jairhenrique does this look like the right move?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
   pull_request:
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - "*"
 
 jobs:
   build:


jairhenrique:
@hartwork this change duplicates the pipe on PR. This change is not necessary.

image

hartwork:
@jairhenrique yes it does. Is it a problem though? Other ideas? We both wanted to run CI in outside of a pull request which is currently not possible, so I believe some change would be great.

jairhenrique:
@hartwork we can enable manual dispatch. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

```yaml
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - "*"
  workflow_dispatch:
```

hartwork:
@jairhenrique that's an option, should I make the pull request do that?

hartwork:
@jairhenrique done, what do you think?

jairhenrique (approved):
LGTM! 🚀

hartwork:
@jairhenrique cool! I'll re-run the CI until green and merge on success, at least that's the plan :smiley:

PR #706, opened 2023-05-25
test_aiohttp.py: Make cookie tests use (local) pytest-httpbin to make them more robust

hartwork:
After https://github.com/kevin1024/vcrpy/actions/runs/5080550552 took 10(!) retries today to become fully green, let's get the flakiness out of `test_aiohttp.py` finally…


jairhenrique (approved):
🚀 🚀 🚀

Issue #512, opened 2020-02-03
Please include VCRMixin and VCRTestCase from vcrpy-unittest


bdrung:
https://github.com/agriffis/vcrpy-unittest adds a `VCRTestCase` (and a `VCRMixin`) class. They consists only of 23 lines of code. Please include those two functions in your project to avoid needing to install another package. Those tiny package have a bigger overhead and packaging them for Debian would be rejected due to this reason.

neozenith:
Hi, I just wanted to update active PRs and issues to let you know I will no longer be an active maintainer on this project.

I recently changed jobs where I no longer use VCRpy or Python at all so I don't have the bandwidth anymore. I won't be monitoring issues or PRs either.

I will have to delegate to other maintainers or ask for some contributors to step up.

I know the Azure CLI uses VCRpy so it would be nice if they could spare some sprint time to maintain this project as it was invaluable in our CI when I was using it.

Kind regards,
Josh

andrescollares:
Solved in #712

PR #709, opened 2023-05-30
Remove "vcrpy-unittest" section from docs

jedie:
https://github.com/agriffis/vcrpy-unittest is abandon

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -89,13 +89,6 @@ all
    the responses are not out of date) or can be used when you simply
    want to log all HTTP requests.
 
-Unittest Integration
---------------------
-
-While it's possible to use the context manager or decorator forms with unittest,
-there's also a ``VCRTestCase`` provided separately by `vcrpy-unittest
-<https://github.com/agriffis/vcrpy-unittest>`__.
-
 Pytest Integration
 ------------------
 


hartwork (commented):
I believe it's time to pull the code from vcrpy-unittest in here instead, see #512. Could be two stages of course, first dropping this, and then re-adding it as two pull requests. @jairhenrique @kevin1024 what do you think?

kevin1024:
Yeah makes sense to pull it in if it’s truly unmaintained and lots of folks are using it.  Paging @agriffis

agriffis:
Makese sense to pull it in here. Please go ahead. I'm no longer using or maintaining it.

agriffis:
I think maybe the reason it was a separate package was to avoid extra deps in this package, but I don't remember exactly. If someone wants to maintain the separate package, I could transfer the repo.

jairhenrique:
@kevin1024 does it make sense for you to create an org on github and move vcrpy and this project into that org?

hartwork:
> I think maybe the reason it was a separate package was to avoid extra deps in this package, but I don't remember exactly.

I see no additional runtime dependencies needed at https://github.com/agriffis/vcrpy-unittest/blob/master/setup.py#LL38C5-L38C21 .

My vote for inclusion so the test integration has a good chance to reach its users. As long as its separate, at least Linux distros will not have it, and its so close to core vcrpy and rather tiny, literally 44 lines in a single file https://github.com/agriffis/vcrpy-unittest/blob/master/vcr_unittest/testcase.py (if we ignore the testcases for the moment). Clear vote for inclusion from me.

kevin1024:
All right, let’s bring it in!

hartwork:
> All right, let’s bring it in!

Here you go: #712

hartwork:
> All right, let’s bring it in!

Closing in favor of pull request #712 …

PR #712, opened 2023-06-02
Integrate vcrpy-unittest (alternative to #709)

hartwork:
Alternative to #709


vcr/unittest/testcase.py
```diff
@@ -0,0 +1,39 @@
+import inspect
+import os
+import unittest
+
+from .. import VCR
+
+
+class VCRMixin(object):
```
jairhenrique, line 8:
```suggestion
class VCRMixin:
```

docs/usage.rst
```diff
@@ -92,9 +92,74 @@ all
...
+        def _get_vcr(self, **kwargs):
+            myvcr = super(MyTestCase, self)._get_vcr(**kwargs)
+            myvcr.register_matcher('mymatcher', mymatcher)
+            myvcr.match_on = ['mymatcher']
+            return myvcr
+
+See
+`the source
+<https://github.com/kevin1024/vcrpy/blob/master/vcr/unittest/testcase.py>`__
+for the default implementations of these methods, and `VCR.py`_ for more
```
agriffis, line 136: probably don't need to refer users to VCR.py here, now that the repos are combined

agriffis:
one small doc comment, otherwise looks fine to me at a glance. I haven't used this in a number of years, so I won't be testing it.

docs/usage.rst
hartwork: Good point, dropped, thanks! :+1:

vcr/unittest/testcase.py
hartwork: Good idea, dropped, thanks! :+1:

jairhenrique (approved):

hartwork:
@kevin1024 I believe this pull request is ready fro prime time. Any concerns left about merging?

kevin1024 (approved):
Lgtm!

PR #713, opened 2023-06-06
VCRHTTPResponse Not Working with Biopython 1.81

mghantous:
I am using Python 3.11, `biopython==1.81`, and `vcrpy==4.3.1`

Newer versions of biopython perform:

```python
handle = io.TextIOWrapper(vcr_response, encoding="utf-8")
```

rather than

```python
handle = io.TextIOWrapper(io.BufferedReader(vcr_response), encoding="utf-8")
```

When iterating over this object, there were no items (but calling this outside of VCR there were). It turned out all that was needed to fix this issue was to implement the `read1` method on `VCRHTTPResponse`.

### DIFF
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -93,7 +93,7 @@ def test_response_parses_correctly_and_fp_attribute_error_is_not_thrown():
         },
     }
     vcr_response = VCRHTTPResponse(recorded_response)
-    handle = io.TextIOWrapper(io.BufferedReader(vcr_response), encoding="utf-8")
+    handle = io.TextIOWrapper(vcr_response, encoding="utf-8")
     handle = iter(handle)
     articles = [line for line in handle]
     assert len(articles) > 1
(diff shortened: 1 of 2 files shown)


jairhenrique (approved):

hartwork:
@jairhenrique I am not sure how to re-produce this and have no prior experience with biopython. If you're sure about this, please go ahead.

mghantous:
Let me know if you need more details or references to biopython code. Both the test and PR description refer to a change they had made in this PR:
https://github.com/biopython/biopython/pull/2521

To replicate the issue you would delete the `read1` method I have added, but leave the test code. The assertion would fail due to an empty list of articles.

hartwork:
@mghantous I'm missing too much context and I'm busy elsewhere. I'll pass on this one. Thanks for your understanding.

hartwork (approved):
I had a chance for a second look at this today. I now understand this better, looks good, will merge in a second…

PR #715, opened 2023-06-13
Try to speedup ci

jairhenrique:
Based on https://mastodon.social/@hynek/110529830948230925

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -73,6 +73,8 @@ basepython = python3.7
 [testenv]
 # Need to use develop install so that paths
 # for aggregate code coverage combine
+package = wheel
+wheel_build_env = .pkg
 usedevelop=true
 commands =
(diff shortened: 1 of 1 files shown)


hartwork (commented):
Hi @jairhenrique,

I get the idea but I think this only has real effect when running the tests with multiple versions of Python on the same machine, while in CI every version of Python has their own container, e.g. the Python 3.7 job is only running `py37-*`tox environments in practice.

Looking at https://github.com/kevin1024/vcrpy/actions/workflows/main.yml I see faster runs in the recent past, which seems to confirm the not-much-effect-to-CI-if-at-all theory to me. This probably helps people running it locally with multiple versions of Python installed more, if any.

What do you think?

jairhenrique:
@hartwork makes sense, this is not so useful, I'll close this pr.

PR #674, opened 2022-12-16
Mark tests with @pytest.mark.online that need access to the Internet

jspricke:

PR #718, opened 2023-06-18
Make CI enforce that all online tests are marked with @pytest.mark.online

hartwork:
Follow-up to #674

This will make sure that distributors like Debian and Gentoo can run tests with `-m 'not online'` without worry that there are  tests accessing the Internet that are just missing `@pytest.mark.online` decoration…

CC @jspricke @jairhenrique

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,5 +31,13 @@ jobs:
       - name: Run tests with tox
         run: tox
 
+      - name: Run offline tests with tox with no access to the Internet
+        run: |
+          # We're using unshare to take Internet access
+          # away from tox so that we'll notice whenever some new test
+          # is missing @pytest.mark.online decoration in the future
+          unshare --map-root-user --net -- \
+              sh -c 'ip link set lo up; tox -- -m "not online"'
+
       - name: Run coverage
         run: codecov
--- a/runtests.sh
+++ b/runtests.sh
@@ -4,4 +4,4 @@
 # If you are getting an INVOCATION ERROR for this script then there is 
 # a good chance you are running on Windows.
 # You can and should use WSL for running tox on Windows when it calls bash scripts.
-REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` pytest $*
+REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` exec pytest "$@"


.github/workflows/main.yml
```diff
@@ -27,9 +27,23 @@ jobs:
...
 
+      - name: Run offline tests with tox without access to the Internet
+        run: |
+          # We're using bwrap/bubblewrap to take Internet access
+          # away from tox so that we'll notice whenever some new test
+          # is missing @pytest.mark.online decoration
+          bwrap \
+            --dev-bind / / \
+            --die-with-parent \
+            --unshare-net \
```
jspricke, line 44:
```suggestion
         unshare --map-root-user --net \
```

`unshare` is in util-linux, so should be installed by default.

runtests.sh
```diff
@@ -4,4 +4,4 @@
 # If you are getting an INVOCATION ERROR for this script then there is 
 # a good chance you are running on Windows.
 # You can and should use WSL for running tox on Windows when it calls bash scripts.
-REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` pytest $*
+REQUESTS_CA_BUNDLE=`python -m pytest_httpbin.certs` exec pytest "$@"
```
jspricke, line 7:
```suggestion
REQUESTS_CA_BUNDLE=`python3 -m pytest_httpbin.certs` exec pytest "$@"
```

hartwork: @jspricke good idea but we have more places calling python rather than python3 in the codebase. Let's fix all of them at once in a dedicated pull request after merge maybe?

.github/workflows/main.yml
hartwork: @jspricke interesting idea but for some reason even offline tests fail wehn called this way. E.g. `unshare --map-root-user --net -- tox -e py310-requests-urllib3-1 -- -m 'not online' tests/integration/test_basic.py` with the fixed `runtests.sh` will show five failing tests:

```
FAILED tests/integration/test_basic.py::test_nonexistent_directory - OSError: [Errno 101] Network is unreachable
FAILED tests/integration/test_basic.py::test_unpatch - OSError: [Errno 101] Network is unreachable
FAILED tests/integration/test_basic.py::test_basic_json_use - OSError: [Errno 101] Network is unreachable
FAILED tests/integration/test_basic.py::test_patched_content - OSError: [Errno 101] Network is unreachable
FAILED tests/integration/test_basic.py::test_patched_content_json - OSError: [Errno 101] Network is unreachable
```

So if we switch to `unshare`, we'll need to call it some other way. Any ideas?

jspricke: Can you try:

`unshare --map-root-user --net -- sh -c 'ip link set lo up; tox...`

runtests.sh
jspricke: fine with me.

.github/workflows/main.yml
hartwork: @jspricke cool! Worked well locally, let's see how the CI likes it…

PR #720, opened 2023-06-18
Use python3 (and pip3) command

hartwork:


jairhenrique:
`python` and `pip` are aliases for `python3` and `pip3`. What was the motivation for this change?

hartwork:
> `python` and `pip` are aliases for `python3` and `pip3`. What was the motivation for this change?

@jairhenrique not in all environments, e.g. not for users of Debian bullseye with [package `python-is-python2`](https://packages.debian.org/bullseye/python-is-python2) installed. In some sense this pull request is a bit late.

The motivator is comment https://github.com/kevin1024/vcrpy/pull/718#discussion_r1233336313, being more explicit about Python 3 and being waterproof about it.

What do you think?

PR #721, opened 2023-06-20
Make response.raw.stream() work for urllib3 v2 (fixes #714)

### LINKED ISSUE #714: [4.3.1] Missing support for Response.raw.stream() API from urllib3 v2
I've noticed this because [https://pypi.org/project/dns-lexicon/ dns-lexicon]'s test suite started failing after upgrading to urllib3-2.  A trivial reproducer is:

```python
import requests
import vcr


with vcr.use_cassette("test.yaml"):
    r = requests.get("https://gentoo.org", stream=True)
    print(list(r.raw.stream()))
```

Without vcrpy, this program works. With vcrpy, it throws:

```pytb
  File "/tmp/test.py", line 7, in <module>
    print(list(r.raw.stream()))
               ^^^^^^^^^^^^
AttributeError: 'VCRHTTPResponse' object has no attribute 'stream'
```

hartwork:
Fixes #714


vcr/stubs/__init__.py
```diff
@@ -170,6 +170,13 @@ def data(self):
     def drain_conn(self):
         pass
 
+    def stream(self):
+        while True:
```
kevin1024, line 174: Huh neat. Is `stream()` normally a generator?

kevin1024 (approved):
👍 from me!

vcr/stubs/__init__.py
hartwork: @kevin1024 yes it is: https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream

I'm realizing now that method signature `def stream(amt=65536, decode_content=None)` will take more then what's in here as soon as someone passes custom `amt` and/or `decode_content`. `amt` may just be passing along, `decode_content` would be less trivial.

I have asked now at https://github.com/kevin1024/vcrpy/issues/714#issuecomment-1599695480 about details/origin of the `.raw.stream` use. With some luck that gets us data if the quick and dirty version here has value or not. If it does, we could go with a simple version first, and extend the implementation in second round later.

hartwork: @kevin1024 quick update here: turned out one of our own tests already needed that signature to work (https://github.com/kevin1024/vcrpy/actions/runs/5327956493/jobs/9665521143#step:5:618) and so I made `amt` work now and while `decode_content` is technically ignored, it is at least accepted as a parameter and that makes all tests pass.

hartwork: @kevin1024 I think we're good for now, see my reply at https://github.com/kevin1024/vcrpy/issues/714#issuecomment-1601173659 .

kevin1024: Excellent!  Looks good to me

jairhenrique (approved):

PR #729, opened 2023-06-27
Enable rule B (flake8-bugbear) on ruff

jairhenrique:


hartwork (commented):
Hi @jairhenrique, I'm having some trouble with these:
```python
with pytest.raises(Exception): # noqa B017
```
Wouldn't it be better to replace `Exception` by a to-the-point exception here, i.e. fixing what B017 considers evil?

hartwork (approved):
@jairhenrique thanks for the adjustments! :+1:

PR #730, opened 2023-06-27
Avoid parent relative imports

jairhenrique:


hartwork (changes requested):
Hi @jairhenrique I would personally fix the exact opposite direction: this seems like the very case where use of relative imports is possible and arguably preferable. What drawback of relative imports is this trying to address that I may be missing?

jairhenrique:
@hartwork I favor only one-level relative import. For me it's cognitively easier to read the complete "path" than having to figure out where I am and where I should walk.

"Explicit is better than implicit."

hartwork:
@jairhenrique I understand that finding the right number of dots and path when adding imports may take a second more, but it's a one-time investment of a few seconds at write time. After it is found I would argue it doesn't matter to the reader how many levels above the source is as long as the import works properly at read time, in fact the import arguably gets simpler. The relative import signals that this is first party code in a way that otherwise only import sorting with an explicit list of first party packages can. (Also it could help import mix-ups in theory but I'm not sure those cases still work and don't feel like it's a strong argument.) The whole topic is maybe not important enough for me to want to raise the veto card here but at least I don't feel like I want to be the one to approve this change while I still consider it the wrong direction. We could let @kevin1024 decide?

PR #732, opened 2023-06-28
Configure read the docs V2

jairhenrique:
Reference: https://blog.readthedocs.com/migrate-configuration-v2/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,22 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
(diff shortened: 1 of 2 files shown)


hartwork (commented):
Hi @jairhenrique I have little to no experience with readthedocs and no account so my ability to do proper review here are limited. What I can say is that:
- `PYTHONPATH="$PWD" sphinx-build -a docs docs/html/` worked to build the docs locally with Sphinx 7.0.1. (There is one unrelated warning about it.)
- (We should probably have simple CI to catch build regressions on documentation.)
- We should probably use the pinning requirements feature to pin Sphinx to 7.0.1 for a stable build. (Dependabot and CI together will keep us in the present with it with little manual effort.)

So my vote for pinning Sphinx via a new file `docs/requirements.txt` at least. What do you think?

hartwork:
> @hartwork I can add you as a documentation maintainer in RTD. Can I use the email in your github profile?

@jairhenrique please don't, I'm really far from a maintainer here, especially with regard to documentation. It would send the wrong message.

jairhenrique:
@hartwork so, can I merge this pr?

hartwork:
> @hartwork so, can I merge this pr?

@jairhenrique I cannot take responsibility for the decision myself due to not seeing the full picture, but no objections from me for merging if you are confident about it. Your call.

PR #735, opened 2023-07-04
Fix typo in docs

quasimik:


hartwork (approved):
@quasimik good point, thanks for your contribution to VCR.py! :pray:

PR #736, opened 2023-07-07
[14 Oct 2024] Drop python 3.8 support

jairhenrique:


hartwork (commented):
@jairhenrique content looks good but I don't get the idea behind the "[06 Jun 2023]" prefix in the title. What's the idea?

jairhenrique:
@hartwork 06 Jun 2023 is EOL of Python 3.8

hartwork:
@jairhenrique that makes more sense now. Do we really want to have a pull request hang around for that long though? It may need constant rebasing for nothing, long-lived pull request are rarely healthy.

jairhenrique:
@hartwork This PR is a draft and I'd rather deal with small jobs than leave a big job for last. I'll keep this as a draft without too much "pressure".

hartwork:
@jairhenrique if it works for you I'm good, okay.

pyproject.toml
```diff
@@ -30,7 +27,8 @@ select = [
     "W",   # pycodestyle warning
 ]
 line-length = 110
-target-version = "py38"
+target-version = "py39"
+ignore = ['COM812', 'ISC001']
```
hartwork, line 31: Just to understand: why ignore the two and do that in this pull request? I ran ruff locally with the new target version and it had no complaints without the two. Should this be dropped? Would like to understand.

pyproject.toml
```diff
@@ -5,10 +5,7 @@ ignore-regex = "\\\\[fnrstv]"
 # ignore-words-list = ''
 
 [tool.pytest.ini_options]
-addopts = [
-    "--strict-config",
-    "--strict-markers",
-]
+addopts = ["--strict-config", "--strict-markers"]
```
hartwork, line 8: (Seems unrelated and status quo was arguably better for future changes with regard to future diffs. Okay to stay but recommending to drop.)

hartwork (changes requested):
@jairhenrique thanks for the adjustments, looks 99% ready to merge to me. Two details below:

pyproject.toml
jairhenrique: I reverted this change, but when I run `ruff format .`, I see this warning:

```
warning: The following rules may cause conflicts when used with the formatter: `COM812`, `ISC001`. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the `select` or `extend-select` configuration, or adding them to the `ignore` configuration.
```

pyproject.toml
jairhenrique: It's from my autoformat code editor. Revertes.

hartwork (approved):

pyproject.toml
hartwork: @jairhenrique thanks for the explanation! I have no strong opinion about it — please create a dedicated pull request if you are pro this addition e.g. if you see the warning repeatedly.

pyproject.toml
hartwork: Thank you!

PR #741, opened 2023-07-10
fixed issue with httpx streams not working

### LINKED ISSUE #597: async httpx streaming doesn't work
Steps to reproduce:
1. Use vcrpy with async httpx along with chunked http. E.g.:
```python
    with vcr.use_cassette("sample.yml"):
        async with client.stream(
                "POST",
                "/sample-path"
                },
            ) as response:
              async for chunk in response.aiter_lines():
                print('chunk', chunk)



```

Expected behaviour: it successfully records / replays content.

Actual behaviour: I get an exception: `vcr/stubs/httpx_stubs.py:72: IndexError`. httpx's `stream` method calls `send` with no args (only kwargs).

parkerhancock:
Fixes #597 by having VCR work when the httpx streaming interface is used.


tests/integration/test_httpx.py
```diff
@@ -40,6 +42,9 @@ def client(self):
 
     def __call__(self, *args, **kwargs):
         return self.client.request(*args, timeout=60, **kwargs)
+    
+    def stream(self, *args, **kwargs):
+        return self.client.stream(*args, timeout=60, **kwargs)
```
hartwork, line 47: When trying to reproduce the issue locally in a way similar to your test `test_simple_stream_fetching`, this method did not seem to be needed or called. Could you help me understand its purpose?

tests/integration/test_httpx.py
```diff
@@ -314,3 +319,19 @@ def test_redirect_wo_allow_redirects(do_request, mockbin, yml):
         assert response.status_code == 308
 
         assert cassette.play_count == 1
+
+@pytest.mark.online
+@pytest.mark.parametrize("url", ["https://github.com/kevin1024/vcrpy/issues/" + str(i) for i in range(3, 6)])
```
hartwork, line 324: It would be great if we could use the httpbin fixture so that we do not need real internet for this test. E.g. we could use `url = f'{httpbin.url}/bytes/<some number>'` and then check if what we read matches that length. The name of the fixture is just `httpbin`. `@pytest.mark.online` could then be dropped. What do you think?

tests/integration/test_httpx.py
```diff
@@ -314,3 +319,19 @@ def test_redirect_wo_allow_redirects(do_request, mockbin, yml):
...
+        with client.stream("GET", url):
+            pass
+
+    with vcr.use_cassette(yml) as cassette:
+        output = io.BytesIO()
+        with client.stream("GET", url) as response:
+            for chunk in response.iter_bytes():
+                output.write(chunk)
+            assert str(response.request.url) == url
+        assert cassette.play_count == 1
```
hartwork, line 337: These two asserts do not seem core to the test. If they are, let's briefly document the idea, if not, let's drop the two. If you like the idea of using `f'{httpbin.url}/bytes/<some number>'` (above) we could `assert len(output.getvalue()) == <some number>` here or compared against a pre-recorded body and add a self-test that the original matches that length, instead.

tests/integration/test_httpx.py
```diff
@@ -314,3 +319,19 @@ def test_redirect_wo_allow_redirects(do_request, mockbin, yml):
...
 
         assert cassette.play_count == 1
+
+@pytest.mark.online
+@pytest.mark.parametrize("url", ["https://github.com/kevin1024/vcrpy/issues/" + str(i) for i in range(3, 6)])
+def test_simple_stream_fetching(yml, url):
+    client = httpx.Client()
+    with vcr.use_cassette(yml):
+        with client.stream("GET", url):
+            pass
```
hartwork, line 329: Not reading the response here would be weird in end user code. How about this pattern:

```python
    with vcr.use_cassette(..........):  # recording
        ........
        expected_content = ..........

    with vcr.use_cassette(..........):  # playback
        actual_content = ................
        assert actual_content == expected_content
```

Would be similar to what this existing test does:

https://github.com/kevin1024/vcrpy/blob/c95b7264a217a5411bc025807d466936699234c1/tests/integration/test_httplib2.py#L34-L43

hartwork (commented):
Hi @parkerhancock thanks for the pull request! Please see my findings below:

tests/integration/test_httpx.py
parkerhancock: This was an attempt to make a test using the existing DoRequest classes. It was easier just to make a new client and use that as the object tested. It can be deleted.

tests/integration/test_httpx.py
parkerhancock: You are right, the byte count is really the core test. I'll update that.

tests/integration/test_httpx.py
parkerhancock: Great idea! I don't have a ton of httpbin experience, but that works. I'll submit an update.

hartwork:
@parkerhancock any news on the next iteration? I'd be happy to do review.

hartwork:
@parkerhancock any news on the next iteration? I'd be happy to do review.

hartwork:
@parkerhancock any news on the next iteration? I'd be happy to do review.

parkerhancock:
Update here - I'm really stuck on this. For some reason, vcrpy "works" with the streaming interface (i.e. it caches something and returns something), but it only caches the first chunk, not the whole request. Not sure where to go from here.

Kontrolix (changes requested):
Please consider adding the same thing to `async def _async_vcr_send` That would be awesome!

parkerhancock:
See #784, which fixes this and all other HTTPX issues.

PR #742, opened 2023-07-10
Move httpx monkeypatching to a lower layer to allow recording request generated by auth flows.

### LINKED ISSUE #684: httpx auth flow is skipped by replay?
Hello, firstly thanks for VCR - I use it tons of projects. I've come across an issue with VCR for custom authorization with HTTPX.

I'm following the custom auth instructions for HTTPX [here](https://www.python-httpx.org/advanced/#customizing-authentication) and testing my implementation with vcrpy.

I've tried to simplify as much as possible:

```python
import httpx

class MyAuth(httpx.Auth):
    requires_response_body = True

    def __init__(self, client_id: str, client_secret: str, auth_url: str):
        self.client_id = client_id
        self.client_secret = client_secret
        self.auth_url = auth_url
        self.access_token = "Bearer empty"

    def auth_flow(self, request):
        try:
            request.headers["Authorization"] = self.access_token
            response = yield request
            response.raise_for_status()
        except httpx.HTTPStatusError:
            auth_response = yield self.build_refresh_request()
            self.update_token(response=auth_response)
            request.headers["Authorization"] = self.access_token
            yield request

    def build_refresh_request(self):
        return httpx.Request(
            "POST",
            self.auth_url,
            data={
                "grant_type": "client_credentials",
                "client_id": self.client_id,
                "client_secret": self.client_secret,
            }
        )

    def update_token(self, response: httpx.Response):
        contents = response.json()
        self.access_token = f"Bearer {contents['access_token']}"
```

And test file:

```python
import os
from typing import Dict

import httpx
import pytest
import vcr

from my_auth import MyAuth


@pytest.fixture
def secrets() -> Dict[str, str]:
    return {
        "client_id": os.environ["CLIENT_ID"],
        "client_secret": os.environ["CLIENT_SECRET"],
        "auth_url": os.environ["AUTH_URL"],
        "api_url": os.environ["API_URL"],
        "api_key": os.environ["API_KEY"],
    }


def test_auth(secrets: Dict[str, str]):
    api_url = secrets["api_url"]
    api_key = secrets["api_key"]

    auth = MyAuth(
        client_id=secrets["client_id"],
        client_secret=secrets["client_secret"],
        auth_url=secrets["auth_url"],
    )

    with vcr.use_cassette("my_auth.yaml"):
        with httpx.Client(base_url=api_url) as client:
            r = client.get(
                "/endpoint",
                auth=auth,
                headers={"api-key": api_key, "Content-Type": "application/json"}
            )

    assert r.status_code == 200
    # On first run:
    # test_my_auth.py::test_auth PASSED
    # On subsequent runs:
    # FAILED test_my_auth.py::test_auth - assert 401 == 200
```

This test runs successfully the first time, as the auth flow is successful, but fails on subsequent test runs. Primarily, it looks like the requests from the auth flow are not replayed from the cassette. I think the mocking of `client.get` obscures the calling of the auth flow. This means the initial unauthorized response is returned, and there is no auth refresh request made.

This is with:
```
Python 3.10.8
httpx==0.23.3
pytest==7.2.2
```

If I could be pointed in the right direction I'd be happy to try to open a PR.

ionelmc:
Not sure if i wrote the test correctly but that test suite is way over my head.

I would like to note that the monkeypatching should be moved even lower but then there's some complicated redirect handling that I can't untangle.

Should fix #684.

### DIFF
--- a/tests/integration/test_httpx.py
+++ b/tests/integration/test_httpx.py
@@ -239,15 +239,13 @@ def test_behind_proxy(do_request):
 
 
 def test_auth_flow(do_request):
-    yml = (
-        os.path.dirname(os.path.realpath(__file__)) + "/cassettes/" + "test_httpx_test_test_auth_flow.yml"
-    )
+    yml = os.path.dirname(os.path.realpath(__file__)) + "/cassettes/" + "test_httpx_test_test_auth_flow.yml"
     url = "https://mockbin.org/headers"
     client = None
 
     class AuthWithExtraRequest(httpx.Auth):
         def auth_flow(self, request):
-            response = yield client.client.build_request("GET", url, params={'foo': 'bar'})
+            response = yield client.client.build_request("GET", url, params={"foo": "bar"})
             assert response
             request.headers["Foobar"] = "somethingsomething"
             yield request
@@ -262,7 +260,7 @@ def test_auth_flow(do_request):
         assert str(cassette_response.request.url) == url
         assert cassette.play_count == 2
 
-        headers = cassette_response.json()['headers']
+        headers = cassette_response.json()["headers"]
         for header in headers:
             if header["name"] == "foobar":
                 assert header["value"] == "somethingsomething"
--- a/tests/integration/test_httpx.py
+++ b/tests/integration/test_httpx.py
@@ -238,6 +238,39 @@ def test_behind_proxy(do_request):
         assert cassette_response.request.url == response.request.url
 
 
+def test_auth_flow(do_request):
+    yml = (
+        os.path.dirname(os.path.realpath(__file__)) + "/cassettes/" + "test_httpx_test_test_auth_flow.yml"
+    )
+    url = "https://mockbin.org/headers"
+    client = None
+
+    class AuthWithExtraRequest(httpx.Auth):
+        def auth_flow(self, request):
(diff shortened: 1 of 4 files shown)


ionelmc:
I think someone needs to push some buttons to get CI approved.

tests/integration/test_httpx.py
```diff
@@ -238,6 +238,37 @@ def test_behind_proxy(do_request):
...
+    class AuthWithExtraRequest(httpx.Auth):
+        def auth_flow(self, request):
+            response = yield client.client.build_request("GET", url, params={"foo": "bar"})
+            assert response
+            request.headers["Foobar"] = "somethingsomething"
+            yield request
+
+    with vcr.use_cassette(yml):
+        with do_request() as client:
+            client("GET", url, auth=AuthWithExtraRequest())
```
jairhenrique, line 255: The default value for `follow_redirects` is `False` on httpx client.
I think if you call the client with `client("GET", url, auth=AuthWithExtraRequest(), follow_redirects=True)` we don't need to overwrite a private method.

ionelmc: I don't really understand what you wrote. If you think something else works please try it.

parkerhancock:
Also fixed in #784

graingert:
Looks like #784 got merged

PR #743, opened 2023-07-15
Remove unnecessary dependency on six.

charettes:
Remove the last remaining usage of it in VCR.testcase.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,6 @@ class PyTest(TestCommand):
 install_requires = [
     "PyYAML",
     "wrapt",
-    "six>=1.5",
     "yarl",
     # Support for urllib3 >=2 needs Python >=3.10
     # so we need to block urllib3 >=2 for Python <3.10 for now.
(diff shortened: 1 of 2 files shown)


hartwork (changes requested):
@charettes thanks for the pull request :+1:, the output I see from running `pytest -v tests/unit/test_vcr.py::TestVCRClass` is unchanged, very nice. There is one tiny formatting issue that running `black .` would fix and that currently causes the red CI at https://github.com/kevin1024/vcrpy/actions/runs/5563344118/jobs/10164187675?pr=743#step:5:37 . Could you apply the tiny fix this needs for a green CI?

@jairhenrique do you have any concerns about the approach taken by this pull request?

charettes:
@hartwork sure thing, thanks for the review! Just forced-pushed a commit that addresses the formatting issue.

hartwork (approved):
> @hartwork sure thing, thanks for the review! Just forced-pushed a commit that addresses the formatting issue.

@charettes perfect, thank you! :+1:

PR #746, opened 2023-07-20
Add action to validate docs

jairhenrique:


.github/dependabot.yml
```diff
@@ -5,10 +5,6 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
-  - package-ecosystem: pip
-    directory: "docs/"
-    schedule:
-      interval: weekly
```
hartwork, line 11: I'm a bit in conflict with this because for the docs code to be healthy and maintained, we would treat it as a regular code part of the project and migrate it to the latest version of its dependencies including the fixes it needs. Making dependabot shut up about it forever goes against keeping the docs code healthy.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,20 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      
+      - run: pip install --exists-action=w --no-cache-dir -r docs/requirements.txt
+      - run: sphinx-build -T -E -b html -d _build/doctrees -D language=en docs/ html
+      - run: cat html/index.html
```
hartwork, line 20: - High-level naming for these steps a la "Installing build dependencies" and "Rendering HTML documentation" would be nice to have for people inspecting the CI job page.
- my vote for dropping `--exists-action=w` unless already needed. When we have dependencies this broken, we want to know and have red CI, I would want to argue.
- Need and intention behind `-E` is not clear to me. Drop or add comment?
- (`-b html` is already default, I see little value in being explicit here, your call.)
- I see no value in `-d _build/doctrees`, my vote for dropping.
- We already have `language = "en"` in `docs/conf.py`, my vote fro dropping `-D language=en`

docs/installation.rst
```diff
@@ -18,7 +18,7 @@ The following HTTP libraries are supported:
 -  ``boto3``
 -  ``http.client``
 -  ``httplib2``
--  ``requests`` (both 1.x and 2.x versions)
+-  ``requests`` (2.x versions)
```
hartwork, line 21: Could elaborate on the intentions and the situation here?

hartwork (changes requested):

.github/dependabot.yml
jairhenrique: With this config, dependabot duplicate PR's. Only the first config is needed.

image
image
image

docs/installation.rst
jairhenrique: I changed this for update the docs, because we don't support requests 1.x and force to run docs validation.

The trigger to run docs validation are changes on docs folder.

```yaml
name: Validate docs

on:
  push:
    paths:
      - 'docs/**'
```

.github/workflows/docs.yml
jairhenrique: @hartwork I wanted to execute the commands closer to what read the docs executes, so I included these parameters.
image

.github/dependabot.yml
hartwork: > With this config, dependabot duplicate PR's. Only the first config is needed.

@jairhenrique seems like you're right :+1:  #744 and #745 seem to be identical

.github/workflows/docs.yml
hartwork: @jairhenrique I'm not a fan of adding parameters than are not needed and/or not explaining themselves without documentation. I'll have a hard time supporting this approach, personally.

jairhenrique: Done!

docs/installation.rst
hartwork: @jairhenrique the trigger is clear but all I find regarding requests version support is the "vcrpy >=4.2.2 and requests <2.16.2 are not compatible" part. Either requests 1.x is supported or we should go further with blocking/detecting requests 1.x, e.g. in `setup.py`. Maybe a dedicated pull request would be the best fit?

.github/workflows/docs.yml
hartwork: @jairhenrique thanks! :pray:

docs/installation.rst
jairhenrique: @hartwork I don't know how validation would be done in setup.py, since requests is not a project dependency.
As this doc points to the latest version and as of version 4.2.2 we no longer support versions < 2.16.2 via runtime, I see no problem keeping this change here.

hartwork: > As this doc points to the latest version and as of version 4.2.2 we no longer support versions < 2.16.2 via runtime, I see no problem keeping this change here.

@jairhenrique you're right, I mis-read "vcrpy >=4.2.2 and requests <2.16.2" earlier, my bad.

> @hartwork I don't know how validation would be done in setup.py, since requests is not a project dependency.

It's in `tests_require` in `setup.py`, we can change plain `requests` to `requests >=2.16.2` there.
Would you be up for adding this change to this pull request?

jairhenrique: Done!

hartwork: @jairhenrique thank you! :+1:

hartwork (approved):

PR #747, opened 2023-07-20
Add editorconfig file

jairhenrique:


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[Makefile]
+indent_style = tab
+
+[*.{yml,yaml,json,js,ts}]
```
hartwork, line 13: - The two JSON files I find in Git have indention width 4 rather than two
- I find no `.js` files and no `.ts` files here

As a consequence, I would suggest to drop "js" and "ts" and "json" from this list. Thank you!

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[Makefile]
+indent_style = tab
```
hartwork, line 11: We don't have a makefile here, let's drop this section please.

hartwork (changes requested):

.editorconfig
jairhenrique: There is a file in the docs folder.

.editorconfig
jairhenrique: Done.

.editorconfig
hartwork: @jairhenrique I see :+1:

hartwork (approved):

Issue #707, opened 2023-05-29
Test failures with Python 3.12.0b1:AttributeError: type object 'VCRHTTPConnection[…]' has no attribute 'debuglevel'


mgorny:
When running the test suite against Python 3.12.0b1, I get a lot of failures from integration tests, they seem to roughly boil down to the same thing:

```pytb
_____________ ERROR at setup of test_matchers[uri-http://httpbin.org/get?p1=q1&p2=q2-http://httpbin.org/get?p2=q2&p1=q1] ______________

tmpdir = local('/tmp/pytest-of-mgorny/pytest-3/test_matchers_uri_http___httpb0')
httpbin = <pytest_httpbin.serve.Server object at 0x7f39ca56e5d0>
httpbin_secure = <pytest_httpbin.serve.SecureServer object at 0x7f39c9c6d520>

    @pytest.fixture
    def cassette(tmpdir, httpbin, httpbin_secure):
        """
        Helper fixture used to prepare the cassette
        returns path to the recorded cassette
        """
        default_uri = _replace_httpbin(DEFAULT_URI, httpbin, httpbin_secure)
    
        cassette_path = str(tmpdir.join("test.yml"))
        with vcr.use_cassette(cassette_path, record_mode=vcr.mode.ALL):
>           urlopen(default_uri)

tests/integration/test_matchers.py:24: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/lib/python3.12/urllib/request.py:212: in urlopen
    _opener = opener = build_opener()
/usr/lib/python3.12/urllib/request.py:593: in build_opener
    opener.add_handler(klass())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <urllib.request.HTTPHandler object at 0x7f39c9c6d7f0>, debuglevel = None

    def __init__(self, debuglevel=None):
[...]

PR #754, opened 2023-08-02
Enable testing on Python 3.12 & PyPy 3.9/3.10

mgorny:
As requested in https://github.com/kevin1024/vcrpy/issues/707#issuecomment-1662706535.

I've also noticed that PyPy3.9 and 3.10 were missing, so added them as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev", "pypy-3.8"]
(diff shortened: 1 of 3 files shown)


jairhenrique:
@mgorny can you apply the patch suggested on #707

mgorny:
> @mgorny can you apply the patch suggested on #707

I've done it in a slightly better way. However, I'm still not convinced it's the correct solution.

mgorny:
The problem with aiohttp is that the current release's sdist has C files generated using old version of Cython. I've added a tox change that disables the C extension when building for 3.12, making it possible to install pure Python version of aiohttp for testing.

mgorny:
I think the remaining test failures are (preexisting) flakiness and not related to my changes.

jairhenrique:
LGTM.
@hartwork can you help with this review?

hartwork:
@jairhenrique sorry, I lack bandwidth for this, I need to cut down my time on VCR.py for now.

Issue #123, opened 2014-11-21
When attempting to replay the same request twice using record_mode="new_episodes", vcr.py raises UnhandledHTTPRequestError


jashugan:
```
In [1]: import vcr, requests

In [2]: my_vcr = vcr.VCR(record_mode='new_episodes')

In [3]: with my_vcr.use_cassette('/tmp/my_cassette.yaml'):
   ...:     requests.get('http://localhost:8000')
   ...:     

In [4]: with my_vcr.use_cassette('/tmp/my_cassette.yaml'):
   ...:     requests.get('http://localhost:8000')
   ...:     requests.get('http://localhost:8000')
   ...:     
---------------------------------------------------------------------------
UnhandledHTTPRequestError                 Traceback (most recent call last)
<ipython-input-4-7a981d83d8af> in <module>()
      1 with my_vcr.use_cassette('/tmp/my_cassette.yaml'):
      2     requests.get('http://localhost:8000')
----> 3     requests.get('http://localhost:8000')
      4 

/Users/nithin/.virtualenvs/weblims/lib/python2.7/site-packages/requests/api.pyc in get(url, **kwargs)
     58 
     59     kwargs.setdefault('allow_redirects', True)
---> 60     return request('get', url, **kwargs)
     61 
     62 

/Users/nithin/.virtualenvs/weblims/lib/python2.7/site-packages/requests/api.pyc in request(method, url, **kwargs)
     47 
     48     session = sessions.Session()
---> 49     return session.request(method=method, url=url, **kwargs)
     50 
     51 

/Users/nithin/.virtualenvs/weblims/lib/python2.7/site-packages/requests/sessions.pyc in request(self, method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
    455         }
    456         send_kwargs.update(settings)
--> 457         resp = self.send(prep, **send_kwargs)
    458 
    459         return resp

/Users/nithin/.virtualenvs/weblims/lib/python2.7/site-packages/requests/sessions.pyc in send(self, request, **kwargs)
    567 
    568         # Send the request
--> 569         r = adapter.send(request, **kwargs)
    570 
    571         # Total elapsed time of the request (approximately)

/Users/nithin/.virtualenvs/weblims/lib/python2.7/site-packages/requests/adapters.pyc in send(self, request, stream, timeout, verify, cert, proxies)
    360                     decode_content=False,
[...]

PR #755, opened 2023-08-03
Fix record of new_episodes mode

### LINKED ISSUE #753: [CORE BUG][Verified Can Replicate] Wrong number of requests recorded under new_episodes mode
AFAIK, when `allow_playback_repeats` is `False`, the number of requests should be equal to the length of cassette data.
But when we use `new_episodes` mode with same/matched requests called several times, the number of cassette data is smaller than the number of requests.

This is the same problem as mentioned here: https://github.com/kevin1024/vcrpy/issues/245
I changed the reproduce script to `httpx` so it's easier to count the number of requests:
https://colab.research.google.com/gist/vickyliin/cf06ea0dc0f7f2f8a1b5c30bb8f3909d/vcrpy-bug.ipynb

In the `new_episodes` context, the number of reuqests is 7, while the length of `cassette.data` is only 4

The cause is `cassette.append` doesn't set `play_counts` just like #245 mentioned. When a new and matched request calls, it uses the previously appended cassette data. So the total number of `cassette.data` for the above script is 1 (old) + 6/2 (new episodes) = 4

For now I use the following code as a hotfix:

```python
class FixedCassette(Cassette):
    def append_and_play(self, request, response) -> None:
        prev_len_data = len(self.data)
        super().append(request, response)
        is_appended = len(self.data) == prev_len_data + 1
        if is_appended:
            self.play_counts[prev_len_data] += 1

    def _load(self) -> None:
        super()._load()
        self.append = self.append_and_play
        self._load = None


context = vcr.use_cassette(...)
context.cls = FixedCassette
wich context:
    run_requests()
```

Related code:

Records are only used when no `play_count`
https://github.com/kevin1024/vcrpy/blob/69de3886494d381c2c1a6620ea9a87dcd4df0d07/vcr/cassette.py#L266-L274

`append` does not set `play_count`
https://github.com/kevin1024/vcrpy/blob/69de3886494d381c2c1a6620ea9a87dcd4df0d07/vcr/cassette.py#L234-L247

`append` is also used when loading cassette so we can't set `play_count` in `append` directly
https://github.com/kevin1024/vcrpy/blob/69de3886494d381c2c1a6620ea9a87dcd4df0d07/vcr/cassette.py#L342-L348

For other modes, we can't see this problem because `rewound` is `False` (or it's `ALL` mode)
https://github.com/kevin1024/vcrpy/blob/69de3886494d381c2c1a6620ea9a87dcd4df0d07/vcr/cassette.py#L262-L264

vickyliin:
Fix #753

I found `play_count` is heavily relied on in the tests to check if a record is newly added or being played.

I'm wondering if
- I should modify the logic of play_count to be added whenever a record has been played no matter from loaded cassettes or from a real request and change all the related tests
- or I should find a work around

### DIFF
--- a/tests/integration/test_disksaver.py
+++ b/tests/integration/test_disksaver.py
@@ -20,7 +20,7 @@ def test_disk_saver_nowrite(tmpdir, mockbin_request_url):
     fname = str(tmpdir.join("synopsis.yaml"))
     with vcr.use_cassette(fname) as cass:
         urlopen(mockbin_request_url).read()
-        assert cass.play_count == 1
+        assert cass.play_count == 0
     last_mod = os.path.getmtime(fname)
 
     with vcr.use_cassette(fname) as cass:
@@ -41,7 +41,7 @@ def test_disk_saver_write(tmpdir, mockbin_request_url):
     fname = str(tmpdir.join("synopsis.yaml"))
     with vcr.use_cassette(fname) as cass:
         urlopen(mockbin_request_url).read()
-        assert cass.play_count == 1
+        assert cass.play_count == 0
     last_mod = os.path.getmtime(fname)
 
     # Make sure at least 1 second passes, otherwise sometimes
@@ -51,7 +51,7 @@ def test_disk_saver_write(tmpdir, mockbin_request_url):
     with vcr.use_cassette(fname, record_mode=vcr.mode.ANY) as cass:
         urlopen(mockbin_request_url).read()
         urlopen(mockbin_request_url + "/get").read()
-        assert cass.play_count == 2
+        assert cass.play_count == 1
         assert cass.dirty
     last_mod2 = os.path.getmtime(fname)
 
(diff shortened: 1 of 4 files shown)


kevin1024:
Thanks for the detailed investigation!

This behavior has an interesting history. The original vcrpy (pre-2014) actually raised an error on the second identical request, similar to Ruby VCR. But #123 requested the current behavior - someone wanted duplicate requests to be recorded rather than error. Commit 5162d183 introduced the current oscillating behavior leading to the play_count situation we have today.

So we've come full circle - some users want errors, some want recording, some want playback.

For your use case (same response reused for identical requests), `allow_playback_repeats=True` should give you the behavior you want:
https://vcrpy.readthedocs.io/en/latest/advanced.html#playback-repeats

If you feel strongly that the default behavior should change, feel free to open a new issue to discuss - it would be a breaking change but worth considering if there's enough interest.

Issue #710, opened 2023-05-30
ResourceWarning: unclosed <ssl.SSLSocket...


jedie:
I get some warnings about unclosed SSL socket, looks like:
```python
...
.../utilities/test_utils.py:110: in __exit__
	self.vcr_mock.__exit__(exc_type, exc_val, exc_tb)
site-packages/vcr/cassette.py:105: in __exit__
	next(self.__finish, None)
site-packages/vcr/cassette.py:65: in _patch_generator
	with contextlib.ExitStack() as exit_stack:
/usr/local/lib/python3.10/contextlib.py:561: in __exit__
	if cb(*exc_details):
site-packages/vcr/patch.py:402: in __exit__
	connection = pool.pool.get()
/usr/local/lib/python3.10/warnings.py:109: in _showwarnmsg
	sw(msg.message, msg.category, msg.filename, msg.lineno,
.../settings/warning_handler.py:47: in __call__
	frames = traceback.extract_stack()
WARNING from "test_sdk_client.py" in test_get_category_by_key():
"/usr/local/lib/python3.10/contextlib.py ResourceWarning: unclosed <ssl.SSLSocket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=...>"
```
Any advise to fix that?

hartwork:
Hi @jedie there are many details missing from your report. Python version seems to be 3.10, version of vcrpy is unclear, user level code doing the requests is unclear, could be requests, urllib3 v1 or v2 etc, is this behavior new or does it happen across all versions of vcrpy etc. Please add more details.

hartwork:
Thanks! Is this behavior new or does it happen across all versions of vcrpy?
File `test_sdk_client.py` was making requests using `requests` directly or with some layer in between? If so which one?

I think vcrpy does not open SSL connections by itself so I would assume it's either a bug in vcrpy's connection pool code or the library outside of vcrpy that does open SSL connections. I'm guessing here, any help with debugging is welcome.

hartwork:
Three weeks without a reproducer, closing as "cannot reproduce". Happy to re-open if someone can provide a [minimal producible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example). Thanks!

jedie:
Sorry, i have currently no time to investigate this further... I currently just ignore the warning ;)

admirito:
Hi @hartwork,

[...]

PR #758, opened 2023-08-07
Fix ResourceWarning unclosed socket

admirito:
This PR fixes issue #710 by properly closing the underlying socket. It first uses `pool._put_conn` to keep the connection in the pool, and later removes and closes it when the context manager exits.

I was unsure about the exact purpose of the `ConnectonRemove` class, so I made minimal changes to minimize the risk of breaking the code and there may be better solutions for fixing this issue.

For example, the `urllib3.connectionpool.HTTPConnectionPool` will utilize a weakref to terminate pool connections. By appending our connection to it, it will also take care of closing our connection. So another solution could be to modify the `__exit__` in `patch.ConnectionRemover` method and add our connection to the pool:

```py
class ConnectionRemover:
    ...

    def __exit__(self, *args):
        for pool, connections in self._connection_pool_to_connections.items():
            for connection in connections:
                if isinstance(connection, self._connection_class):
                    pool._put_conn(connection)
```

### DIFF
--- a/vcr/patch.py
+++ b/vcr/patch.py
@@ -367,6 +367,7 @@ class ConnectionRemover:
     def add_connection_to_pool_entry(self, pool, connection):
         if isinstance(connection, self._connection_class):
             self._connection_pool_to_connections.setdefault(pool, set()).add(connection)
+            pool._put_conn(connection)
 
     def remove_connection_to_pool_entry(self, pool, connection):
         if isinstance(connection, self._connection_class):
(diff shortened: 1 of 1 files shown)


graingert:
It looks like the tests are failing here, but the logs have expired @admirito can you rebase this on the default branch?

graingert:
Can you also test that this is fixed by enabling filterwarnings=error in the pytest config

graingert:
I've added some more commits to get the tests passing with filterwarnings=error https://github.com/kevin1024/vcrpy/pull/801

graingert:
this seems to cause a problem in `tests/integration/test_urllib3.py::test_post[https]`

Issue #394, opened 2018-09-22
Tell between recorded and real response without logging


deeplook:
It seems like the only way to tell if a request response actually comes from a cassette or is a real, _fresh_ response of some API is to use the logging mechanism described for debugging here: https://vcrpy.readthedocs.io/en/latest/debugging.html. It would be helpful to have a simpler way without involving the logging machinery, maybe by writing a flag into `vcr.VCR` objects.

deeplook:
Well, something like this might work. Is this the recommended way?

```python
import vcr
from vcr.cassette import Cassette
from vcr.request import Request

path = 'headers.yml'

# execute one request, record response
with vcr.use_cassette(path):
    resp = requests.get('http://httpbin.org/headers')

# test if some request has been recorded 
req = Request('GET', 'http://httpbin.org/headers', 'irrelevant body?', {'foo': 'not checked'})
assert Cassette(None).load(path=path).can_play_response_for(req) == True
assert req in Cassette(None).load(path=path)
```

vEpiphyte:
aiohttp stub does not log playback of of recorded responses

neozenith:
A lot of changes have happened to VCRpy since this ticket was opened. As this ticket has become stale and we have a lot of old tickets that need to be groomed, I'm closing this for now to make it easier to see what is still relevant.

However if it is still needed, please feel free to re-open or create a new ticket.

Thanks! 🙏

PR #759, opened 2023-08-12
Expose Request object so we can see if a replay occurred.

lewisdawson:
In some cases, we need to determine if the actual request was invoked or the cached version was returned. In order to do that, we need to do an evolution of #394. We need to reconstruct the underlying `Request` object:

```python
def make_vcr_request(req):
      body = req.read().decode("utf-8")
      uri = str(req.url)
      headers = dict(req.headers)
      return Request(req.method, uri, body, headers)


self._current_request_used_cache = self.cassete.can_play_response_for(make_vcr_request(request))
```

To do that, we need VCR's `Request` object exposed in the package. 

**This PR simply exposes that `Request` object for use. Open to thoughts on how to better accomplish this by folks that know this tool better.**

### DIFF
--- a/vcr/__init__.py
+++ b/vcr/__init__.py
@@ -3,6 +3,7 @@ from logging import NullHandler
 
 from .config import VCR
 from .record_mode import RecordMode as mode  # noqa: F401
+from .request import Request
 
 __version__ = "5.1.0"
 


lewisdawson:
@kevin1024 Any interest in merging this PR? It's a very simple change.

Issue #208, opened 2015-09-13
'rerecord' mode that essentially ignores the recorded cassette


IvanMalison:
Is there a reason that this mode doesn't exist? Perhaps because you can simply delete the existing cassette and rerecord?

kevin1024:
IIRC I pretty much just ripped off the modes from Ruby VCR.  That said, no reason we can't add more if there's demand.

agriffis:
Doesn't `all` already do this? https://vcrpy.readthedocs.org/en/latest/usage.html#all

IvanMalison:
@agriffis not exactly. I think `all` will append to the existing cassette instead of completely rewriting it. I might be wrong about that.

agriffis:
Nah, `all` is "Never replay previously recorded interactions. This can be temporarily used to force VCR to re-record a cassette"

The mode you're thinking of is `new_episodes`

Pretty sure that `all` is the same as `rerecord`

IvanMalison:
> Nah, all is "Never replay previously recorded interactions. This can be temporarily used to force VCR to re-record a cassette"
> 
> The mode you're thinking of is new_episodes
> 
> Pretty sure that all is the same as rerecord

From what I can tell from reading the code, the only place where all has a *different* effect than new_episodes is

https://github.com/kevin1024/vcrpy/blob/8a5bf23d3448892ea10a29b5392927d0a8784271/vcr/cassette.py#L273

Given that new episodes DOES preserve existing records, I'm not sure how it would be possible for all mode NOT to preserve existing records.

In fact, there does not seem to be anything special about the string 'new_episodes' at all you could use any other string that isn't once, none or all and acheive the same effect.

agriffis:
Ugh, you're right, `all` doesn't behave like I thought. It appends to the cassette, even though the doc implies that it should be overwriting.

And that sucks about using any string, because it means a misspelling will be interpreted as `new_episodes`. It should be a strict validation and raise `ValueError` if not recognized.

IvanMalison:
> Ugh, you're right, all doesn't behave like I thought. It appends to the cassette, even though the doc implies that it should be overwriting.

As such, do you think we should consider this a bug? Is the current functionality of 'all' something someone could conceivably want? @kevin1024 I'd like to hear your thoughts too.

> And that sucks about using any string, because it means a misspelling will be interpreted as new_episodes. It should be a strict validation and raise ValueError if not recognized.

Yeah this is definitely something that I'd like to change as well.

[...]

PR #763, opened 2023-08-20
Add an option to remove unused requests from cassette

danielnsilva:
This PR adds an option to remove previously recorded interactions if they are not used. I believe this feature can address the issue discussed in #208. We don't need to focus on ignoring all requests in the cassette file, but instead, we can write to file only replayed interactions and new ones. If no previous request is used, VCR will save only new interactions.

There is a [similar feature](https://github.com/vcr/vcr/pull/946) in Ruby VCR.

Replaces #677

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -438,4 +438,4 @@ option to ``True``, VCR will not save old HTTP interactions if they are not used
 
     my_vcr = VCR(drop_unused_requests=True)
     with my_vcr.use_cassette('fixtures/vcr_cassettes/synopsis.yaml'):
-        # your http here
+        ...  # your HTTP interactions here
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -426,3 +426,16 @@ If you want to save the cassette only when the test succeedes, set the Cassette
 
     # Since there was an exception, the cassette file hasn't been created.
     assert not os.path.exists('fixtures/vcr_cassettes/synopsis.yaml')
+
+Drop unused requests
+--------------------
+
+Even if any HTTP request is changed or removed from tests, previously recorded
+interactions remain in the cassette file.  If set the ``drop_unused_requests``
+option to ``True``, VCR will not save old HTTP interactions if they are not used.
+
+.. code:: python
+
+    my_vcr = VCR(drop_unused_requests=True)
+    with my_vcr.use_cassette('fixtures/vcr_cassettes/synopsis.yaml'):
+        # your http here
--- a/vcr/cassette.py
+++ b/vcr/cassette.py
@@ -348,8 +348,10 @@ class Cassette:
         """List of new HTTP interactions (request/response tuples)"""
         new_interactions = []
         for request, response in self.data:
-            if all(not requests_match(request, old_request, self._match_on)
-                   for old_request, _ in self._old_interactions):
+            if all(
+                not requests_match(request, old_request, self._match_on)
+                for old_request, _ in self._old_interactions
+            ):
                 new_interactions.append((request, response))
         return new_interactions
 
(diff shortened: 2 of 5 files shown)


jairhenrique:
LGTM 🚀 

@danielnsilva can you fix some lint erros?

jairhenrique (approved):

jairhenrique:
@hartwork can you help review this pull request?

vcr/cassette.py
```diff
@@ -337,10 +344,29 @@ def find_requests_with_most_matches(self, request):
...
+        responses = self.responses
+        if self.drop_unused_requests:
+            interactions = self._played_interactions + self._new_interactions()
+            requests = [request for request, _ in interactions]
+            responses = [response for _, response in interactions]
+        return {"requests": requests, "responses": responses}
 
     def _save(self, force=False):
+        if len(self._played_interactions) < len(self._old_interactions):
+            force = True
```
hartwork, line 369: Is this addition missing a check for `drop_unused_requests`?

vcr/cassette.py
```diff
@@ -337,10 +344,29 @@ def find_requests_with_most_matches(self, request):
...
+        return new_interactions
+
     def _as_dict(self):
-        return {"requests": self.requests, "responses": self.responses}
+        requests = self.requests
+        responses = self.responses
+        if self.drop_unused_requests:
+            interactions = self._played_interactions + self._new_interactions()
+            requests = [request for request, _ in interactions]
+            responses = [response for _, response in interactions]
```
hartwork, line 364: This behavior seems unexpected in a method `_as_dict` from a user point of view. Can this behavior be moved elsewhere?

docs/advanced.rst
```diff
@@ -426,3 +426,16 @@ If you want to save the cassette only when the test succeedes, set the Cassette
...
+
+Even if any HTTP request is changed or removed from tests, previously recorded
+interactions remain in the cassette file.  If set the ``drop_unused_requests``
+option to ``True``, VCR will not save old HTTP interactions if they are not used.
+
+.. code:: python
+
+    my_vcr = VCR(drop_unused_requests=True)
+    with my_vcr.use_cassette('fixtures/vcr_cassettes/synopsis.yaml'):
+        # your http here
```
hartwork, line 441:
```suggestion
        ...  # your HTTP interactions here
```
?

(to make it valid Python)

hartwork (changes requested):

vcr/cassette.py
danielnsilva: You're right, it does need a check for drop_unused_requests.

vcr/cassette.py
danielnsilva: A better approach might be to move it to the `_save` method.

```python
def _as_dict(self):
    return {"requests": self.requests, "responses": self.responses}

def _build_used_interactions_dict(self):
    interactions = self._played_interactions + self._new_interactions()
    cassete_dict = {
        "requests": [request for request, _ in interactions],
        "responses": [response for _, response in interactions],
    }
    return cassete_dict

def _save(self, force=False):
    if self.drop_unused_requests and len(self._played_interactions) < len(self._old_interactions):
        cassete_dict = self._build_used_interactions_dict()
        force = True
    else:
        cassete_dict = self._as_dict()
    if force or self.dirty:
        self._persister.save_cassette(self._path, cassete_dict, serializer=self._serializer)
        self.dirty = False
```

hartwork: @danielnsilva looks like a good idea to me! :+1:

hartwork (approved):
@danielnsilva thanks for the adjustments! :+1:

PR #764, opened 2023-08-24
fix: Handle Binary Responses in httpx

Kontrolix:
This pull request addresses a critical issue encountered when handling binary responses using httpx. The existing code assumed that response content could always be decoded as UTF-8 text. However, in cases where binary data was present, decoding errors occurred, and these errors were hidden using the "ignore" decoding mode. This led to corrupted data being recorded and replayed. To fix this issue, this pull request introduces changes to properly handle binary content and utilize base64 encoding when necessary.

**Changes Made:**

1. Modified `_to_serialized_response` function:
   - Previously, the code attempted to decode response content as UTF-8, causing issues with binary data. The function now checks for UnicodeDecodeError and, if encountered, encodes the content using base64 and sets a flag to indicate base64 encoding.
   - Added the `base64encoded` flag to the serialized response dictionary.

2. Modified `_from_serialized_response` function:
   - If the `base64encoded` flag is set in the serialized response, the function decodes the content using base64. Otherwise, it treats the content as UTF-8 encoded text.
   - This change ensures that both text and binary responses are handled correctly.

**Benefits:**

- Resolves a critical issue where binary response content caused silent decoding errors and led to corrupted data being recorded and replayed.
- Ensures proper handling of binary data by utilizing base64 encoding when needed.
- Prevents errors from being silently ignored, thus maintaining the integrity of recorded interactions.

### DIFF
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -1,3 +1,4 @@
+import base64
 import functools
 import inspect
 import logging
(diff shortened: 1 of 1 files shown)


parkerhancock:
Also fixed in #784

parkerhancock:

PR #767, opened 2023-09-12
Migrate to declarative Python package config

deronnax:
Declarative is always preferable to imperative, and setuptools is no exception, [recommending it themselves](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#:~:text=This%20approach%20allows%20automation%20scenarios%20and%20can%20reduce%20boilerplate%20code.). Classical `setup.py` are not hard-deprecated, but they are considered a bad-practice if there is not a strong reason necessiting it (like uncommon packages/path complicated to find on the build host machine).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,72 @@
+[project]
+name = "vcrpy"
+authors = [{name = "Kevin McCarthy", email = "me@kevinmccarthy.org"}]
+license = {text = "MIT"}
+description = "Automatically mock your HTTP interactions to simplify and speed up testing"
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Topic :: Software Development :: Testing",
+    "Topic :: Internet :: WWW/HTTP",
+    "License :: OSI Approved :: MIT License",
+]
+urls = {Homepage = "https://github.com/kevin1024/vcrpy"}
+requires-python = ">=3.10"
+dependencies = [
+    "PyYAML",
+    "wrapt",
+]
+dynamic = ["version"]
+
+[project.readme]
+file = "README.rst"
+content-type = "text/x-rst"
+
+[project.optional-dependencies]
+tests = [
+    "aiohttp",
+    "boto3",
+    "cryptography",
+    "httpbin",
+    "httpcore",
+    "httplib2",
+    "httpx",
+    "pycurl; platform_python_implementation !='PyPy'",
+    "pytest",
+    "pytest-aiohttp",
+    "pytest-asyncio",
+    "pytest-cov",
+    "pytest-httpbin",
+    "requests>=2.22.0",
+    "tornado",
+    "urllib3",
+    "werkzeug==2.0.3",
+]
+
+[tool.setuptools]
+include-package-data = false
+
+[tool.setuptools.packages.find]
+exclude = ["tests*"]
+namespaces = false
+
+[tool.setuptools.dynamic]
+version = {attr = "vcr.__version__"}
+
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
 [tool.codespell]
 skip = '.git,*.pdf,*.svg,.tox'
 ignore-regex = "\\\\[fnrstv]"


setup.cfg
```diff
@@ -1,2 +1,52 @@
...
+    Topic :: Internet :: WWW/HTTP
+    License :: OSI Approved :: MIT License
+
+[options]
+packages = find:
+install_requires =
+    PyYAML
+    wrapt
+    yarl
+    urllib3 <2; python_version <'3.10'
```
hartwork, line 37: We're losing an important comment here for why we have this arguably unexpected dependency in place. Please bring it back.

setup.cfg
```diff
@@ -1,2 +1,52 @@
...
+    aiohttp
+    boto3
+    httplib2
+    httpx
+    pytest
+    pytest-aiohttp
+    pytest-httpbin
+    requests>=2.16.2
+    tornado
+    Werkzeug==2.0.3
```
hartwork, line 49: Also here: We're losing an important comment here for why we have this arguably unexpected dependency in place. Please bring it back.

hartwork (changes requested):
@deronnax before we go into details, I'd like to ask for the motivation behind this change when we could (and will eventually) migrate to `build` and `pyproject.toml`. Please help me see the value in this intermediary stage. Thank you!

deronnax:
@hartwork pyproject.toml is a bit of a bigger leap. initially I offered direct move to pyproject.toml but people were reluctant to do the leap. But if you are willing, I will do the conversion straight to pyproject.toml.

jairhenrique:
Following [PEP 621](https://peps.python.org/pep-0621/), projects metadata should be put on `pyproject.toml` file.

Using the pypa tool [hatch](https://hatch.pypa.io/latest/), following this steps, hatch automatically create or update the `pyproject.toml` file.

```bash
pip install hatch
hatch new --init
```

setup.cfg
deronnax: Indeed. sorry, restored.

setup.cfg
deronnax: sorry. Also restored.

deronnax:
@hartwork sorry I meant "people in general on GitHub", not the vcrpy community. Some people don't feel to do the jump to pyproject.toml yet. 
I did the pyproject.toml conversion on today's master, and restored the gone comments.

hartwork:
> @hartwork sorry I meant "people in general on GitHub", not the vcrpy community. Some people don't feel to do the jump to pyproject.toml yet.

@deronnax I understand, thanks for the clarification :+1: 

> I did the pyproject.toml conversion on today's master, and restored the gone comments.

Thanks! The CI will need adjustment to use `build` and all, but before we know if @kevin1024 is on board with this direction there is a risk for wasting your time, just to avoid potential disappointment.

deronnax:
@kevin1024 what's your stance on the subject? I can also do the move to the pyproject.toml if you prefer.

kevin1024:
Hello! I’m afraid I haven’t been following new developments in Python packaging very closely for awhile now.  My general opinion on this is:  if it increases compatibility with the newer packaging ecosystem without breaking compatibility with any of our existing environments we support, then I am in favor.

There are a number of Linux distributions that package VCR.py from source.  I believe if we make this change it may also affect their build scripts, so ideally we would also coordinate with those maintainers.

hartwork:
> Hello! I’m afraid I haven’t been following new developments in Python packaging very closely for awhile now. My general opinion on this is: if it increases compatibility with the newer packaging ecosystem without breaking compatibility with any of our existing environments we support, then I am in favor.

@kevin1024 that seems to be the case, I think you're in favor then.

Regarding *need* to migrate https://packaging.python.org/en/latest/discussions/setup-py-deprecated/ could be of interest, but it's not meant as an argument to no migrate.

> There are a number of Linux distributions that package VCR.py from source. I believe if we make this change it may also affect their build scripts, so ideally we would also coordinate with those maintainers.

Speaking with my Gentoo downstream maintainer hat on here: Distros already need to support various ways to build Python software, use of `pyproject.toml` should not cause trouble there or the distro would be in big trouble. At least as long as mainstream values for `build-system` are used, see https://projects.gentoo.org/python/guide/distutils.html#pep-517-build-systems maybe for a list of backends supported in Gentoo. I personally don't expect problems o that front.

deronnax:
@kevin1024 can we get a go :) ?

deronnax:
Having this merged when it was time would have avoid https://github.com/kevin1024/vcrpy/issues/855

pyproject.toml
```diff
@@ -26,3 +26,82 @@ target-version = "py38"
...
+    "yarl",
+    # Support for urllib3 >=2 needs CPython >=3.10
+    # so we need to block urllib3 >=2 for Python <3.10 and PyPy for now.
+    # Note that vcrpy would work fine without any urllib3 around,
+    # so this block and the dependency can be dropped at some point
+    # in the future. For more Details:
+    # https://github.com/kevin1024/vcrpy/pull/699#issuecomment-1551439663
+    "urllib3 <2; python_version <'3.10'",
+    # https://github.com/kevin1024/vcrpy/pull/775#issuecomment-1847849962
+    "urllib3 <2; platform_python_implementation =='PyPy'",
```
pjonsson, line 72: I don't speak for the package, but if this gets wind in the sails now and gets merged, there's a third case for urllib3 added to `setup.py` after this PR was created.

pyproject.toml
```diff
@@ -26,3 +26,82 @@ target-version = "py38"
...
+    "Topic :: Internet :: WWW/HTTP",
+    "License :: OSI Approved :: MIT License",
+]
+urls = {Homepage = "https://github.com/kevin1024/vcrpy"}
+requires-python = ">=3.8"
+dependencies = [
+    "PyYAML",
+    "wrapt",
+    "yarl",
+    # Support for urllib3 >=2 needs CPython >=3.10
```
LewisCowlesMotive, line 64: So what I don't like about this approach, is that it does not attempt to lock. If the non-locking is to preserve some old use-cases for python 3.8, then I suggest making those explicit.

This wild west, "go grab whatever versions of these packages!" is what caused the most breakages, and uncertainty, because it's harder for a dependency resolver to pick the right thing for all the packages.

graingert: Let's leave this PR to migrate the config as is without changes, we can have a follow up PR to change the config

LewisCowlesMotive: What are the benefits of the PR? It declares the same things as the imperative code. Is that the benefit?

graingert: Yes

hartwork:
@deronnax the pull request has conflicts now and needs a rebase with conflict resolution and also an update to match current…

https://github.com/kevin1024/vcrpy/blob/19bd4e012c8fd6970fd7f2af3cc60aed1e5f1ab5/setup.py#L92-L96

…so I cannot approve status quo just yet. After a rebase and conflicts resolved I can do review again. What do you think?

deronnax:
Hello Kevin.
Sorry I won't sugarcoat this: you killed this PR by yourself alone. We repeatedly asked you for a greenlight and questions, which you repeatdly never answered. Hartwork's last question is about a trivial merge conflict that I can fix in seconds. I did not because you never answered our previous more important questions, so what would have been the point?
It could be merged this week-end if you cooperate.

kevin1024:
I'm game if you are.  Reopening, and I'll merge if you are volunteering to:

1. Rebase and fix the merge conflicts
2. Make sure CI passes
3. Are confident this won't break packaging for VCR.py users, including the distros that package vcr.py
4. Volunteer your help when it comes time for me to cut the next release - I'm kind of out of date on the python packaging world these days.

Thanks again for your help on this; I know we are all just volunteering our time.

deronnax:
Thank you very much.
I re-did the changes, against the current master. I still need your greenlight for the CI to run on this PR 🙂 
I used [wheeldiff](https://pypi.org/project/wheeldiff/) to be sure the wheel is the same: it's almost exactly the same : just the "python2-compatible" tag is gone, but it was undue : vcr.py was not python2-compatible.
I don't know every packaging of vcr.py in the world, so I can't swear with my blood none will break, but I can confidently claim the risk is close to zero.
Anyway, would by miracle a problem occure, it's a very easy to revert change, and I will stick around just in case.
And sure, happy to help to cut the next release.

kevin1024:
Merged + released.  Thanks for your contribution.  Let's wait and see if anyone has packaging issues.

hseg:
No issues, but @kevin1024 -- I think you squashed all the commit history when merging 4883e3eefa3ac2f5ed968c487c282861e658031b ? (I don't see f6c46cc138550334240690ec4aca005e9bd29bff which its commit message mentions anywhere in master, only in the history of this PR)
This made it confusing to read the history, and sent me here. Not sure what options Github gives you for this, but `git merge --squash && git commit --author=PRAUTHOR -m NEW_COMMITMSG` would do the trick in CLI

kevin1024:
Yep I bet I clicked the "squash and merge" button.

PR #768, opened 2023-09-13
Fix aiohttp

markis:
**Issue**

There is an issue with aiohttp, when a developer sends a json kwarg it is overwritten by the data kwarg.  In aiohttp the data kwarg is set to None by default and VCR will always default to `None`.

**Resolution**

We should change the kwargs check to check for the json key first and fall back to data.  Because data is defaulted and will always exist.  But json doesn't always exist.

### DIFF
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -159,6 +159,36 @@ def test_post(tmpdir, body, caplog, mockbin_request_url):
 
 
 @pytest.mark.online
+@pytest.mark.asyncio
+async def test_post_data_vs_json(tmpdir, caplog, mockbin_request_url):
+    caplog.set_level(logging.INFO)
+    data = {"key1": "value1", "key2": "value2"}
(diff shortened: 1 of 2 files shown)


jose-lpa:
I _think_ this might not be the final fix to this issue. When I apply it, the results I get for the `body` section comes in unrolled as more YAML, instead of being represented as a JSON string line.

I.e. when I `POST` a JSON body like this one to my app
```json
{"collections": ["fields"], "intersects": {"type": "Point", "coordinates": [12.29683,55.95417]}
```
...it prints this:
Screenshot from 2023-09-28 07-41-21

QuentinBrosse:
Same here, can we merge?

cadlagtrader:
Hello - has this been merged in another fix ?

kevin1024:
Thank you for this fix\! We're going to merge #624 instead which addresses the same issue with some additional edge case handling. We really appreciate you taking the time to contribute\!

PR #771, opened 2023-09-29
docs: Fix list formatting in changelog for v5.1.0

Mr0grog:
This is really just a typo fix. I noticed that the changelog for v5.1.0 is getting rendered as a paragraph instead of a list because it wasn't indented enough:

Screenshot of changelog rendered as a paragraph

This change just fixes the indentation so it renders as a list:

Screenshot of changelog after this change

While I was at it I removed dangling spaces at the end of lines in the changelog (actually my editor just did this automatically, since it’s in the project’s `.editorconfig` file) and added a blank line between the latest 4.x release and 5.0.0, since all the other major versions have extra blank lines between them.


jairhenrique (approved):
LGTM

PR #775, opened 2023-10-06
Finish up on Python 3.12 support

hartwork:
Follow-up to #754

PR #776, opened 2023-10-07
switch to using python 3.12 final release

evgeni:


evgeni:
huh, #775 is probably a better one

PR #777, opened 2023-10-15
Uncap urllib3 version on Python < 3.10

dgw:
Between myself and another maintainer of our downstream project, we've tested Python 3.8 & 3.9 with both OpenSSL 1.1.1 and 3.0.2, using vcrpy 5.1.0 and a forcibly upgraded urllib3 2.0.6. Based on those results, I'm optimistically hoping that this incompatibility is resolved and https://github.com/kevin1024/vcrpy/pull/699#issuecomment-1563444199 has become true.

However, since the test suite already appears to fail on the current HEAD revision prior to applying this change, I'm not sure how to use the test results to definitively verify this patch (except by noting that in manual checks of the relevant environments I didn't see any _new_ failures as compared to running the same tox env on master).

Downstream, we plan to unblock ourselves on the cassette cross-compatibility problem we're having by adding a step to our CI that ensures urllib3 2.x prior to running tests—but if we're lucky, that will be only a temporary kludge.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,13 +46,6 @@ install_requires = [
     "PyYAML",
     "wrapt",
     "yarl",
-    # Support for urllib3 >=2 needs Python >=3.10
-    # so we need to block urllib3 >=2 for Python <3.10 for now.
-    # Note that vcrpy would work fine without any urllib3 around,
-    # so this block and the dependency can be dropped at some point
-    # in the future. For more Details:
-    # https://github.com/kevin1024/vcrpy/pull/699#issuecomment-1551439663
-    "urllib3 <2; python_version <'3.10'",
 ]
 
 tests_require = [
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ envlist =
   cov-clean,
   lint,
   {py38,py39,py310,py311,py312}-{requests-urllib3-1,httplib2,urllib3-1,tornado4,boto3,aiohttp,httpx},
-  {py310,py311,py312}-{requests-urllib3-2,urllib3-2},
+  {py38,py39,py310,py311,py312}-{requests-urllib3-2,urllib3-2},
   {pypy3}-{requests-urllib3-1,httplib2,urllib3-1,tornado4,boto3},
   #{py310}-httpx019,
   cov-report


piiq:
Hi 👋
First of all, a big thank you to all of the maintainers and contributors of vcrpy. Great tool, does it's job.

A <2 pin of urllib3 is causing some troubles for us too and it would be great if this PR could solve it. I wanted to ask @kevin1024 if this PR is actually solves the issue? How can I contribute to testing this out? It would be great if the solution to unpin urllib3 would be found and a new version of vcrpy released?

Cheers and thanks 🙏

hartwork:
If after a rebase CI is all green, this can be merged. If it's not, it may be the same issue as back then, more details at https://github.com/kevin1024/vcrpy/pull/699#issuecomment-1551439663 .

dgw:
Rebased!

> If after a rebase CI is all green

I hope this means the `urllib` envs and does not apply to the `boto3` env that is failing under py3.11 and pypy3.10 on master?

dgw:
@hartwork I mean that every commit with a check status on `master` has been ❌ since c9da7a102fe5b260c0341a5fdf693123f19845a2, and I have never been able to cleanly build this patch because of that. Can't even claim that "it was green when I first opened the PR".

But it appears I will indeed need to look at the test suite and reevaluate something; 3.8 and 3.9 failed on this patch after rebasing. Probably won't get to that today. Would you like this PR placed in draft state until it's ready for another look?

hartwork:
> @hartwork I mean that every commit with a check status on `master` has been ❌ since [c9da7a1](https://github.com/kevin1024/vcrpy/commit/c9da7a102fe5b260c0341a5fdf693123f19845a2), and I have never been able to cleanly build this patch because of that. Can't even claim that "it was green when I first opened the PR".

@dgw yes, that is understood, I should have put myself more clearly, sorry. That red-CI-from-the-begging state is a killer for pull requests and something we need to get rid of with priority. I have created pull request #786 on this topic just now and I hope that @jairhenrique is okay with this approach also.

> But it appears I will indeed need to look at the test suite and reevaluate something; 3.8 and 3.9 failed on this patch after rebasing. Probably won't get to that today. Would you like this PR placed in draft state until it's ready for another look?

It's the same thing that breaks PyPy 3.10 apparently. I'd be good with status "draft", I support the idea.

dgw:
In reference to the errors here, the `ConnectionResetError` shown in CI logs indeed seems similar to what was mentioned back in May. I'll leave myself a link to https://github.com/kevin1024/vcrpy/pull/699#issuecomment-1551249375 as reference; that and your (@hartwork) earlier link should give me a starting point when I sit down to examine the failure more closely.

> I'd be good with status "draft", I support the idea.

Great, I've marked the PR as a draft and will be sure to comment again when (if?) I have something worth looking at again. Meanwhile I will keep an eye on the status of #786.

dgw:
I started rebasing this to try again, just with Python 3.9 this time (rest in peace 3.8), and got stuck locally when I tried to prepare my venv for running tests with `urllib3>=2`. It turns out that `botocore` has _**always**_ capped `urllib3`'s version to something below 2, and it only allows urllib3 2.x on Python 3.10+ as well.

Keeping `boto3` and `urllib3>=2` apart in the GitHub Actions matrix is not at all clear to me. Admittedly I have only a basic understanding of how to define and manipulate GHA matrices—but I _really_ don't know how one would go about skipping a requirement defined in `setup.cfg` based on matrix parameters.

The good news is that if I comment out `boto3` in the `tests` extra and `pip install -e '.[tests]' 'urllib3>=2'` in my Python 3.9 venv, running `./runtests.sh` collects `478 items / 1 skipped` and succeeds with `447 passed, 32 skipped in 25.21s`.

Presumably the skipped item at collection is `integration/test_boto3.py`. The skipped individual tests are all in `integration/test_tornado.py`. `importorskip` and `skipif` annotations in the test files appear to cover all of these.

My rebased branch is (for now) available at https://github.com/dgw/vcrpy/tree/rebase-uncap-urllib3 — pending some more thought (or hints) at how to finish setting up the GHA matrix. Alternatively we can just drop this whole idea. 😅 Python 3.9 has less than a year to live at this point, and it hasn't been as big of a deal as I thought it might be to use a custom fork of `vcrpy` for our CI upstream. This might be too much effort for not enough gains.

hartwork:
@vivodi good point! :+1: Closing…

PR #779, opened 2023-11-07
feat: use regex for ignoring hosts

silviumarcu:
Allow to use regexes for `ignore_hosts` to easily ignore hosts that are frequently changed e.g. `pipelinesghubeus*.actions.githubusercontent.com`

### DIFF
--- a/tests/integration/test_ignore.py
+++ b/tests/integration/test_ignore.py
@@ -44,6 +44,14 @@ def test_ignore_httpbin(tmpdir, httpbin):
             assert len(cass) == 1
 
 
+def test_ignore_httpbin_regex(tmpdir, httpbin):
+    with overridden_dns({"httpbin.org": "127.0.0.1"}):
+        cass_file = str(tmpdir.join("filter_qs.yaml"))
+        with vcr.use_cassette(cass_file, ignore_hosts=[r"h\w*bin.+"]) as cass:
+            urlopen(f"http://httpbin.org:{httpbin.port}/")
+            assert len(cass) == 0
+
+
 def test_ignore_localhost_and_httpbin(tmpdir, httpbin):
     with overridden_dns({"httpbin.org": "127.0.0.1"}):
         cass_file = str(tmpdir.join("filter_qs.yaml"))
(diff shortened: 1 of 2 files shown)


hartwork:
I would like to note that this change would break API backwards-compatibility (e.g. because `.` no longer matches just literal dots) and hence would need a major version bump if merged.

silviumarcu:
@hartwork do you have any suggestion on how to modify it to make it API backwards compatible?

jairhenrique:
The signature of `ignore_hosts` could be like this:

```python
ignore_hosts: set[str | re.Pattern]
```

In the `_build_ignore_hosts` you can check the type of each item of `set` and do the validation for strings and regex.

silviumarcu:
Correct me if I am wrong, but that will work only if the regex pattern is compiled, `re.compile("regex")`. If it's passed as a normal string `r"regex"` then it's just a plain string when checking the type.

vEpiphyte:
Even if the function signature is changed, that still has an interaction with the `ignore_localhost` option which adds multiple items to the (potentially empty) `hosts_to_ignore` set. If the handling of that that option may was changed to just add its own filter function, then the function `_build_ignore_hosts` could change to the following:

```
    @staticmethod
    def _build_ignore_hosts(hosts_to_ignore):

        if isinstance(re.Pattern):
            def filter_ignored_hosts(request):
                # DISCUSS: Would need to decide if search() or match() is best here.
                # search() tends to be more user friendly behavior
                # match() is more specific though.
                if hasattr(request, "host") and hosts_to_ignore.search(request.host):
                    return
                return request

        else:
            hosts_to_ignore = set(hosts_to_ignore)
            def filter_ignored_hosts(request):
                if hasattr(request, "host") and request.host in hosts_to_ignore:
                    return
                return request

        return filter_ignored_hosts

```

Then in the config.py code:

```
        if ignore_localhost:
            filter_functions.append(self._build_ignore_hosts(("localhost", "0.0.0.0", "127.0.0.1")))
        if ignore_hosts:
            filter_functions.append(self._build_ignore_hosts(ignore_hosts))
```

This should be backwards compatible, at the cost of splitting out the localhost filter function from the existing host filters.

Issue #619, opened 2021-11-08
Cannot replay httpx 302


lucemia:
httpx 0.20.0 https://github.com/encode/httpx/releases/tag/0.20.0 change default redirects behaviors that will break vcrpy.

https://github.com/kevin1024/vcrpy/blob/8c0bb73658e300065eddfafc569e656e1ef6f7bf/vcr/stubs/httpx_stubs.py#L101

1. The parameter in httpx called `follow_redirects` not `allow_redirects`

`respose = client.get(url, follow_redirects=True)`

2. the default value should be `False` not `True`

> Unlike requests, HTTPX does not follow redirects by default. 

https://www.python-httpx.org/compatibility/#redirects

## Sample Code

```
import httpx
import pytest


@pytest.mark.vcr()
def test_sample():
    with httpx.Client() as client:
        client.get("https://httpbin.org/status/302")
```

## Output

```
self = <vcr.cassette.Cassette object at 0x10840ce20>, request = <Request (GET) https://httpbin.org/status/302>

    def play_response(self, request):
        """
        Get the response corresponding to a request, but only if it
        hasn't been played back before, and mark it as played
        """
        for index, response in self._responses(request):
            if self.play_counts[index] == 0 or self.allow_playback_repeats:
                self.play_counts[index] += 1
                return response
        # The cassette doesn't contain the request asked for.
>       raise UnhandledHTTPRequestError(
            "The cassette (%r) doesn't contain the request (%r) asked for" % (self._path, request)
        )
E       vcr.errors.UnhandledHTTPRequestError: "The cassette ('/Users/davidchen/repo/ad3p/cassettes/test_sample.yaml') doesn't contain the request (<Request (GET) https://httpbin.org/status/302>) asked for"

venv/lib/python3.8/site-packages/vcr/cassette.py:268: UnhandledHTTPRequestError
```

Issue #600, opened 2021-08-12
Httpx stub cause tests to breaks trying to access httpx.Response elapsed time property


isaquealves:
When building an `httpx` response at `vcr.stubs.httpx_stubs._from_serialized_response`, the elapsed property was not covered. This cause tests to break as the elapsed property is dynamic calculated and depends on the patched method `close`.

Example:

```python 
# script.py
import httpx

def get_elapsed_microsecs():
    response = httpx.get('https://api.github.com/events')
    return round(response.elapsed.total_seconds() * 1000, 2)
```
```python
# test_script.py
import pytest
import vcr

@vcr.use_cassete("cassete.yml")
@pytest.mark.asyncio
def test_get_elapsed_microsecs():
  ''' Breaks raising ".elapsed may ony be accessed after the response has been read or closed" '''
    microsecs = get_elapsed_microsecs()
    assert microsecs >= 0.02

**Comment by parkerhancock** [comment]
I suggest we close this one in view of #784

**Comment by pbasista** [comment]
Unfortunately, even the latest release `v6.0.1` of vcrpy does not seem to resolve this issue. From my experiments it looks like `httpx.Response.elapsed` still cannot be accessed in tests.

As was mentioned in the original message by @isaquealves from almost 3 years ago, the problem is that the `httpx.Response.elapsed` property is only being set by the `httpx.Response.close` method which is being mocked while creating a response from serialized data in `vcr.stubs.httpx_stubs._from_serialized_response` function. And since there is no additional patching done by vcrpy, this property remains unavailable.

A naive fix might look similar to this:

```diff
diff --git a/vcr/stubs/httpx_stubs.py b/vcr/stubs/httpx_stubs.py
index 759cb72..b9fd59f 100644
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -1,4 +1,5 @@
 import asyncio
+from datetime import timedelta
 import functools
 import inspect
 import logging
@@ -100,6 +101,7 @@ def _from_serialized_response(request, serialized_response, history=None):
         history=history or [],
         extensions=extensions,
     )
+    response.elapsed = timedelta(seconds=1)

     return response
```

Perhaps until a proper approach is found, this might be an acceptable work-around.

kevin1024:
[...]

PR #784, opened 2023-12-08
FIX: ALL OF HTTPX SUPPORT (MERGE ME FIRST!!!)

### LINKED ISSUE #656: [BUG] HTTPX Binary upload data unsuported

### LINKED ISSUE #684: httpx auth flow is skipped by replay?

parkerhancock:
Guys, VCRpy's support for httpx is hella broken. Like for serious. So I fixed it. Summary of the changes:

## Changed patch target from .send to ._send_single_request

This change moves the patch target to `client._send_single_request`, which is the lowest, and most atomic request for data over the wire. By moving the patch lower down the stack, it fixes #684 (support for httpx.Auth), #619 (errors handling redirects), #600 (stubs trying to access an invalid property), #599 (encoding types), #597 (async httpx streaming doesn't work). 

## Removed attempt to decode raw body content

For some reason, in the record function, the serializer was trying to decode binary body content (i.e. `content.decode("utf-8", "ignore")`. This one line meant that any raw binary data that happened not to resolve to a valid Unicode code point was deleted, which meant subsequent requests for binary content would be corrupted. This fixes #656 (Binary upload data unsupported), and #597 (async httpx streaming). 

## Simplified Stubs & Test Suite

The stubs and tests have been dramatically pared down and simplified, because we've moved the patch target so much lower. We no longer need to test for redirects or auth flow. The only thing we need to hold httpx's hand on is cookies, which is still tested. Also added a test case for streaming data (#597).

## Replaced mockbin with httpbin

Mockbin went commercial, which meant the whole test suite was broken. I moved all the tests to now target httpbin.org. 

## Other Benefits

I've also:
 - Added a separate fixture for testing streaming, since httpx uses `client.stream` rather than `client.request(. . . , stream=True)`
 - Made corresponding changes so that both httpx.AsyncClient and httpx.Client both work.

**PLEASE PLEASE PLEASE MERGE THIS REQUEST AND SAVE US POOR HTTPX USERS!**

### DIFF
--- a/tests/integration/test_httpx.py
+++ b/tests/integration/test_httpx.py
@@ -9,6 +9,14 @@ httpx = pytest.importorskip("httpx")
 
 from vcr.stubs.httpx_stubs import HTTPX_REDIRECT_PARAM  # noqa: E402
 
+@pytest.fixture(params=["https", "http"])
+def scheme(request):
+    """Fixture that returns both http and https."""
+    return request.param
+
+@pytest.fixture
+def httpbin(scheme):
+    return scheme + "://httpbin.org"
 
 class BaseDoRequest:
     _client_class = None
@@ -16,6 +24,7 @@ class BaseDoRequest:
     def __init__(self, *args, **kwargs):
         self._client_args = args
         self._client_kwargs = kwargs
+        self._client_kwargs['follow_redirects'] = self._client_kwargs.get('follow_redirects', True)
 
     def _make_client(self):
         return self._client_class(*self._client_args, **self._client_kwargs)
@@ -40,7 +49,10 @@ class DoSyncRequest(BaseDoRequest):
 
     def __call__(self, *args, **kwargs):
         return self.client.request(*args, timeout=60, **kwargs)
-
+    
+    def stream(self, *args, **kwargs):
+        with self.client.stream(*args, **kwargs) as response:
+            return b"".join(response.iter_bytes())
 
 class DoAsyncRequest(BaseDoRequest):
     _client_class = httpx.AsyncClient
@@ -75,8 +87,22 @@ class DoAsyncRequest(BaseDoRequest):
 
         # Use one-time context and dispose of the loop/client afterwards
         with self:
-            return self(*args, **kwargs)
+            return self._loop.run_until_complete(self.client.request(*args, **kwargs))
+    
+    async def _get_stream(self, *args, **kwargs):
+        async with self.client.stream(*args, **kwargs) as response:
+            content = b""
+            async for c in response.aiter_bytes():
+                content += c
+        return content
+                
+    def stream(self, *args, **kwargs):
+        if hasattr(self, "_loop"):
+            return self._loop.run_until_complete(self._get_stream(*args, **kwargs))
 
+        # Use one-time context and dispose of the loop/client afterwards
+        with self:
+            return self._loop.run_until_complete(self._get_stream(*args, **kwargs))
 
 def pytest_generate_tests(metafunc):
     if "do_request" in metafunc.fixturenames:
@@ -89,8 +115,8 @@ def yml(tmpdir, request):
 
 
 @pytest.mark.online
-def test_status(tmpdir, mockbin, do_request):
-    url = mockbin
+def test_status(tmpdir, httpbin, do_request):
+    url = httpbin
 
     with vcr.use_cassette(str(tmpdir.join("status.yaml"))):
         response = do_request()("GET", url)
@@ -102,8 +128,8 @@ def test_status(tmpdir, mockbin, do_request):
 
 
 @pytest.mark.online
-def test_case_insensitive_headers(tmpdir, mockbin, do_request):
-    url = mockbin
+def test_case_insensitive_headers(tmpdir, httpbin, do_request):
+    url = httpbin
 
     with vcr.use_cassette(str(tmpdir.join("whatever.yaml"))):
         do_request()("GET", url)
@@ -116,8 +142,8 @@ def test_case_insensitive_headers(tmpdir, mockbin, do_request):
 
 
 @pytest.mark.online
-def test_content(tmpdir, mockbin, do_request):
-    url = mockbin
+def test_content(tmpdir, httpbin, do_request):
+    url = httpbin
 
     with vcr.use_cassette(str(tmpdir.join("cointent.yaml"))):
         response = do_request()("GET", url)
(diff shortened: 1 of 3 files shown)


vcr/stubs/httpx_stubs.py
```diff
@@ -38,7 +38,7 @@ def _to_serialized_response(httpx_response):
         "status_code": httpx_response.status_code,
         "http_version": httpx_response.http_version,
         "headers": _transform_headers(httpx_response),
-        "content": httpx_response.content.decode("utf-8", "ignore"),
+        "content": httpx_response.content#.decode("utf-8", "ignore"),
```
jairhenrique, line 41:
```suggestion
        "content": httpx_response.content
```

jairhenrique:
@parkerhancock can you help us moving mockbin tests to httpbin?

tests/integration/test_httpx.py
```diff
@@ -275,42 +273,19 @@ def response_cookies(response):
...
     with vcr.use_cassette(testfile) as cassette:
-        response = do_request()("GET", url, **redirect_kwargs)
-        assert len(response.history) == 2
-        assert response.json()["url"].endswith("request")
-
-        assert cassette.play_count == 3
-
-
-@pytest.mark.online
-def test_redirect_wo_allow_redirects(do_request, mockbin, yml):
```
jairhenrique, line 299: I think it would be cool to keep the tests with follow redirects param with true and false

vcr/stubs/httpx_stubs.py
parkerhancock: fixed in commit below.

parkerhancock:
> @parkerhancock can you help us moving mockbin tests to httpbin?

Test suite was failing on this PR, so I fixed it for you. All tests on all backends now pass on py311, supported by the pytest-httpbin plugin (local server, no calls to external servers). 

mockbin has been summarily executed.

tests/integration/test_proxy.py
parkerhancock, line 1: These changes were necessary because trying to spawn a new process lead to an exception caused by trying to pickle a thread lock. So the same proxy process now runs in a separate thread, rather than a separate process.

jairhenrique (approved):

parkerhancock:
> @parkerhancock LGTM

🙏 YOU ARE MY HERO! LET'S GO 🚀 🚀 🚀 

Any idea when we can get this pushed out to PyPI in a release? It is a blocker to having a decent release over at [patent_client](https://github.com/parkerhancock/patent_client)

(And thanks again for your prompt review!)

ionelmc:
I've tried this, it works. Just a small note, I found out that I needed to recreate some of my cassettes for tests that dealt with binary content as the previous implementation incorrectly saved responses as text instead of bytes.

PR #786, opened 2023-12-08
Drop broken `pypy-3.10` (CI kept failing; alternative to #787)

hartwork:
In reaction to https://github.com/kevin1024/vcrpy/pull/775#issuecomment-1847849962
Alternative to #787

PR #787, opened 2023-12-08
Fix CI / Block urllib3 >=2 for PyPy (alternative to #786)

hartwork:
In reaction to https://github.com/kevin1024/vcrpy/pull/775#issuecomment-1847849962
Alternative to #786

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,13 +46,15 @@ install_requires = [
     "PyYAML",
     "wrapt",
     "yarl",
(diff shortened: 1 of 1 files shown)


jairhenrique (approved):

hartwork:
@josumoreno-BP this is not the right place for a bug report and after re-checking the change in here, it still seems correct. I'm not sure I understand the poetry part. if this matter it will need a dedicated bug report and more elaboration, at least with me. Thanks.

PR #791, opened 2023-12-10
Migrate linting from a tox environment to a GitHub Actions action

jairhenrique:
Trying to make the test reports clearer and simpler...


hartwork (changes requested):
@jairhenrique great idea! Could you drop the `tox -e lint` from `docs/contributing.rst` also?

hartwork (approved):
@jairhenrique thanks! :+1:

PR #792, opened 2023-12-10
Save CI resources + document why each version is covered

hartwork:


hartwork:
@jairhenrique do you have any concerns on saving CPython 3.11 in the CI? Do you have a minute to review?

jairhenrique:
@hartwork for me it doesn't make sense to remove python 3.11 from the testing matrix. I believe we should run the tests on all supported python versions as a minimum guarantee.

PR #794, opened 2023-12-11
Run tests without tox

jairhenrique:
This change remove flaky errors on ci.
Tests are faster with dependency caching.


jairhenrique:
@hartwork What do you think about this approach? I didn't have any strange errors when running the tests, the suite was executed without the need to run the pipeline to force the tests to turn green.

hartwork (changes requested):
@jairhenrique this reduces coverage with regard to covering with urllib3 1.x and 2.x for the same version of Python, e.g. for Python 3.10 we cover:

- `py310-requests-urllib3-1`
- `py310-urllib3-1`
- `py310-requests-urllib3-2`
- `py310-urllib3-2`

Too keep the same level of coverage and be protected against regressions this would probably need the [`include` variant of the matrix feature of GitHub Actions](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations). And if we take this route we probably want to git rid of tox altogether to not have to maintain it and not have it uncovered. (Dropping tox will affect which tests contributors are running locally one way or another also.)

.github/workflows/main.yml
```diff
@@ -13,7 +13,21 @@ jobs:
...
+          [
+            "3.8",
+            "3.9",
+            "3.10",
+            "3.11",
+            "3.12",
+            "pypy-3.8",
+            "pypy-3.9",
+            "pypy-3.10",
+          ]
```
hartwork, line 26: This could be a lot simpler, even in multiline mode:

```yaml
- 3.8
- 3.9
- "3.10"
- 3.11
- 3.12
- pypy-3.8
- pypy-3.9
- pypy-3.10
```
Please use a non-JSON YAML list or revert this change.

.github/workflows/main.yml
```diff
@@ -13,7 +13,21 @@ jobs:
...
+        urllib3-version: ["urllib3"]
```
hartwork, line 27: - `urllib3-version` is rather misleading for a name, in pip lingo this is a "requirement specifier". How about `urllib3-dependency` or `urllib3-requirement`?
- Value `urllib3>=2` would make things more obvious

.github/workflows/main.yml
```diff
@@ -13,7 +13,21 @@ jobs:
...
+        include:
+          - python-version: "3.10"
+            urllib3-version: "urllib3<2"
```
hartwork, line 30: Is there a reason to not do the same for Python 3.11 and 3.12?

.github/workflows/main.yml
```diff
@@ -22,22 +36,16 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
+          cache: pip
 
       - name: Install project dependencies
         run: |
-          pip3 install --upgrade pip
-          pip3 install codecov tox tox-gh-actions
+          pip install --upgrade pip
+          pip install codecov .[tests]
+          pip install '${{ matrix.urllib3-version }}'
```
hartwork, line 45: - The first of these two always installs urllib3 2.x without apparent need to. Let's combine these two lines into one?
- Let's append `pip check` at the end to avoid unknown dependency issues
- (Let's add `set -x` to the first line to help CI logs make sense. It's hard to see where each command ends right now.)

setup.py
```diff
@@ -57,24 +57,29 @@ def run_tests(self):
...
-    # Needed to un-break httpbin 0.7.0. For httpbin >=0.7.1 and after,
-    # this pin and the dependency itself can be removed, provided
-    # that the related bug in httpbin has been fixed:
-    # https://github.com/kevin1024/vcrpy/issues/645#issuecomment-1562489489
-    # https://github.com/postmanlabs/httpbin/issues/673
-    # https://github.com/postmanlabs/httpbin/pull/674
-    "Werkzeug==2.0.3",
-]
+tests_require = {
+    "tests": [
```
hartwork, line 61: This variable should be named `extras_require`.

setup.py
```diff
@@ -89,7 +94,7 @@ def run_tests(self):
     python_requires=">=3.8",
     install_requires=install_requires,
     license="MIT",
-    tests_require=tests_require,
+    extras_require=tests_require,
```
hartwork, line 97: Is there value in dropping `tests_require` rather than using both? How about:

```
[..]
extras_require=extras_require,
tests_require=extras_require['tests'],
[..]
```

hartwork (changes requested):
@jairhenrique the general direction is good, some details need fixing:

- `git grep tox` shows more places that need adjustment:
  ```console
  # git --no-pager grep tox
  .gitignore:.tox
  runtests.sh:# https://blog.ionelmc.ro/2015/04/14/tox-tricks-and-patterns/#when-it-inevitably-leads-to-shell-scripts
  runtests.sh:# You can and should use WSL for running tox on Windows when it calls bash scripts.
  ```
  
- The pull request does 5+ different things all in one commit. It would be great to split this up into semantic units using `git rebase -i`.

- Plus everything I commented below:

.github/workflows/main.yml
```diff
@@ -22,22 +40,17 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
 
       - name: Install project dependencies
         run: |
-          pip3 install --upgrade pip
-          pip3 install codecov tox tox-gh-actions
+          set -xe
```
hartwork, line 47: @jairhenrique `set -e`is anabled in GitHub Actions by default.

.github/workflows/main.yml
```diff
@@ -22,22 +45,16 @@ jobs:
...
-
-      - name: Run offline tests with tox with no access to the Internet
-        run: |
-          # We're using unshare to take Internet access
-          # away from tox so that we'll notice whenever some new test
-          # is missing @pytest.mark.online decoration in the future
-          unshare --map-root-user --net -- \
-              sh -c 'ip link set lo up; tox -- -m "not online"'
+      - name: Run tests
+        run: ./runtests.sh --cov=./vcr --cov-branch --cov-report=xml --cov-append
```
hartwork, line 57: @jairhenrique could you bring back the split between `-m online` and `-m 'not online'` please? Where loosing something of value here. Thanks!

hartwork (changes requested):
@jairhenrique `exclude` is a nice idea — it's more future-proof than `include` — but there is one key showstopper left below:

hartwork:
> @jairhenrique the general direction is good, some details need fixing:
> 
> * `git grep tox` shows more places that need adjustment:
>   ```
>   # git --no-pager grep tox
>   .gitignore:.tox
>   runtests.sh:# https://blog.ionelmc.ro/2015/04/14/tox-tricks-and-patterns/#when-it-inevitably-leads-to-shell-scripts
>   runtests.sh:# You can and should use WSL for running tox on Windows when it calls bash scripts.
>   ```
> 
> * The pull request does 5+ different things all in one commit. It would be great to split this up into semantic units using `git rebase -i`.
> 
> * Plus everything I commented below:

@jairhenrique did you see this^^?

hartwork (commented):
@jairhenrique thanks for the adjustments! Is keeping `.tox` in `.gitignore` intentional?

jairhenrique:
> @jairhenrique thanks for the adjustments! Is keeping `.tox` in `.gitignore` intentional?

@hartwork yes, to prevent someone from uploading this folder to the repository.

hartwork:
> > @jairhenrique thanks for the adjustments! Is keeping `.tox` in `.gitignore` intentional?
> 
> @hartwork yes, to prevent someone from uploading this folder to the repository.

@jairhenrique I'm not sure I understand — wouldn't we see an addition like that in the pull request? Please me understand.

jairhenrique:
> @jairhenrique I'm not sure I understand — wouldn't we see an addition like that in the pull request when it's happening? Please help me understand.

Yes, we would see. I just left it so I would have to avoid this work. I can remove it if you understand that this is a problem.

hartwork (approved):
> > @jairhenrique I'm not sure I understand — wouldn't we see an addition like that in the pull request when it's happening? Please help me understand.
> 
> Yes, we would see. I just left it so I would have to avoid this work. I can remove it if you understand that this is a problem.

@jairhenrique I'll leave it to you, we'll probably be good both ways.

Thanks for the other adjustments, the nice split up into multiple commits — this pull request has come a long way! :+1:

graingert:
This broke the tornado tests: https://github.com/kevin1024/vcrpy/actions/runs/7620100960/job/20754258427?pr=794#step:5:38 note all the warnings

jairhenrique:
> actually it looks like it was broken before

@graingert yes 😞

PR #799, opened 2023-12-12
Show ruff diff errors

jairhenrique:


hartwork (approved):
@jairhenrique thank you! :+1:

PR #800, opened 2023-12-12
Start using pre-commit in CI

hartwork:
@jairhenrique I have added one deliberate formatting error to showcase how this integration will present formatting errors in practice. Looking forward to your review :beers: 

PS: Review on commit level is advised :smiley:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under the MIT license
+
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.7
+    hooks:
+      - id: ruff
+      - id: ruff-format
```
jairhenrique, line 9:
```suggestion
      - id: ruff-format
        args: [ --check, --diff ]
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under the MIT license
+
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.7
+    hooks:
+      - id: ruff
```
jairhenrique, line 8:
```suggestion
      - id: ruff
        args: [--show-source]
```

.pre-commit-config.yaml
hartwork: @jairhenrique I think the current CI output shows that we already get a diff and that hence this is not needed, but I can add them for an experiment to see how the results differ with these, give me a second…

.pre-commit-config.yaml
hartwork: @jairhenrique good idea, done with next push in a minute

.pre-commit-config.yaml
hartwork: @jairhenrique actually I just noticed that this addition will break auto-formatting. Currently you can use pre-commit to have it do the formatting for you, with this change adding `--check` you no longer can. So status quo is actually what we want, regarding this detail. I have add `--diff` as suggested though, since it may be of help during local development. Coming up with next push in a second

hartwork:
@jairhenrique cool! Let me revert the error-trigger commit and see if CI gets all green…

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under the MIT license
+
+name: Run pre-commit
+
+on:
+- pull_request
+- push
```
jairhenrique, line 8: To avoid running twice on pull request.

```suggestion
- push
    branches:
      - master
```

jairhenrique: @hartwork 
image

hartwork: @jairhenrique that's actually a feature since the code in isolation and the code merged are not necessarily the same (e.g. when master moved in the mean time) and both states should arguably be tested. I know it's not soemthing that everyone agrees upon. If it means a lot to you I can change it, but I consider this a feature rather than a bug.

jairhenrique (approved):
🔁 👍 🚀

hartwork:
@jairhenrique cool! Merging…

PR #809, opened 2024-01-22
Fix HTTPS proxy handling

alga:
When HTTPS requests were made through a proxy, the proxy host name and port ended up in the cassette URLs.

In this PR, I extend the `Proxy` fixture to handle the `CONNECT` method, reproduce the bug and fix it.

### DIFF
--- a/tests/integration/test_proxy.py
+++ b/tests/integration/test_proxy.py
@@ -85,8 +85,6 @@ def test_use_proxy(tmpdir, httpbin, proxy_server):
     with vcr.use_cassette(str(tmpdir.join("proxy.yaml")), mode="none") as cassette:
         cassette_response = requests.get(httpbin.url, proxies={"http": proxy_server})
 
-    for key in set(cassette_response.headers.keys()) & set(response.headers.keys()):
-        assert cassette_response.headers[key] == response.headers[key]
     assert cassette_response.headers == response.headers
     assert cassette.play_count == 1
 
@@ -102,8 +100,6 @@ def test_use_https_proxy(tmpdir, httpbin_secure, proxy_server):
             proxies={"https": proxy_server},
         )
 
-    for key in set(cassette_response.headers.keys()) & set(response.headers.keys()):
-        assert cassette_response.headers[key] == response.headers[key]
     assert cassette_response.headers == response.headers
     assert cassette.play_count == 1
 
--- a/tests/integration/test_proxy.py
+++ b/tests/integration/test_proxy.py
@@ -82,7 +82,7 @@ def test_use_proxy(tmpdir, httpbin, proxy_server):
     with vcr.use_cassette(str(tmpdir.join("proxy.yaml"))):
         response = requests.get(httpbin.url, proxies={"http": proxy_server})
 
-    with vcr.use_cassette(str(tmpdir.join("proxy.yaml")), mode="once") as cassette:
+    with vcr.use_cassette(str(tmpdir.join("proxy.yaml")), mode="none") as cassette:
         cassette_response = requests.get(httpbin.url, proxies={"http": proxy_server})
 
     for key in set(cassette_response.headers.keys()) & set(response.headers.keys()):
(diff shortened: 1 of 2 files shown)


graingert:
Previously you set the host of the proxy server to an explicit 127.0.0.x IP. Is that needed?

alga:
> Previously you set the host of the proxy server to an explicit 127.0.0.x IP. Is that needed?

Not strictly speaking, just the proxy having a different IP than httpbin was convenient for debugging and made the test a bit more precise. If the URLs vary just by the port number, there can conceivably be a regression uncaught by tests if the host selection logic breaks.

I reverted it to minimize the random changes in the diff when resolving a conflict. Do you think it's worth putting it back?

alga:
Hello, guys, this is a genuine bugfix with a repro.

agroszer:
please merge this

agroszer:
@graingert any chance to merge this before the next release?

jairhenrique (approved):

agroszer:
@graingert please merge and release

agroszer:
or @kevin1024 
thank you

jairhenrique:
@hartwork can you help to review this!?

tests/integration/test_proxy.py
```diff
@@ -59,3 +89,23 @@ def test_use_proxy(tmpdir, httpbin, proxy_server):
...
     assert cassette_response.headers == response.headers
     assert cassette.play_count == 1
+
+
+def test_use_https_proxy(tmpdir, httpbin_secure, proxy_server):
+    """Ensure that it works with an HTTPS proxy."""
+    with vcr.use_cassette(str(tmpdir.join("proxy.yaml"))):
+        response = requests.get(httpbin_secure.url, proxies={"https": proxy_server})
+
+    with vcr.use_cassette(str(tmpdir.join("proxy.yaml")), mode="once") as cassette:
```
hartwork, line 99:
```suggestion
    with vcr.use_cassette(str(tmpdir.join("proxy.yaml")), mode="none") as cassette:
```
?

https://vcrpy.readthedocs.io/en/latest/usage.html#record-modes

tests/integration/test_proxy.py
```diff
@@ -59,3 +89,23 @@ def test_use_proxy(tmpdir, httpbin, proxy_server):
...
+        cassette_response = requests.get(
+            httpbin_secure.url,
+            proxies={"https": proxy_server},
+        )
+
+    for key in set(cassette_response.headers.keys()) & set(response.headers.keys()):
+        assert cassette_response.headers[key] == response.headers[key]
```
hartwork, line 106:
```suggestion
```
?

Seems to be fully covered by the upcoming line `assert cassette_response.headers == response.headers`. Am I missing something?

tests/integration/test_proxy.py
```diff
@@ -59,3 +89,23 @@ def test_use_proxy(tmpdir, httpbin, proxy_server):
...
+    assert cassette_response.headers == response.headers
+    assert cassette.play_count == 1
+
+    # The cassette url points to httpbin, not the proxy
```
hartwork, line 110:
```suggestion
    # The cassette URL points to httpbin, not the proxy
```
(not critical, true)

hartwork (approved):
> @hartwork can you help to review this!?

@jairhenrique I have some open questions about proxying in general, but I had a look now and played with it a bit and what I found that should likely be changed — please see below — is non-critical; looks like good work in general — approving.

CC @alga

tests/integration/test_proxy.py
alga: You're probably right, I copied these assertions from `test_use_proxy` above. I assumed whoever wrote that might know more than I do. Should I clean it up there as well?

tests/integration/test_proxy.py
alga: Again, I used an existing test above as a template. Should I fix both?

tests/integration/test_proxy.py
hartwork: @alga I think there are two ways to look at it:
- a) it is considered out of scope and is not fixed in here
- b) it is considered loosely related (or boy scout rule is applied) and then it nice to improve this finding elsewhere also *but* in separate commits so that it's clear that is not strongly related to the core matter of the pull request.

I'm good either way, please choose as you prefer.

agroszer:
thank you
@kevin1024 any chance for a release please?

PR #810, opened 2024-01-23
fix tornado related warnings by replacing pytest-tornado with a simple coroutine runner

graingert:


setup.py
```diff
@@ -66,7 +66,6 @@ def run_tests(self):
         "pytest-aiohttp",
         "pytest-asyncio",
         "pytest-cov",
-        "pytest-httpbin",
         "pytest-tornado",
```
graingert, line 69:
```suggestion
        "pytest-httpbin",
```

jairhenrique (approved):

PR #812, opened 2024-01-23
remove utcfromtimestamp DeprecationWarning ignore

graingert:
Tornado doesn't call this anymore

We used to test on tornado4 and latest so this was a problem. It might be worth dropping support for old tornado versions


graingert:
Nope it's still needed

PR #820, opened 2024-02-05
fix(stubs/httpx): fix decode error on utf-8 response with file

kszinhu:
On httpx response should be read with ignore error

@hartwork


76tr75b:
Can we get this merged? It solves the issue when recording file uploads with httpx.

hartwork (changes requested):
Please explain why this fix would make sense. In particular hardcoding `ISO-8859-1` and serving bytes on one branch but a strong in the other seem… *very* surprising. What is the idea?

PR #823, opened 2024-02-16
Improve test compatibility with legacy httpbin index

mgorny:
Make the tests slightly more flexible to match both the flasgger-based and legacy httpbin index.  This is needed for compatibility with https://github.com/psf/httpbin/pull/44 when flasgger is not installed (e.g. on architectures that are not supported by Rust).


mgorny:
Thanks!

Issue #846, opened 2024-06-27
Chunked body as an iterator doesn't work when body is a streamed iterator


sathieu:
https://github.com/kevin1024/vcrpy/pull/739 fixed chunked response for most of the usecases, but this is not enough :wink: .

When `list(body)`  is processed, this consumes the iterator.:

https://github.com/kevin1024/vcrpy/blob/86b114f2f569a9914b182718ab62090ad5913ba6/vcr/matchers.py#L91

But this code is called many times, and all but the first call will return an empty list.

This could be worked-around with:

```diff
diff --git a/vcr/stubs/__init__.py b/vcr/stubs/__init__.py
index 4d4bb39..c2af59d 100644
--- a/vcr/stubs/__init__.py
+++ b/vcr/stubs/__init__.py
@@ -254,6 +254,9 @@ class VCRConnection:
         """Retrieve the response"""
         # Check to see if the cassette has a response for this request. If so,
         # then return it
+        if self._vcr_request.headers.get('Transfer-Encoding', '') == 'chunked':
+            if not (isinstance(self._vcr_request.body, str) or isinstance(self._vcr_request.body, bytes) or isinstance(self._vcr_request.body, bytearray)):
+                self._vcr_request.body = list(self._vcr_request.body)
         if self.cassette.can_play_response_for(self._vcr_request):
             log.info(f"Playing response for {self._vcr_request} from cassette")
             response = self.cassette.play_response(self._vcr_request)
```

But this is hackish and probably not the right place ...

sathieu:
Here is a better patch, for stream objects (`BytesIO`) too:

```diff
diff --git a/vcr/stubs/__init__.py b/vcr/stubs/__init__.py
index 4d4bb39..2ea9fdf 100644
--- a/vcr/stubs/__init__.py
+++ b/vcr/stubs/__init__.py
@@ -254,6 +254,12 @@ class VCRConnection:
         """Retrieve the response"""
         # Check to see if the cassette has a response for this request. If so,
         # then return it
+        if self._vcr_request.headers.get('Transfer-Encoding', '') == 'chunked':
+            if not (isinstance(self._vcr_request.body, str) or isinstance(self._vcr_request.body, bytes) or isinstance(self._vcr_request.body, bytearray)):
+                if hasattr(self._vcr_request.body, 'read'):
+                    self._vcr_request.body = self._vcr_request.body.read()
+                else:
+                    self._vcr_request.body = list(self._vcr_request.body)
[...]

PR #847, opened 2024-06-27
Ensure body is consumed only once (alternative to #851)

sathieu:
Fixes: #846

### DIFF
--- a/tests/unit/test_stubs.py
+++ b/tests/unit/test_stubs.py
@@ -1,9 +1,10 @@
 import contextlib
+import http.client as httplib
 from unittest import mock
 
 from pytest import mark
 
-from vcr import mode
+from vcr import mode, use_cassette
 from vcr.cassette import Cassette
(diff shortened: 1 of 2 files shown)


sathieu:
@hartwork Please take a look at this PR :pray: .

hartwork:
@sathieu I had a quick look and I think understand the problem now but from a quick look the patch raises questions, may not be the best place to patch yet, and is also lacking a new test to both document and regression-protect the case that it's fixing. I have other priorities at the moment, I won't be able to help more at the moment, thanks for your understanding.

hartwork (changes requested):
Marking as "Changes requested" due to CI failing at multiple places and the comment above (in particular a missing test)…

sathieu:
Thanks @hartwork I've fixed CI and added tests. Back to you :ping_pong: !

hartwork:
Superseded by just-merged #851, closing…

PR #851, opened 2024-07-07
Ensure body is consumed only once (alternative to #847)

sathieu:
Fixes: #846

### DIFF
--- a/tests/unit/test_stubs.py
+++ b/tests/unit/test_stubs.py
@@ -1,9 +1,12 @@
 import contextlib
+import http.client as httplib
+from io import BytesIO
+from tempfile import NamedTemporaryFile
 from unittest import mock
 
 from pytest import mark
 
-from vcr import mode
+from vcr import mode, use_cassette
 from vcr.cassette import Cassette
 from vcr.stubs import VCRHTTPSConnection
 
@@ -21,3 +24,52 @@ class TestVCRConnection:
             vcr_connection.cassette = Cassette("test", record_mode=mode.ALL)
             vcr_connection.real_connection.connect()
             assert vcr_connection.real_connection.sock is not None
+
+    def test_body_consumed_once_stream(self, tmpdir, httpbin):
+        self._test_body_consumed_once(
+            tmpdir,
+            httpbin,
+            BytesIO(b"1234567890"),
+            BytesIO(b"9876543210"),
+            BytesIO(b"9876543210"),
+        )
+
+    def test_body_consumed_once_iterator(self, tmpdir, httpbin):
+        self._test_body_consumed_once(
+            tmpdir,
+            httpbin,
+            iter([b"1234567890"]),
+            iter([b"9876543210"]),
+            iter([b"9876543210"]),
+        )
+
+    # data2 and data3 should serve the same data, potentially as iterators
+    def _test_body_consumed_once(
+        self,
+        tmpdir,
+        httpbin,
+        data1,
+        data2,
+        data3,
+    ):
+        with NamedTemporaryFile(dir=tmpdir, suffix=".yml") as f:
+            testpath = f.name
+            # NOTE: ``use_cassette`` is not okay with the file existing
+            #       already.  So we using ``.close()`` to not only
+            #       close but also delete the empty file, before we start.
+            f.close()
+            host, port = httpbin.host, httpbin.port
+            match_on = ["method", "uri", "body"]
+            with use_cassette(testpath, match_on=match_on):
+                conn1 = httplib.HTTPConnection(host, port)
+                conn1.request("POST", "/anything", body=data1)
+                conn1.getresponse()
+                conn2 = httplib.HTTPConnection(host, port)
+                conn2.request("POST", "/anything", body=data2)
+                conn2.getresponse()
+            with use_cassette(testpath, match_on=match_on) as cass:
+                conn3 = httplib.HTTPConnection(host, port)
+                conn3.request("POST", "/anything", body=data3)
+                conn3.getresponse()
+            assert cass.play_counts[0] == 0
+            assert cass.play_counts[1] == 1
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -1,0 +1,33 @@
+from io import BytesIO, StringIO
+
+import pytest
+
+from vcr import request
+from vcr.util import read_body
+
+
+@pytest.mark.parametrize(
+    "input_, expected_output",
+    [
+        (BytesIO(b"Stream"), b"Stream"),
+        (StringIO("Stream"), b"Stream"),
+        (iter(["StringIter"]), b"StringIter"),
+        (iter(["String", "Iter"]), b"StringIter"),
+        (iter([b"BytesIter"]), b"BytesIter"),
+        (iter([b"Bytes", b"Iter"]), b"BytesIter"),
+        (iter([70, 111, 111]), b"Foo"),
+        (iter([]), b""),
+        ("String", b"String"),
+        (b"Bytes", b"Bytes"),
+    ],
+)
+def test_read_body(input_, expected_output):
+    r = request.Request("POST", "http://host.com/", input_, {})
+    assert read_body(r) == expected_output
+
+
+def test_unsupported_read_body():
+    r = request.Request("POST", "http://host.com/", iter([[]]), {})
+    with pytest.raises(ValueError) as excinfo:
+        assert read_body(r)
+    assert excinfo.value.args == ("Body type <class 'list'> not supported",)
--- a/vcr/util.py
+++ b/vcr/util.py
@@ -89,9 +89,28 @@ def compose(*functions):
     return composed
 
 
+def _is_nonsequence_iterator(obj):
+    return hasattr(obj, "__iter__") and not isinstance(
+        obj,
+        (bytearray, bytes, dict, list, str),
+    )
+
+
 def read_body(request):
     if hasattr(request.body, "read"):
         return request.body.read()
+    if _is_nonsequence_iterator(request.body):
+        body = list(request.body)
+        if body:
+            if isinstance(body[0], str):
+                return "".join(body).encode("utf-8")
+            elif isinstance(body[0], (bytes, bytearray)):
+                return b"".join(body)
+            elif isinstance(body[0], int):
+                return bytes(body)
+            else:
+                raise ValueError(f"Body type {type(body[0])} not supported")
+        return b""
     return request.body
 
 
--- a/vcr/request.py
+++ b/vcr/request.py
@@ -3,7 +3,7 @@ import warnings
 from io import BytesIO
 from urllib.parse import parse_qsl, urlparse
 
-from .util import CaseInsensitiveDict
+from .util import CaseInsensitiveDict, _is_nonsequence_iterator
 
 log = logging.getLogger(__name__)
 
@@ -17,8 +17,11 @@ class Request:
         self.method = method
         self.uri = uri
         self._was_file = hasattr(body, "read")
+        self._was_iter = _is_nonsequence_iterator(body)
         if self._was_file:
             self.body = body.read()
+        elif self._was_iter:
+            self.body = list(body)
         else:
             self.body = body
         self.headers = headers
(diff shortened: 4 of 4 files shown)


sathieu:
@hartwork Here is another way to fix #846. Probably better, because at the same place as streams handling.

tests/unit/test_util.py
```diff
@@ -0,0 +1,23 @@
...
+        (iter(["StringIter"]), b"StringIter"),
+        (iter([b"BytesIter"]), b"BytesIter"),
+        (iter([70, 111, 111]), b"Foo"),
+        ("String", b"String"),
+        (b"Bytes", b"Bytes"),
+    ],
+)
+def test_read_body(input, output):
+    r = request.Request("POST", "http://host.com/", input, {})
+    assert read_body(r) == output
```
hartwork, line 23:
```suggestion
def test_read_body(input_, expected_output):
    r = request.Request("POST", "http://host.com/", input_, {})
    assert read_body(r) == expected_output
```

`input` name-collides with is a [Python built-in](https://docs.python.org/3/library/functions.html#built-in-functions).

At least one more cases like that in this pull requests.

tests/unit/test_stubs.py
```diff
@@ -21,3 +23,52 @@ def testing_connect(*args):
...
+            conn2.request("POST", "/anything", body=data2)
+            resp2 = conn2.getresponse()
+            assert resp2.status == 501  # Chunked request Not implemented
+        with use_cassette(testpath, match_on=match_on) as cass:
+            conn3 = httplib.HTTPConnection(host, port)
+            conn3.request("POST", "/anything", body=data3)
+            resp3 = conn3.getresponse()
+            assert resp3.status == 501  # Chunked request Not implemented
+            assert cass.play_counts[0] == 0
+            assert cass.play_counts[1] == 1
```
hartwork, line 74: This secretly assumes and only works if `data2` and `data3` served the same data. Let's at least add a comment about that please if we cannot easily `assert` about due to `data2` and `data3` potentially being iterators.

tests/unit/test_stubs.py
```diff
@@ -21,3 +23,52 @@ def testing_connect(*args):
...
+    def _test_body_consumed_once(
+        self,
+        tmpdir,
+        httpbin,
+        testfile,
+        data1,
+        data2,
+        data3,
+    ):
+        testpath = str(tmpdir.join(testfile))
```
hartwork, line 56: Any particular reason why we want control over the precise filename here? If not, `with NamedTemporaryFile(...)` would work: 
- `suffix=".yml"` allows to control the file extension if needed
- `dir=tmpdir` allows using the directory we have been given.

It would resolve one no-longer-necessary parameter. I think it could help simplicity.

tests/unit/test_stubs.py
```diff
@@ -21,3 +23,52 @@ def testing_connect(*args):
...
+        host, port = httpbin.host, httpbin.port
+        match_on = ["method", "uri", "body"]
+        with use_cassette(testpath, match_on=match_on):
+            conn1 = httplib.HTTPConnection(host, port)
+            conn1.request("POST", "/anything", body=data1)
+            resp1 = conn1.getresponse()
+            assert resp1.status == 501  # Chunked request Not implemented
```
hartwork, line 63: Please help me see how the checking for `.status` helps the test. Can these three asserts be just dropped?

tests/unit/test_util.py
```diff
@@ -0,0 +1,23 @@
...
+from vcr.util import read_body
+
+
+@pytest.mark.parametrize(
+    "input, output",
+    [
+        (BytesIO(b"Stream"), b"Stream"),
+        (StringIO("Stream"), b"Stream"),
+        (iter(["StringIter"]), b"StringIter"),
+        (iter([b"BytesIter"]), b"BytesIter"),
```
hartwork, line 15: Maybe the last two should (a) have more than one list element or (b) be duplicated so that we cover single and non-single cases. What do you think?

vcr/util.py
```diff
@@ -92,6 +92,21 @@ def composed(incoming):
 def read_body(request):
     if hasattr(request.body, "read"):
         return request.body.read()
+    if hasattr(request.body, "__iter__") and not isinstance(
+        request.body,
+        (bytearray, bytes, dict, list, str),
```
hartwork, line 97: This very 5-types list exists in _two_ places in the code base now, so there is risk of the two going out of sync. Could this be (a) extracted to a non-public constant-like symbol somewhere or (b) commented about keeping in sync on both sides? (a) would likely be preferable. What do you think?

vcr/util.py
```diff
@@ -92,6 +92,21 @@ def composed(incoming):
...
+    ):
+        body = list(request.body)
+        if body:
+            if isinstance(body[0], str):
+                return "".join(body).encode("utf-8")
+            elif isinstance(body[0], bytes):
+                return b"".join(body)
+            elif isinstance(body[0], int):
+                return bytes(body)
+            else:
```
hartwork, line 107: I wonder if we should append `# pragma: no cover` to this line or add a test instead. What do you think?

vcr/util.py
```diff
@@ -92,6 +92,21 @@ def composed(incoming):
...
+                raise ValueError(f"Body type {type(body[0])} not supported")
+        return None
```
hartwork, line 109: - This line seems not to be covered by any test
- I wonder if returning `b""` or `bytes()` would be more adequate. What do you think?

vcr/util.py
```diff
@@ -92,6 +92,21 @@ def composed(incoming):
...
         return request.body.read()
+    if hasattr(request.body, "__iter__") and not isinstance(
+        request.body,
+        (bytearray, bytes, dict, list, str),
+    ):
+        body = list(request.body)
+        if body:
+            if isinstance(body[0], str):
+                return "".join(body).encode("utf-8")
+            elif isinstance(body[0], bytes):
```
hartwork, line 103:
```suggestion
            elif isinstance(body[0], (bytes, bytearray)):
```
?

tests/unit/test_stubs.py
hartwork, line 74: The last two `assert` statements could be dedented by one level. It would keep the context manager scope more minimal.

hartwork (changes requested):
@sathieu thanks for this new approach. Please see the details I found during review, below:

tests/unit/test_stubs.py
sathieu: I had to use `TemporaryDirectory` instead, because vcrpy fails when cassette is an empty file.

tests/unit/test_stubs.py
sathieu: Done.

tests/unit/test_stubs.py
sathieu: added a comment.

sathieu: done

tests/unit/test_util.py
sathieu: done.

tests/unit/test_util.py
sathieu: fixed, thanks (this was not failing on my computer).

vcr/util.py
sathieu: added a new `is_iterator` function. WDYT?

vcr/util.py
sathieu: done

vcr/util.py
sathieu: added a test.

vcr/util.py
sathieu: added a test, and returned `b""`

sathieu (commented):
Thanks @hartwork for your review. I've implemented your suggestions. Back to you!

tests/unit/test_util.py
hartwork: @sathieu it was working, yes, just not ideal. Thanks for the adjustments! :+1:

vcr/util.py
hartwork: @sathieu the function idea and the function body I like, but the name is a bit misleading and I'm worried that this de-facto internal helper function is becoming part of the public API this way. How about `_is_nonsequence_iterator`? The naming idea is based on [`collections.abc.Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) and it makes me realize as a side effect that we could make replace `isinstance(request.body, (bytearray, bytes, dict, list, str))` by `isinstance(request.body, Sequence)`. What do you think?

tests/unit/test_stubs.py
hartwork: @sathieu interesting! I got curious, played with it and can confirm that the file cannot exits already when we start. Here's what I did to make it work still:

```diff
# git --no-pager diff
diff --git a/tests/unit/test_stubs.py b/tests/unit/test_stubs.py
index c177061..8373923 100644
--- a/tests/unit/test_stubs.py
+++ b/tests/unit/test_stubs.py
@@ -1,7 +1,7 @@
 import contextlib
 import http.client as httplib
 from io import BytesIO
-from tempfile import TemporaryDirectory
+from tempfile import NamedTemporaryFile
 from unittest import mock
 
 from pytest import mark
@@ -52,8 +52,12 @@ class TestVCRConnection:
         data2,
         data3,
     ):
-        with TemporaryDirectory(dir=tmpdir) as cassete_dir:
-            testpath = str(cassete_dir.join("cass.yml"))
+        with NamedTemporaryFile(dir=tmpdir, suffix='.yml') as f:
+            testpath = f.name
+            # NOTE: ``use_cassette`` is not okay with the file existing
+            #       already.  So we using ``.close()`` to not only
+            #       close but also delete the empty file, before we start.
+            f.close()
             host, port = httpbin.host, httpbin.port
             match_on = ["method", "uri", "body"]
             with use_cassette(testpath, match_on=match_on):
```

I'm not saying that's better, maybe it's not :shrug: How about adding a sentence of comment why `TemporaryDirectory` is prefered over `NamedTemporaryFile` in our case? Something like that would maybe be best?

hartwork (changes requested):
@sathieu we're getting there, please see my new comments below for the remaining small semi-lose ends :beers:

tests/unit/test_stubs.py
sathieu: Yes, changed to your suggestion. thanks!

vcr/util.py
sathieu: Renamed to `_is_nonsequence_iterator`.

Unfortunately, `isinstance(request.body, Sequence)` is breaking a few tests...

sathieu:
@hartwork Sorry for the delay, the mail was lost in my inbox :wink: ...

Back to you, see my above comments :point_up:.

hartwork (approved):
@sathieu thanks for the adjustements! This looks ready-to-merge to me. Any concerns from @kevin1024 or @jairhenrique with regard to merging?

kevin1024:
The only contributions I have are (1) Thank you for your work, this looks great!  and (2) this PR title sounds like some kind of strange horror movie.

sathieu:
@hartwork When will this change be released?

NB: I currently a package removed from Debian testing because of it (see [bugs.debian.org#1073420](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1073420))

hartwork:
> @hartwork When will this change be released?

@sathieu I cannot do releases but @kevin1024 can.


> NB: I currently a package removed from Debian testing because of it (see [bugs.debian.org#1073420](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1073420))

From a quick look, I'm not sure it's the same issue. Is it?

sathieu:
@hartwork 

> From a quick look, I'm not sure it's the same issue. Is it?

Yes. There are two bugs :

- The consumed only once body🧟 (i.e. this merged PR)
- A change in urllib3 wordings (`Cannot connect to proxy.` ➡️ `Unable to connect to proxy`, see [aa73e725](https://gitlab.com/gitlabracadabra/gitlabracadabra/-/merge_requests/531/diffs?commit_id=aa73e72596e3a31317e544fcb6aa4b8ba3e42ebd))

PR #858, opened 2024-07-29
Fix compatibility with setuptools 72

ilpianista:
pytest [advises](https://github.com/pytest-dev/pytest/pull/5546) to not integrate with setuptools anymore, thus I've dropped that part.

Solves #855 #854

PR #856, opened 2024-07-29
Fix package install: remove use of deprecated setuptools.command.test

connortann:
Fixes #855 , fixes #854

At present the package install fails in typical conditions, as it is incompatible with the latest setuptools. The failure can be reproduced with:

```bash
pip wheel --no-cache-dir --use-pep517 "vcrpy (==6.0.1)"
```

This PR removes the used of deprecated setuptools modules, which has led to the package build breaking when installed in isolated environments.

Once this issue is fixed it would be useful to make a swift patch release, as it appears this issue is having quite a large affect on the wider python ecosystem:
https://github.com/pypa/setuptools/issues/4519

Issue #855, opened 2024-07-29
Fix compatibility with setuptools 72


strujilloc:
Due to [Test command removal](https://github.com/pypa/setuptools/issues/4519) in setuptools, installation fails with version 72:
```
Collecting vcrpy
  Using cached vcrpy-6.0.1.tar.gz (84 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "/private/var/folders/pb/0cllx_zx6vb88jht981qy96c0000gn/T/pip-build-env-c0ajc0p7/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 327, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
        File "/private/var/folders/pb/0cllx_zx6vb88jht981qy96c0000gn/T/pip-build-env-c0ajc0p7/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 297, in _get_build_requires
          self.run_setup()
        File "/private/var/folders/pb/0cllx_zx6vb88jht981qy96c0000gn/T/pip-build-env-c0ajc0p7/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 497, in run_setup
          super().run_setup(setup_script=setup_script)
        File "/private/var/folders/pb/0cllx_zx6vb88jht981qy96c0000gn/T/pip-build-env-c0ajc0p7/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 313, in run_setup
          exec(code, locals())
        File "<string>", line 9, in <module>
      ModuleNotFoundError: No module named 'setuptools.command.test'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.
```

Toeplitz:
what's the most correct fix for this currently?

agroszer:
Looks like GH actions just use `exec pytest` so is that piece of code still in use?

DanieleIsoni:
I think this is a duplicate of #854

agroszer:
> I think this is a duplicate of #854

It is, but title is misleading.

bitphage:
[...]

PR #859, opened 2024-07-29
fix: use poetry, not old setuptools

LewisCowlesMotive:
This aims to solve the setuptools problem of #858 #856 and #855 but by using a more modern tool

`poetry build` should, AFAIK build a universal wheel for you.

https://github.com/LewisCowlesMotive/vcrpy/releases/tag/v6.0.2 has notes on testing this fix.

I don't know how your CI works, so there may yet be more adjustments there.


graingert:
I think we should stick to setuptools for now. I'd merge a PR that switched to pyproject.toml based setuptools

pjonsson:
@graingert is that what #767 provides, or did you have something else in mind?

Lewiscowles1986:
I'll close today. Not interested in explicit dependency on setuptools

pyproject.toml
```diff
@@ -1,3 +1,66 @@
...
+[tool.poetry.dependencies]
+python = ">=3.8"
+pyyaml = "^6.0.1"
+wrapt = "^1.16.0"
+yarl = "^1.9.4"
+urllib3 = [
+    { version = "<2", markers = "python_version < '3.10'" },
+    { version = "<2", markers = "platform_python_implementation == 'PyPy'" },
+    { version = ">=2.2.2", markers = "python_version >= '3.10' and platform_python_implementation != 'PyPy'" }
+]
```
LewisCowlesMotive, line 39: after thought. This is incorrect. This package has no dependencies as it is, itself a dev dependency.

Issue #848, opened 2024-06-28
Test regressions with Python 3.13.0b2


mgorny:
Compared to Python 3.12, I'm seeing a few new failing tests:

```pytb
$ EPYTHON=python3.13 bash runtests.sh --deselect tests/integration/test_urllib2.py::test_multiple_requests --deselect tests/integration/test_urllib2.py::test_random_body tests/integration/test_urllib2.py 
/usr/lib/python3.13/site-packages/httpbin/helpers.py:32: SyntaxWarning: invalid escape sequence '\_'
  ASCII_ART = """
/usr/lib/python3.13/site-packages/httpbin/helpers.py:77: SyntaxWarning: invalid escape sequence '\ '
  ANGRY_ASCII ="""
/usr/lib/python3.13/site-packages/httpbin/helpers.py:444: SyntaxWarning: invalid escape sequence '\s'
  match = re.search('\s*(W/)?\"?([^"]*)\"?\s*', part)
[2024-06-28 20:55:41,591] WARNING in core: Swagger not found, legacy index will be used.
========================================================= test session starts =========================================================
platform linux -- Python 3.13.0b3, pytest-8.2.2, pluggy-1.5.0
rootdir: /tmp/vcrpy
configfile: pyproject.toml
plugins: xdist-3.6.1, subtests-0.12.1, asyncio-0.23.7, xprocess-1.0.2, anyio-4.4.0, pkgcore-0.12.27, hypothesis-6.103.5, timeout-2.3.1, httpbin-2.0.0
asyncio: mode=Mode.STRICT
collected 18 items / 4 deselected / 14 selected                                                                                       

tests/integration/test_urllib2.py ......pytest-httpbin server hit an exception serving request: [SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] ssl/tls alert certificate unknown (_ssl.c:1024)
attempting to ignore so the rest of the tests can run
Fpytest-httpbin server hit an exception serving request: [SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] ssl/tls alert certificate unknown (_ssl.c:1024)
attempting to ignore so the rest of the tests can run
Fpytest-httpbin server hit an exception serving request: [SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] ssl/tls alert certificate unknown (_ssl.c:1024)
attempting to ignore so the rest of the tests can run
[...]

PR #862, opened 2024-07-30
Start testing with CPython 3.13

graingert:
investigate  #848


.github/workflows/main.yml
```diff
@@ -48,11 +49,12 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
+          allow-prereleases: true
 
       - name: Install project dependencies
         run: |
           pip install --upgrade pip setuptools
-          pip install codecov '.[tests]' '${{ matrix.urllib3-requirement }}'
+          pip install codecov 'cffi>=1.17.0rc1' 'pytest-httpbin @ https://github.com/kevin1024/pytest-httpbin/archive/refs/heads/test-on-3-13-and-pypy-3-10.zip' 'httpbin @ https://github.com/graingert/httpbin/archive/refs/heads/patch-1.zip' '.[tests]' '${{ matrix.urllib3-requirement }}'
```
graingert, line 57:
```suggestion
          pip install 'pytest-httpbin @ https://github.com/kevin1024/pytest-httpbin/archive/refs/heads/test-on-3-13-and-pypy-3-10.zip' 'httpbin @ https://github.com/graingert/httpbin/archive/refs/heads/patch-1.zip' '.[tests]' '${{ matrix.urllib3-requirement }}'
```

.github/workflows/main.yml
```diff
@@ -48,11 +49,12 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
+          allow-prereleases: true
 
       - name: Install project dependencies
         run: |
           pip install --upgrade pip setuptools
-          pip install codecov '.[tests]' '${{ matrix.urllib3-requirement }}'
+          pip install 'pytest-httpbin @ https://github.com/kevin1024/pytest-httpbin/archive/refs/heads/test-on-3-13-and-pypy-3-10.zip' 'httpbin @ https://github.com/graingert/httpbin/archive/refs/heads/patch-1.zip' '.[tests]' '${{ matrix.urllib3-requirement }}'
```
graingert, line 57:
```suggestion
          pip install codecov 'pytest-httpbin @ https://github.com/kevin1024/pytest-httpbin/archive/refs/heads/test-on-3-13-and-pypy-3-10.zip' 'httpbin @ https://github.com/graingert/httpbin/archive/refs/heads/patch-1.zip' '.[tests]' '${{ matrix.urllib3-requirement }}'
```

.github/workflows/main.yml
```diff
@@ -48,11 +49,12 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
+          allow-prereleases: true
 
       - name: Install project dependencies
         run: |
           pip install --upgrade pip setuptools
-          pip install codecov '.[tests]' '${{ matrix.urllib3-requirement }}'
+          pip install codecov 'pytest-httpbin @ https://github.com/kevin1024/pytest-httpbin/archive/refs/heads/test-on-3-13-and-pypy-3-10.zip' 'httpbin @ https://github.com/graingert/httpbin/archive/refs/heads/patch-1.zip' '.[tests]' '${{ matrix.urllib3-requirement }}'
```
graingert, line 57:
```suggestion
          pip install codecov 'pytest-httpbin>=2.1.0rc1' '.[tests]' '${{ matrix.urllib3-requirement }}'
```

.github/workflows/main.yml
```diff
@@ -48,11 +49,12 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
+          allow-prereleases: true
 
       - name: Install project dependencies
         run: |
           pip install --upgrade pip setuptools
-          pip install codecov '.[tests]' '${{ matrix.urllib3-requirement }}'
+          pip install codecov 'pytest-httpbin>=2.1.0rc1' '.[tests]' '${{ matrix.urllib3-requirement }}'
```
graingert, line 57:
```suggestion
          pip install codecov '.[tests]' '${{ matrix.urllib3-requirement }}'
```

hartwork (changes requested):
@graingert please rebase/resquash to ease review in a way where…

- there a no more reverting commits canceling out prior commits (— let's get addition and revert out of history instead —)
- there are no merge commits in here
- the branch targets latest `master`

…to ease review — thank you!

hartwork (changes requested):
@graingert thanks for the adjustments, much cleaner now. The use of `pytest_httpbin.certs.where` is a nice idea but the way it's introduced is probably more complex than necessary (and the tight coupling with Python 3.13 support is not clear); also no longer calling `load_cert_chain` could use an explanation why it's expected to not be needed for all environments (I'm curious, happy to understand).

The key question should be: what problem are we trying to solve? The answer seem to be: Invoking the test suite without need for `REQUESTS_CA_BUNDLE="$(python3 -m pytest_httpbin.certs)"` hacks and without default failure — anything else?

Probably the simplest and backwords-compatible way forward would be a tiny patch that…
- uses `ssl_ca_location = os.environ.get("REQUESTS_CA_BUNDLE", pytest_httpbin.certs.where())`
- starts using `ssl_ca_location` in two places more in the same function
- then drop `REQUESTS_CA_BUNDLE` from `runtests.sh`

If we need any of the other parts of the suggest diff, please share their motivation so that it becomes apparant and documented why we need them. Thank you!

graingert:
> @graingert thanks for the adjustments, much cleaner now. The use of `pytest_httpbin.certs.where` is a nice idea but the way it's introduced is probably more complex than necessary (and the tight coupling with Python 3.13 support is not clear); also no longer calling `load_cert_chain` could use an explanation why it's expected to not be needed for all environments (I'm curious, happy to understand).

It's coupled to 3.13 because python 3.13 changed the validation requirements for certificates in the default SSL context, and I updated the pytest-httpbin certs using trustme which names them differently. pytest_httpbin.certs.where() is the public interface to these files, they should not be accessed by filename

load_cert_chain is needed for server contexts and is redundant here
> 
> The key question should be: what problem are we trying to solve? The answer seem to be: Invoking the test suite without need for `REQUESTS_CA_BUNDLE="$(python3 -m pytest_httpbin.certs)"` hacks and without default failure — anything else?
No, it's because I renamed the certs in pytest-httpbin because they are the names that trustme uses, and otherwise the tests crash here

> Probably the simplest and backwords-compatible way forward would be a tiny patch that…
> 
>     * uses `ssl_ca_location = os.environ.get("REQUESTS_CA_BUNDLE", pytest_httpbin.certs.where())`
> 
>     * starts using `ssl_ca_location` in two places more in the same function
> 
>     * then drop `REQUESTS_CA_BUNDLE` from `runtests.sh`
Backwards compatibility is not a concern here, it's just in the tests
REQUESTS_CA_BUNDLE is still required in runtests.sh
> 
> If we need any of the other parts of the suggest diff, please share their motivation so that it becomes apparant and documented why we need them. Thank you!

I'm not sure what other parts of the diff you're referring to

hartwork:
> > @graingert thanks for the adjustments, much cleaner now. The use of `pytest_httpbin.certs.where` is a nice idea but the way it's introduced is probably more complex than necessary (and the tight coupling with Python 3.13 support is not clear); also no longer calling `load_cert_chain` could use an explanation why it's expected to not be needed for all environments (I'm curious, happy to understand).
> 
> It's coupled to 3.13 because python 3.13 changed the validation requirements for certificates in the default SSL context, and I updated the pytest-httpbin certs using trustme which names them differently. pytest_httpbin.certs.where() is the public interface to these files, they should not be accessed by filename.
> 
> load_cert_chain is needed for server contexts and is redundant here.

I found your commit https://github.com/kevin1024/pytest-httpbin/commit/7bf62b4e8848ce0990cb8d09b8b630dc4140f41b now, so the VCR.py tests will break with `pytest-httpbin>=v2.1.0rc1` unless patched, I see.

> I'm not sure what other parts of the diff you're referring to.

The diff in 58ef14708dfaf74bed9a734bf9176bc31125f60d is quite big, e.g. `httpbin_ssl_context` is also resolved rather than fixing it in place (which would give a tiny diff). If you see value in inlining it after fixing, making that two commits would make things clear. Right now none of the parts are separated and non of the intentions/reasoning are documented. I can demo how I'd do it, if you want.

hartwork (changes requested):
(Formally requesting changes for https://github.com/kevin1024/vcrpy/pull/862#issuecomment-2359267315 above)

hartwork (approved):
@graingert that's a very nice story told by commits now — thank you! :+1:

PR #906, opened 2025-01-22
fix: usage of io-like interface with VCR.py

tito:
This pull request aims to resolve an issue with using VCR.py in conjunction with io-like data.
Let's have a look at this example:

```python
@vcr.use_cassette("issue_vcr_aiohttp.yaml")
def test():
    test = io.BytesIO(b"hello")
        async with session.post("https://httpbin.org/post", data=test, json=None) as response:
            ret = await response.json()
            assert ret["data"] == "hello"
```

It will trigger the assert, because the body received by aiohttp will be empty.
When using cassette and building Request, the [body is read()](https://github.com/kevin1024/vcrpy/blob/master/vcr/request.py#L22) and later usage of body by aiohttp won't work.

### DIFF
--- a/tests/integration/test_aiohttp.py
+++ b/tests/integration/test_aiohttp.py
@@ -1,3 +1,4 @@
+import io
 import logging
 import ssl
 import urllib.parse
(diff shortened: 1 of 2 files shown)


tests/integration/test_aiohttp.py
```diff
@@ -466,3 +467,20 @@ def test_filter_query_parameters(tmpdir, httpbin):
         cassette_content = f.read()
         assert "password" not in cassette_content
         assert "secret" not in cassette_content
+
+
+@pytest.mark.online
+@pytest.mark.asyncio
```
jairhenrique, line 473:
```suggestion

```

tito: Why is the mark asyncio is not required ?

tito: Sorry i didn't saw the tests run, thanks for the suggestion.

jairhenrique (approved):

Issue #628, opened 2022-01-07
Compatibility with starlette / fastapi TestClient


johnraz:
Hi!

I've been facing some issues with a test suite that makes use of [Starlette's TestClient](https://www.starlette.io/testclient/), which is also used by [FastAPI](https://fastapi.tiangolo.com/).

Important note, I'm using pytest with [pytest-recording](https://github.com/kiwicom/pytest-recording).

There are 2 types of issues I see, and both are random and not easily reproducible:

1. cassettes are sometimes not recorded when running the full test suite but are when running the test alone.
2. tests are failing when running test2, and trying to access the cassette in test2, while test2 is not even using `use_cassette`... This mostly happens in parametrized tests but I'm not sure this is the "key" reason.

I think the problem mostly comes from the fact that the TestClient starts an async loop behind the scene which seems to hit the thread safety problem described here: https://github.com/kevin1024/vcrpy/issues/295, do you think it is indeed related?

Thanks for you time and for the otherwise very useful project 😃

garbusbeach:
Hi, I've hit similar problem. Did you succeed to resolve this maybe?

johnraz:
No sadly not - it makes vcrpy hard to use with Starlette test client.

mawkee:
I just hit the same problem. I'm convinced this is a problem with pytest-recording

chris-erickson:
Seems to "work" for us, but tests that use the client (rather than testing some more direct python code) have this issue: 
```No match for the request (<Request (POST) http://testserver/orgs>) was found.```

Was hoping this would suffice, but nope:
```"ignore_hosts": ["testserver"],```

PR #909, opened 2025-01-27
fix(httpx): don't save response if filtered to None

alekseik1:
VCR py fails to record response made by starlette (and FastAPI as well) `TestClient` #628.
It seems like the problem hides [in this assert](https://github.com/alekseik1/vcrpy/blob/master/vcr/stubs/httpx_stubs.py#L41) that is triggered by [this func call](https://github.com/alekseik1/vcrpy/blob/master/vcr/stubs/httpx_stubs.py#L139).

I took a quick look and it looks like we're trying to record response to `cassete` object here.
But at the same time, [here](https://github.com/alekseik1/vcrpy/blob/master/vcr/stubs/httpx_stubs.py#L119) we see that whenever filters render `response` to None, a real request is always performed. So, why should we record real response if we don't use it anyway? I made a small check with `should_save_response` variable that runs same filters and skips recording if it will never be replayed.

There's one catch though. Image following scenario:
1. User sets `ignore_hosts=["some.real.host.org"]`, runs tests and gets his cassettes.
2. Neither of cassettes contains requests to "some.real.host.org" since we skipped writing them (by `should_save_response` var).
3. Now, user decides to set `ignore_hosts=[]` and rerun his tests.
4. His cassettes are no longer valid because of a missing request to "some.real.host.org". Such behavior is introduced by this PR.

Does it look sensible?

### DIFF
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -136,9 +136,7 @@ async def _record_responses(cassette, vcr_request, real_response, aread):
         # final redirect value
         vcr_request = _make_vcr_request(real_response.request)
 
-    should_save_response = cassette._before_record_request(vcr_request) is not None
-    if should_save_response:
-        cassette.append(vcr_request, await _to_serialized_response(real_response, aread))
+    cassette.append(vcr_request, await _to_serialized_response(real_response, aread))
     return real_response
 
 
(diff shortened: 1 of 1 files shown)


alekseik1:
damn, this `anyio` stuff is really hard to debug. By trials and errors, I found out that `run_coroutine_threadsafe` seem to fix problems with `TestClient`. My idea is that starlette's `TestClient` might lead to race conditions when reading response because `TestClient` uses threads internally (see [here](https://github.com/encode/starlette/blob/4d72fd87ea1c358691a19468168f7217d8ca6a87/starlette/testclient.py#L656)).

Unfortunately, any attempt to reproduce bug in test environment failed - all tests were green without my fix applied. I believe that problems with `TestClient` start on large codebase, when FastAPI has lots of Tasks/Futures running in handles.

PR #918, opened 2025-04-11
Remove urllib3 dependency

clavedeluna:
I'm not sure if this is possible at this time, but I'm working with the `uv`  package manager and having so many problems trying to install `vcrpy==7.0.0` with other packages needing `urllib3>2`. Removing this dependency here fixed all my issues.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,18 +30,6 @@ install_requires = [
     "PyYAML",
     "wrapt",
     "yarl",
-    # Support for urllib3 >=2 needs CPython >=3.10
-    # so we need to block urllib3 >=2 for Python <3.10 and PyPy for now.
-    # Note that vcrpy would work fine without any urllib3 around,
-    # so this block and the dependency can be dropped at some point
-    # in the future. For more Details:
-    # https://github.com/kevin1024/vcrpy/pull/699#issuecomment-1551439663
-    "urllib3 <2; python_version <'3.10'",
-    # https://github.com/kevin1024/vcrpy/pull/775#issuecomment-1847849962
-    "urllib3 <2; platform_python_implementation =='PyPy'",
-    # Workaround for Poetry with CPython >= 3.10, problem description at:
-    # https://github.com/kevin1024/vcrpy/pull/826
-    "urllib3; platform_python_implementation !='PyPy' and python_version >='3.10'",
 ]
 
 extras_require = {


adamJLev:
Agreed,  not sure what the best fix is here but im having some pain now with this too

```
$ poetry env info

Virtualenv
Python:         3.13.3
Implementation: CPython
Path:           /Users/ajl/Library/Caches/pypoetry/virtualenvs/project-bZWYOXtw-py3.13
Executable:     /Users/ajl/Library/Caches/pypoetry/virtualenvs/project-bZWYOXtw-py3.13/bin/python
Valid:          True

$ poetry lock  && poetry install --with=dev
Resolving dependencies... (0.2s)

Because vcrpy (7.0.0) depends on urllib3 (<2)
 and project depends on urllib3 (>2.0.0), vcrpy is forbidden.
So, because project depends on vcrpy (7.0.0), version solving failed.
```

Abello966:
signal-boosting this because there is an CVE relating to urllib<2.5.0

https://nvd.nist.gov/vuln/detail/CVE-2025-50182

deeleeramone:
Bumping for visibility here. Requirement pinning urllib3 < 2 is blocking lots of security and package updates - and is quite unnecessary. 

>  "urllib3 <2; platform_python_implementation =='PyPy'",

clavedeluna:
bump, could we get this merged and released? 🙏

CharString:
> signal-boosting this because there is a CVE relating to urllib<2.5.0
> 
> https://nvd.nist.gov/vuln/detail/CVE-2025-50182

That CVE is regarding 2.2.0 <= urllib3 < 2.5.0. There is nothing in these pins that pins anyone to a vulnerable version.

And that bug in poetry dependency resolution ought to be solved in poetry not in the rest of the whole ecosystem.

piiq:
For anyone looking for a solution:


1. On the `poetry` side: you can include the markers into the dependency definition in the `pyproject.toml` to prevent old `urllib3` version leaking into the lock file or the environment. For example:

```toml
vcrpy = { version = ">=7", markers = "platform_python_implementation == 'CPython'" }
```

3. On the `uv` side: you can use the dependency override feature by specifying the platform and supported python version for a specific dependency. Here's a minimalistic `pyproject.toml` that will get you a `uv.lock` that doesn't mention the old version of `urllib3`

```toml
[project]
name = "temp"
version = "0.1.0"
authors = [{ name = "piiq", email = "noreply@github.com" }]
requires-python = ">=3.12"
dependencies = ["vcrpy>=7.0.0"]


[tool.uv]
override-dependencies = [
    "urllib3>=2 ; implementation_name == 'cpython' and python_version >= '3.12'",
]
```

P.S. If this works for you feel free to link to this comment elsewhere.

PR #923, opened 2025-05-04
fix: don't record message if filtered

### LINKED ISSUE #628: Compatibility with starlette / fastapi TestClient

alekseik1:
From `docs/advanced.rst`:
> Requests that are ignored by VCR will not be saved in a cassette, nor played back from a cassette. VCR will completely ignore those requests as if it didn't notice them at all, and they will continue to hit the server as if VCR were not there.

Which means it is safe to completely ignore these requests when recording responses. 

Why this matters? Because `assert not hasattr(resp, "_decoder")` fails in `_to_serialized_response` func, which breaks `vcr` for Starlette TestClient, see #628.

Fixes #628.

### DIFF
--- a/vcr/stubs/httpx_stubs.py
+++ b/vcr/stubs/httpx_stubs.py
@@ -127,6 +127,8 @@ def _shared_vcr_send(cassette, real_send, *args, **kwargs):
 
 
 async def _record_responses(cassette, vcr_request, real_response, aread):
+    if not cassette.filter_request(vcr_request):
+        return real_response
     for past_real_response in real_response.history:
         past_vcr_request = _make_vcr_request(past_real_response.request)
         cassette.append(past_vcr_request, await _to_serialized_response(past_real_response, aread))


alekseik1:
an alternative approach would be to change `_run_async_function` to
```python
def _run_async_function(sync_func, *args, **kwargs):
    """
    Safely run an asynchronous function from a synchronous context.
    Handles both cases:
    - An event loop is already running.
    - No event loop exists yet.
    """
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        return asyncio.run(sync_func(*args, **kwargs))
    else:
        # If inside a running loop, create a task and wait for it
        return loop.run_in_executor(None, partial(sync_func, *args, **kwargs))  # <<< see here
```

Because right now it is bugged - we run a task in asyncio but never wait for it to finish.

@jairhenrique what do you think?

alekseik1:
you know what. I think these asyncio-hacks are bad things to support. It's better to duplicate code than to manage event loops.

alekseik1:
any updates on this? @kevin1024 @jairhenrique

Issue #936, opened 2025-09-02
Add recursive replace_post_data_parameters fields mask for json body


JargeZ:
This currently only supports the first level of keys
However, it really begs to be recursive.
https://github.com/kevin1024/vcrpy/blob/d50f3385a6828280def801ac7f544fe04a37e39c/vcr/filters.py#L100
I'm leaving it just in case, I don't have time to make a pull request myself yet, but it's possible in the future.

Put your upvotes or concerns!

JargeZ:
@tighwm is on it!

PR #937, opened 2025-09-03
Feature recusive filter request body

tighwm:
Added recursive filtering of request body on this [issue](https://github.com/kevin1024/vcrpy/issues/936) and unit test on this case

### DIFF
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -212,7 +212,7 @@ def test_replace_recursive_post_data_parameters():
             ("key", None),
             ("nested", "aboba"),
         ],
-        True
+        recursive=True
     )
     request_data = json.loads(request.body)
     expected_data = json.loads('{"nested": "aboba", "one": {"nested": "aboba"}}')
--- a/vcr/filters.py
+++ b/vcr/filters.py
@@ -70,7 +70,7 @@ def remove_query_parameters(request, query_parameters_to_remove):
 
 
 def filtering_body(request, body_data, replacements):
-    """Filtering the request body by default to only high level keys"""
+    """Filtering the request body by default to only top level keys"""
     for k, rv in replacements.items():
         if k in body_data:
             ov = body_data.pop(k)


mahenzon (approved):

tighwm:
@vEpiphyte Hi! I understand your concerns. 
Turning this function into a utility seems like a more flexible solution. 
On the other hand, I’d suggest a simpler alternative: using the `recursion_filter` flag, which defaults to `False`.

JargeZ:
@vEpiphyte thanks for the quick comment and opinion.  
I see that @tighwm has already added this as a separate filter.

I’d like to finally propose thinking through the desired usage API. Since I’d prefer not to make the configuration of tools too complicated with lots of different settings.  
Adding an entirely new parameter just for the recursive variation seems redundant to me.  
We could either extend the mechanism for advanced usage of the current replace configuration, or introduce a completely new, more independent configuration parameter.

## Suggested solutions

### Modify existing semantic

Facts for risk assessment:
- does not affect existing cassettes
- affects recording of new cassettes
- potentially unexpected behavior if a user previously had `filter_post_data_parameters` with a key that appears not only at the top level of a request
- AND at the same time the user really expect this field to be included, necessarily expects this field in the request to be taken into account in the request logic
- The standard behavior is matching on the `uri`

So the question is: what percentage of users simultaneously have both a configured `filter_post_data_parameters` and body matching and/or expect request to be recorded with a repeating key that are filtered out on the root level?

In other words, I want to assess how real this risk is - whether we must preserve full backward compatibility, or if this is an unnecessary safeguard that will just complicate the library’s API.


### Modify current functionality with a backward-compatibility helper

Assuming that recursive search and sanitization of this key is more intuitive and convenient behavior (based **only on my personal experience and assumption**)

It would be possible to change the default behavior to recursive and update the documentation, noting that if you want only top-level keys, you should now wrap the key in a helper, for example:

```python
@pytest.fixture(scope="module")
def vcr_config():
    return {
        "decode_compressed_response": True,
        ...
        "filter_post_data_parameters": [
            "api_key",
            vcr.RootKey("password"),
        ],
    }
```

#### Variation of existing functionality with backward compatibility

We could do something similar by leaving full backward compatibility by default.  
For example, introducing a helper for the opposite situation:  
`vcr.RecursiveKey("password")`

This option is safe in terms of compatibility, but makes the behavior less explicit, which, **in my assumption**, is actually what users expect by default.

### New configuration parameter

Based on all this, we could also introduce new sanitization functionality altogether.  
The goal here is to have the library proactively handle a problem that users are very likely to encounter.

For example, we could add a new API usage pattern like:

```python
with my_vcr.use_cassette('test.yml', sensetive=True):
    # sensitive HTTP request goes here
```

In this simple usage, it would automatically search for and remove potentially sensitive keys (headers, query params, body).  
The list of built-in parameters could be discussed, but covering most default use cases would only require popular ones.

And for advanced usage, the list of sensitive fields could be user-defined:

```python
with my_vcr.use_cassette('test.yml', sensetive=True, sensetive_fields=['password']):
    # sensitive HTTP request goes here
```
(Specifying `sensetive_fields` explicitly could make it unnecessary to also set `sensetive=True`).

That way, the my original use case, where you want to apply guards to all cassettes, would look like this:
```python
@pytest.fixture(scope="module")
def vcr_config():
    return {
        "decode_compressed_response": True,
        "sensetive": True,
        "sensetive_fields": [
            "authorisation",
            "access_token",
            "token",
            "api_key",
            "password",
        ],
    }
```

PR #957, opened 2025-12-04
chore: fix async execution when running VCR inside event loop

pappacena:
When running VCR inside an already existing event loop, running `asyncio.ensure_future(sync_func(*args, **kwargs))` is not enough to actually execute `sync_func`.

Since the return value is important in this case, there is no straightforward way to call the async method from a sync one that is already running inside the event loop (more or less [this scenario](https://stackoverflow.com/a/78826778)). 

So, this  PR creates a new event loop to run the method, and waits for its result.

In my local tests, this patch seems to solve the issue, but feel free to edit as needed (or point me to anything that needs to be done to merge this).

### DIFF
--- a/vcr/stubs/httpcore_stubs.py
+++ b/vcr/stubs/httpcore_stubs.py
@@ -190,7 +190,6 @@ def _run_async_function(sync_func, *args, **kwargs):
     else:
         # If inside a running loop, run the task in a separated
         # event loop in a new thread.
-        print("Yep, inside loop")
         result = None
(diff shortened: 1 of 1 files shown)


kevin1024:
Closing this PR as v8.0.0 includes a complete rewrite of httpx support (now patching httpcore instead of httpx). This architectural change should address the async execution issues. If you're still experiencing problems with v8.0.0, please open a new issue and we can revisit. Thanks for the contribution\!

PR #965, opened 2025-12-11
Split sync and async HTTPX handlers

seowalex:
In https://github.com/kevin1024/vcrpy/pull/943#discussion_r2446643448, sending sync requests in async contexts was accidentally broken. As #957 and #958 correctly point out, this is because the future's result was not awaited.

Instead of creating a separate event loop as #957 and #958 do, I have opted to properly separate the sync and async HTTPX handlers, as the only reason async functions were being run in a sync context was to reuse the functions.

To do this, I have pushed the logic to convert the byte stream (which is the only truly async function) up into the respective sync/async handlers, which allows the other functions to be sync (and thus callable from the respective handlers without additional work).

I have also added a test for this regression and removed the extraneous `pytest.mark.online` markers.


seowalex:
Tagging @jairhenrique given you reviewed my previous PR, I think this is a fairly important bugfix.

dsfaccini:
looking forward to this

seowalex:
@kevin1024 @hartwork @jairhenrique do you have time to take a look at this?

kevin1024:
I have to admit I’m not familiar with the httpx internals, but tests are passing.  Unless there are objections from the other maintainers I’ll merge.

Happy New Year by the way! And thanks for your contribution.

seowalex:
Yes, I think merging and releasing this soon would be best since this fixes a bug (caused by me 🥲) in the new v8 release which cause a subset of HTTPX requests to fail when using vcrpy.

seowalex:
@kevin1024 any objections to merging and releasing this? There has been another issue created regarding this bug.

kevin1024:
This has been released in v8.1.1. Thanks for the fix @seowalex!

Issue #968, opened 2025-12-14
Support for intercepting httpx custom transports (BaseTransport) such as httpx-curl-cffi?


epicwhale:
I use vcrpy to record/replay httpx requests for offline unit tests. This works when httpx uses the default stack, but fails when httpx.Client/AsyncClient is configured with a custom transport  (e.g https://github.com/vgavro/httpx-curl-cffi)

It looks like VCR patches httpcore, so custom transports bypass interception?

Is there an intended/official way to support recording/replay for such custom httpx.BaseTransport implementations?

seowalex:
I'm surprised custom transports do not work given #943 was supposed to fix exactly this case (and works for Hishel, which uses a custom caching transport). Patching at the `httpcore` level means that any higher-level transports should continue to work with the patched `httpcore`.

What error are you facing?

epicwhale:
> I'm surprised custom transports do not work given [#943](https://github.com/kevin1024/vcrpy/pull/943) was supposed to fix exactly this case (and works for Hishel, which uses a custom caching transport). Patching at the `httpcore` level means that any higher-level transports should continue to work with the patched `httpcore`.
> 
> What error are you facing?

It just wasn't recording from my try outs. When I dug deeper, these were my findings I had written down - 

```
    With default httpx transport:
        httpx.AsyncClient
            → httpx.AsyncHTTPTransport
            → httpcore.AsyncConnectionPool.handle_async_request  ← VCR patches HERE
            → network (via httpcore/HTTP libs)

    With curl_cffi transport:
        httpx.AsyncClient
            → httpx_curl_cffi.AsyncCurlTransport.handle_async_request  ← VCR MISSES this!
            → curl_cffi.requests.AsyncSession
            → network (via libcurl)
```

Perhaps it's because not all httpx transport layers necessarily use the httpcore level?

Greesb:
We have a nearly similar issue, we use a `httpx.ASGITransport` and vcrpy seems to bypass the request done with an ASGITransport, causing them to not be recorded, or replayed if using a cassette from vcrpy < 8.0.

seowalex:
Ah, I hadn't thought about that. Just curious though, previously when it was patched at the `httpx` level, (if I'm not wrong) the custom transports would still have been bypassed, but that was ok for your use cases as long as the request was recorded?

Greesb:
[...]

PR #972, opened 2026-01-10
Patch httpx transports instead of httpcore

seowalex:
While #943 brings a lot of advantages by moving the patch target to a lower level, it inadvertently broke the use of custom transports that do not wrap the default transports (see #968). To ensure that both entirely custom transports and custom transports that wrap the default transports work with VCR.py, we change the patch target to the transport level instead.

This necessitates wrapping each transport that HTTPX supports, such as `HTTPTransport`, `AyncHTTPTransport`, `WSGITransport`, `ASGITransport` and `MockTransport`. I have also added support for `CurlTransport` and `AsyncCurlTransport` from https://github.com/vgavro/httpx-curl-cffi, and `SyncHttpxTransport` and `HttpxTransport` from https://github.com/MarkusSintonen/pyreqwest which seem to be the only 2 popular alternative transports I can find on GitHub.

This also fixes a bug found in #977 causing the connection pool to be exhausted due to not properly closing the original response stream.


seowalex:
@kevin1024 @hartwork @jairhenrique apologies for tagging, but do you have any feedback for this?

seowalex:
Hi, any interest in merging this?

Issue #990, opened 2026-05-17
Support httpx2 + httpcore2


hugovk:
httpx2 and httpcore2 are forks of httpx and httpcore:

* https://github.com/pydantic/httpx2
* https://github.com/pydantic/httpx2/tree/main/src/httpx2
* https://github.com/pydantic/httpx2/tree/main/src/httpcore2

It would be useful to be able to test packages with VCR.py that switch from httpx to httpx2.

dsfaccini:
David's AICA here: hitting this from the pydantic-ai side in https://github.com/pydantic/pydantic-ai/pull/5503 (migrating internal HTTP calls from `httpx` to `httpx2`).

The silent-bypass failure mode is real and concrete — `tests/models/test_download_item.py::test_download_item_no_content_type` was passing on `httpx` (cassette has no `content-type`, asserts `data_type == 'text/markdown'` via the URL-extension fallback) and started failing after the swap because the request bypassed the cassette and live raw.githubusercontent.com returns `text/plain` for `.md` files. Nothing in the test output suggests "VCR didn't intercept" — it just looks like the assertion drifted.

We worked around it locally with a small monkey-patch on `vcr.patch.CassettePatcherBuilder.build` ([pydantic-ai shim](https://github.com/dsfaccini/pydantic-ai/blob/use-httpx2/tests/_vcr_httpcore2_compat.py)). The shim is essentially `httpcore_stubs.py` with imports re-pointed at `httpcore2` — the model API (`Response`, `Request`, `ByteStream`) is intentionally compatible at the transport boundary, so the diff is mechanical.

Happy to upstream a proper `vcr/stubs/httpcore2_stubs.py` PR if you'd accept the change. Two questions before we open it:

1. Optional-import gating — same pattern as `try: import httpcore` in `vcr/patch.py:97`, just adding a parallel `try: import httpcore2` block?
2. Test coverage — is the existing `tests/integration/test_httpx.py` pattern sufficient as a template (i.e. add a parallel `test_httpx2.py`), or would you want a different shape?

(Tagging @hugovk who filed this.)

kevin1024:
[.](https://github.com/kevin1024/vcrpy/issues/990#issuecomment-4566646428)

> Happy to upstream a proper vcr/stubs/httpcore2_stubs.py PR if you'd accept the change. Two questions before we open it:
>
[...]

PR #993, opened 2026-05-29
Add httpx2 support

dsfaccini:
David's AICA here:

Closes the silent-bypass failure from #990: code using `httpx2` (a fork of `httpx`) bypasses the cassette entirely and hits the live server, with no signal that VCR didn't intercept.

**Rebased onto the new transport-patching design (#972).** Since #972 landed, vcrpy patches **httpx transports** directly (`HTTPTransport`, `AsyncHTTPTransport`, `WSGITransport`, `ASGITransport`, `MockTransport`) rather than httpcore. This PR has been re-targeted accordingly: it now adds `httpx2` support at the same transport layer (the earlier httpcore2 approach is gone — vcrpy no longer touches httpcore).

**No duplicated stub.** Per @seowalex's / @kevin1024's feedback on #990, the two libs share a single stub. `vcr/stubs/httpx_stubs.py` is parameterized by the httpx module to patch against (`httpx` or `httpx2`) — only `Response` / `ByteStream` and the `request_context` helper differ; everything else operates on the duck-typed request/response.

**Changes:**
1. `vcr/stubs/httpx_stubs.py` — drop the top-level `from httpx import ...`; take the httpx module as a parameter so it loads when only `httpx2` is installed and serves both.
2. `vcr/patch.py` — `try: import httpx2` save block; a `_httpx2` patcher mirroring `_httpx`'s core transports (registered in `build()`); a matching `reset_patchers()` block. `_httpx` now passes the `httpx` module into the shared stub (httpx core + `httpx_curl_cffi` + `pyreqwest`, all of which produce httpx objects).
3. `pyproject.toml` / `setup.cfg` — add `httpx2` to the `tests` extra.
4. `tests/integration/test_httpx2.py` — mirrors `test_httpx.py` with `httpx` → `httpx2` (plus a `gzip_httpx2_old_format.yaml` cassette). The `httpx_curl_cffi` / `pyreqwest` alt-transport cases are omitted — those belong to the httpx ecosystem and aren't compatible with `httpx2.Client`; `MockTransport` coverage is kept.
5. `docs/installation.rst` — list `httpx2` as supported.

`pytest tests/integration/test_httpx.py tests/integration/test_httpx2.py` → 72 passed locally; full unit suite (181) green.

### DIFF
--- a/tests/integration/test_httpx2.py
+++ b/tests/integration/test_httpx2.py
@@ -1,0 +1,403 @@
+import os
+
+import pytest
+
+import vcr
+
+from ..assertions import assert_is_json_bytes
+
+asyncio = pytest.importorskip("asyncio")
+httpx2 = pytest.importorskip("httpx2")
+
+
+@pytest.fixture(params=["https", "http"])
+def scheme(request):
+    """Fixture that returns both http and https."""
+    return request.param
+
+
+class BaseDoRequest:
+    _client_class = None
+
+    def __init__(self, *args, **kwargs):
+        self._client_args = args
+        self._client_kwargs = kwargs
+        self._client_kwargs["follow_redirects"] = self._client_kwargs.get("follow_redirects", True)
+
+    def _make_client(self):
+        return self._client_class(*self._client_args, **self._client_kwargs)
+
+
+class DoSyncRequest(BaseDoRequest):
+    _client_class = httpx2.Client
+
+    def __enter__(self):
+        self._client = self._make_client()
+        return self
+
+    def __exit__(self, *args):
+        self._client.close()
+        del self._client
+
+    @property
+    def client(self):
+        try:
+            return self._client
+        except AttributeError as e:
+            raise ValueError('To access sync client, use "with do_request() as client"') from e
+
+    def __call__(self, *args, **kwargs):
+        if hasattr(self, "_client"):
+            return self.client.request(*args, **kwargs)
+
+        # Use one-time context and dispose of the client afterwards
+        with self:
+            return self.client.request(*args, **kwargs)
+
+    def stream(self, *args, **kwargs):
+        if hasattr(self, "_client"):
+            with self.client.stream(*args, **kwargs) as response:
+                return b"".join(response.iter_bytes())
+
+        # Use one-time context and dispose of the client afterwards
+        with self, self.client.stream(*args, **kwargs) as response:
+            return b"".join(response.iter_bytes())
+
+
+class DoAsyncRequest(BaseDoRequest):
+    _client_class = httpx2.AsyncClient
+
+    def __enter__(self):
+        # Need to manage both loop and client, because client's implementation
+        # will fail if the loop is closed before the client's end of life.
+        self._loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self._loop)
+        self._client = self._make_client()
+        self._loop.run_until_complete(self._client.__aenter__())
+        return self
+
+    def __exit__(self, *args):
+        try:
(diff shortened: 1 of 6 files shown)


tests/integration/test_httpx2.py
```diff
@@ -0,0 +1,403 @@
...
+    transport = httpx2.MockTransport(lambda _: httpx2.Response(200, json={"text": "Hello, world!"}))
+
+    with vcr.use_cassette(str(tmpdir.join("mock_transport.yaml"))):
+        response = DoSyncRequest(transport=transport)("GET", url)
+
+    with vcr.use_cassette(str(tmpdir.join("mock_transport.yaml"))) as cassette:
+        cassette_response = DoSyncRequest(transport=transport)("GET", url)
+
+        assert cassette_response.status_code == response.status_code
+        assert cassette.play_count == 1
```
dsfaccini, line 394: David's AICA here: judgement call worth flagging — `test_custom_transports` keeps the `MockTransport` case but **omits** the `httpx_curl_cffi` and `pyreqwest` alt-transport blocks that the parallel `test_httpx.py` has.

Rationale: those transports are part of the **httpx** ecosystem — `httpx_curl_cffi.CurlTransport` / `pyreqwest...SyncHttpxTransport` subclass httpx's `BaseTransport`, so passing one to `httpx2.Client` (a separate fork with its own transport base) isn't a supported combination. There's no `httpx2`-flavored equivalent to exercise here, so mirroring them would only test an unsupported pairing. `MockTransport` is kept because `httpx2` ships its own and `_httpx2` patches it.

If you'd prefer them present-but-skipped (e.g. an explicit `pytest.skip` documenting the gap) rather than omitted, happy to switch.

seowalex:
Looked through this and looks very reasonable. Just one comment: why is the cassette duplicated for httpx2?

dsfaccini:
> Looked through this and looks very reasonable. Just one comment: why is the cassette duplicated for httpx2?

claude quirk, it doesn't need to be 

---

David's AICA here: good catch — it was a needless copy. `test_load_cassette_format` only checks that the legacy HTTPX cassette format deserializes, which is library-agnostic, so I've pointed the httpx2 test at the existing `gzip_httpx_old_format.yaml` and deleted the duplicate (776ad5e).

kevin1024:
Looks like some test failures

dsfaccini:
David's AICA here:

The red CI here is **not** from the httpx2 changes — it's the aiohttp 3.14 incompatibility tracked in #995 and fixed by #996.

aiohttp 3.14.1 (just released) removed `aiohttp.streams.AsyncStreamReaderMixin`, which `vcr/stubs/aiohttp_stubs.py` subclasses at import time. That throws on collection and takes down the **entire** suite (tornado, urllib3, even pure unit tests) — on `master` too, not just this branch.

Verified locally against the exact CI version (`aiohttp==3.14.1`):

- **This branch alone** → 48/181 unit tests fail with `AttributeError: module 'aiohttp.streams' has no attribute 'AsyncStreamReaderMixin'` (collection-time break).
- **This branch + #996's fix applied** → **253 passed**, including the full `test_httpx.py` and `test_httpx2.py`.

So once #996 lands on `master`, I'll rebase this PR and CI will go green. The httpx2 transport-patching here is independent of the aiohttp fix.

dsfaccini:
@kevin1024 seems like 996 needs to go first

kevin1024:
Ok, merged, but now we need a rebase

dsfaccini:
re pushed! but there's a weird failure, not sure I can fix that on my side @kevin1024

PR #1002, opened 2026-06-16
Forward streaming request bodies with Content-Length instead of chunked

kevin1024:
vcr forwarded file-like/iterator request bodies to the real server via the body property (a fresh BytesIO/iterator), so http.client chunk-encoded them. Chunked request bodies are rejected by some servers including the wsgiref-based test server, which closes the connection mid-send and raises BrokenPipeError (an intermittent failure in test_body_consumed_once). Since vcr already materializes the body for the cassette, forward those bytes via a new Request.body_bytes so http.client sets a Content-Length. Adds a regression test asserting the forwarded request uses Content-Length, not chunked.


kevin1024:
On reflection I'm not sure about this direction — it changes record semantics (chunked → Content-Length on passthrough) to work around a limitation in the wsgiref server pytest-httpbin uses. vcr altering how a request actually goes out the wire feels wrong, even if the body is already buffered for the cassette.

What if we remove the flaky test or change it instead — e.g. point it at a chunked-capable local server, or drop the chunked passthrough assertion? Leaving this open while I think about it.

PR #1003, opened 2026-06-18
fix: convert file-like request body to bytes to avoid pickle error

### LINKED ISSUE #737: Uploading file with aiohttp seems broken: TypeError: cannot pickle '_io.BufferedReader' object
When uploading files with aiohttp, the following error occurs.

Suspect [commit](https://github.com/kevin1024/vcrpy/commit/3b41f0ede302312895a5143514f0343302ba39aa).

Error
```
TypeError: cannot pickle '_io.BufferedReader' object
```

aiohttp code to reproduce:

```python
async def upload_file():
        with open("run.sh", "rb") as f:
            response = await session.post("https://example.com", data={"file": f})
```

decembercomposer697-hue:
Fixes #737: when uploading files with aiohttp, vcrpy receives a
BufferedReader which can't be pickled. Now converts file-like objects
to bytes.

### DIFF
--- a/vcr/stubs/aiohttp_stubs.py
+++ b/vcr/stubs/aiohttp_stubs.py
@@ -287,6 +287,11 @@ def vcr_request(cassette, real_request):
         if cookie_header:
             headers[hdrs.COOKIE] = cookie_header
 
+        # Convert file-like data to bytes to avoid pickle errors (GH#737)
+        if hasattr(data, 'read'):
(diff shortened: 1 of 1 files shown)


andrescollares:
Hey, could you add a test? thanks

decembercomposer697-hue:
Added unit tests as requested. Clean re-submit: #?

PR #1008, opened 2026-06-25
fix: show descriptive error when cassette contains unsupported YAML tags

### LINKED ISSUE #1007: Incorrect error mesage when loading python objects
Hi, after the change to use a safe YAML loader: https://github.com/kevin1024/vcrpy/commit/045acb1b5f8940f82adbb3faafd13e43e46289d6 ,if we try to load a casette that have a tag `tag:yaml.org,2002:python/object/new` it's raising an error: "Your cassette files were generated in an older version of VCR", even if we re-create the casettes with 8.2.1.  Probably the easiest change is to change that error message, for a more generic "There was a problem loading the casette".
Is there a way to allow custom python constructors, without using private methods? Something like this works if anyone is having the same problem:
```
yamlserializer._CassetteLoader.add_constructor(
    "tag:yaml.org,2002:python/object/new:some.route",
    some_converter_function,
)
```

gaoflow:
Fixes #1007.

## What changed

`vcr/serialize.py`'s `deserialize()` function caught both `ImportError` and `yaml.constructor.ConstructorError` in a single `except` clause and forwarded both to the same "Your cassette files were generated in an older version of VCR" message.

The `ConstructorError` path fires whenever the cassette contains a YAML tag that VCR's safe loader does not recognise, including `!!python/object/new:some.custom.Object` and similar tags. In that case the "older version" message can be misleading because the real problem is that the safe loader does not know how to construct that tag.

This PR splits the two exception cases into separate handlers:

- `ImportError` -> keeps the existing old-format cassette message.
- `yaml.constructor.ConstructorError` -> raises a new `ValueError` that includes the underlying unsupported-tag error and suggests deleting/re-recording or running the migration script only if the cassette is old.

The follow-up commit also removes the stale `git.io` migration link from both old-format and unsupported-tag error messages, because the short link now returns 404.

## Before

```text
ValueError: Your cassette files were generated in an older version of VCR.
Delete your cassettes or run the migration script.See http://git.io/mHhLBg for more details.
```

## After

```text
ValueError: There was a problem loading the cassette: could not determine a
constructor for the tag 'tag:yaml.org,2002:python/object/new:some.custom.Object'
  in "<unicode string>", line 3, column 11.
If this is an old cassette, delete it and re-record, or run the migration script.
```

## Tests

- Updated `test_deserialize_yaml_cassette_does_not_execute_python_object_tags` to remove the stale comment ("surfaced as the existing old-format error") — behaviour unchanged, `ValueError` is still raised, code is not executed.
- Added `test_deserialize_yaml_cassette_unknown_tag_gives_clear_error` which asserts the new message fires and the misleading "older version of VCR" string does not appear.
- Follow-up verification: `uv run --with-editable . --extra tests python -m pytest tests/unit/test_serialize.py tests/unit/test_persist.py -q` -> 22 passed.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/tests/unit/test_serialize.py
+++ b/tests/unit/test_serialize.py
@@ -30,11 +30,35 @@ def test_deserialize_yaml_cassette_does_not_execute_python_object_tags(tmpdir):
         f"_x: !!python/object/apply:os.system ['touch {marker}']\n"
         "version: 1\n"
     )
-    # The dangerous tag is rejected (surfaced as the existing old-format error)
-    # rather than executed.
+    # The dangerous tag is rejected with a ValueError rather than executed.
     with pytest.raises(ValueError):
         deserialize(malicious, yamlserializer)
     assert not marker.check(), "malicious cassette executed code during load"
+
+
+def test_deserialize_yaml_cassette_unknown_tag_gives_clear_error():
+    # A cassette with an unrecognized Python object tag (e.g. a custom class)
+    # used to raise "Your cassette files were generated in an older version of
+    # VCR", even for freshly re-recorded cassettes. It should now raise a
+    # clear, descriptive ValueError instead.
+    cassette_with_custom_tag = (
+        "interactions:\n"
+        "- request:\n"
+        "    body: !!python/object/new:some.custom.Object\n"
+        "      state: [hello]\n"
+        "    headers: {}\n"
+        "    method: GET\n"
+        "    uri: http://example.com/\n"
+        "  response:\n"
+        "    body: {string: ok}\n"
+        "    headers: {}\n"
+        "    status: {code: 200, message: OK}\n"
+        "version: 1\n"
+    )
+    with pytest.raises(ValueError, match="There was a problem loading the cassette") as excinfo:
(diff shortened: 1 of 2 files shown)


Polandia94:
The link https://git.io/mHhLBg points to a 404 error

> since old-format cassettes with Python object tags remain the most common trigger
I don't think that's true, the migration is very old

binggao1230:
Thanks, both points are fair. I pushed `aebf7c8` and updated the PR description.

Changes made:

- removed the stale `git.io` link from the new unsupported-tag error and from the existing old-format cassette error, since it now returns 404
- removed the claim that old-format Python object cassettes are the most common trigger; the message now only says to use the migration script if the cassette is old

Local verification:

```text
uv run --with-editable . --extra tests python -m pytest tests/unit/test_serialize.py tests/unit/test_persist.py -q
uv run --with ruff ruff check vcr/serialize.py tests/unit/test_serialize.py tests/unit/test_persist.py
git diff --check
```

PR #1009, opened 2026-06-28
disallow saving invalid casettes

Polandia94:
Related to https://github.com/kevin1024/vcrpy/issues/1007#issuecomment-4809579524
This create a safe dumper to match the safe loader,to avoid save casettes that can't be read. It also add a method `register_constructor `to allow python objects in both, serialization and deserialization.

Another option, instead of `register_constructor` is to create a function that call: `_CassetteDumper.add_representer(class_name,  some_callable)` or `add_multi_representer`, where the callable receive the dumper and the data and expect a valid yaml.Node. If you want @kevin1024,  I can change the PR on that direction.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -90,6 +90,31 @@ Finally, register your class with VCR to use your new serializer.
     with my_vcr.use_cassette('test.bogo'):
         # your http here
 
+Load cassettes containing custom Python objects
+-----------------------------------------------
+
+For security, VCR.py loads YAML cassettes with a *safe* loader that refuses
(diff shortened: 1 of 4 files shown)


kevin1024:
Thanks @Polandia94 — the dumper guard was exactly right, so I took it into #1012 with you as co-author. The one change: custom tags are scoped per VCR instance via `with_custom_tags(...)` + the existing `register_serializer` hook, instead of a module-global `register_constructor` — everything else in vcrpy is configured per-instance, so this keeps it consistent (and answers your "which direction" question). Closing in favor of #1012, which will go out in today's release.

PR #1014, opened 2026-07-09
Set raw_headers on replayed aiohttp responses (fixes #970)

### LINKED ISSUE #970: `aiohttp` stubs aren't setting `raw_headers`
https://github.com/karpetrosyan/httpx-aiohttp/pull/29 moving to use `raw_headers` exposes that `vcrpy`'s `aiohttp` stubs aren't setting `raw_headers`, and they should be: https://github.com/kevin1024/vcrpy/blob/v8.1.0/vcr/stubs/aiohttp_stubs.py#L70-L75

Here is a minimal reproducer with Python 3.13, `aiohttp==3.13.2`, `vcrpy==8.1.0`, and `pytest==9`:

```python
import aiohttp
import pytest
from vcr.request import Request
from vcr.stubs.aiohttp_stubs import build_response


@pytest.mark.asyncio
async def test_minimal_repro_raw_headers() -> None:
    async with (
        aiohttp.ClientSession() as session,
        session.get("https://httpbingo.org/gzip") as real_response,
    ):
        real_response.raise_for_status()
        body = await real_response.read()
    assert real_response.raw_headers is not None, "Response should have raw_headers"

    vcr_request = Request(
        real_response.request_info.method,
        str(real_response.request_info.url),
        None,
        {k: [v] for k, v in real_response.request_info.headers.items()},
    )
    vcr_response = {
        "status": {"code": real_response.status, "message": real_response.reason},
        "headers": {k: [v] for k, v in real_response.headers.items()},
        "body": {"string": body},
    }
    vcr_mock_response = build_response(vcr_request, vcr_response, [])
    # This assertion will fail
    assert (
        vcr_mock_response.raw_headers is not None
    ), "Mock response should have raw_headers"
```

---

Here's a starting point for the fix:

```python
response._raw_headers = tuple(
    (k.encode("utf-8"), v.encode("utf-8"))
    for k, v in response._headers.items()
)
```

apoorvdarshan:
## Summary

The aiohttp stub's `build_response()` (`vcr/stubs/aiohttp_stubs.py`) populated `response._headers` but never set `response._raw_headers`. aiohttp's `ClientResponse.raw_headers` simply returns `self._raw_headers`, which defaults to `None`, so a replayed response had `raw_headers=None` while a real response exposes the raw header bytes.

This breaks callers that rely on `raw_headers` under VCR replay — e.g. [httpx-aiohttp#29](https://github.com/karpetrosyan/httpx-aiohttp/pull/29), which motivated the issue.

Fixes #970.

## Fix

Derive `_raw_headers` from the already-deserialized `_headers` as UTF-8 encoded `(name, value)` byte tuples, mirroring what aiohttp populates from the real wire message:

```python
response._raw_headers = tuple(
    (k.encode("utf-8"), v.encode("utf-8")) for k, v in response._headers.items()
)
```

## Tests / Verification

- Added `test_aiohttp_test_client_raw_headers` in `tests/integration/test_aiohttp.py`. It uses the existing offline `aiohttp_client` local-server pattern, records a cassette, replays it, and asserts the replayed response's `raw_headers` is not `None`, is composed of `bytes` pairs, and is consistent with the decoded `headers`.
- Confirmed the test **fails** on `master` (`AssertionError: assert None is not None`) and **passes** with the fix.
- Full `tests/integration/test_aiohttp.py` module is green (30 passed) with aiohttp 3.14.1 / pytest 9.1.1 on Python 3.14.
- `ruff check` and `ruff format --check` (v0.15.17, matching `.pre-commit-config.yaml`) are clean on the changed files.
- Added a changelog entry under 8.3.0.

Disclosure: prepared with AI assistance; reviewed and verified locally.


docs/changelog.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 All help in providing PRs to close out bug issues is appreciated. Even if that is providing a repo that fully replicates issues. We have very generous contributors that have added these to bug issues which meant another contributor picked up the bug and closed it out.
 
 -  8.3.0
+    - Set ``raw_headers`` on replayed aiohttp responses so they match real responses (#970)
```
vEpiphyte, line 7: I believe this would need to start a `Unreleased` section.

apoorvdarshan: Added the `Unreleased` section at the top of `docs/changelog.rst` in the current PR head.

PR #1036, opened 2026-08-22
Preserve blank values when parsing query parameters

### LINKED ISSUE #1035: Query matcher ignores blank-valued parameters
### Reproduction

```python
from vcr.request import Request
from vcr import matchers

with_blank = Request("GET", "http://host/?a=b&flag=", "", {})
without_blank = Request("GET", "http://host/?a=b", "", {})

print(with_blank.query)  # currently [('a', 'b')]
matchers.query(with_blank, without_blank)  # currently matches
```

### Current behavior

`Request.query` uses `urllib.parse.parse_qsl` with its default `keep_blank_values=False`, so `flag=` disappears. The default query matcher therefore treats a request containing that parameter as equivalent to one where it is absent. Query filtering also cannot preserve it.

### Expected behavior

Blank-valued query parameters should be represented as `("flag", "")`. Reordered query parameters should still match, while presence versus absence of `flag=` should differ.

I have a one-line implementation change, a regression test, and a changelog entry ready.

### AI assistance disclosure

OpenAI Codex (GPT-5) assisted with investigation, the reproducer, regression test, implementation, and drafting this report. I reviewed the changes and ran the relevant tests locally.

liuyaohui666:
## Summary

`Request.query` currently drops parameters such as `flag=` because `parse_qsl` defaults to `keep_blank_values=False`. Consequently, the default query matcher treats a request containing that parameter as equivalent to one where it is absent, and query filtering cannot preserve it.

This change enables `keep_blank_values`, adds a regression test proving that ordering remains irrelevant while presence versus absence is distinguished, and records the behavior change in the changelog.

Fixes #1035.

## Verification

- `pytest tests/unit/test_matchers.py -q` — 29 passed
- `ruff check vcr/request.py tests/unit/test_matchers.py` — passed
- `ruff format --check vcr/request.py tests/unit/test_matchers.py` — passed
- The broader unit suite reached 187 passed; two tests lacked the optional `httpbin` fixture and two unrelated path-separator assertions are not Windows-compatible.
- `git diff --check`

## AI assistance disclosure

OpenAI Codex (GPT-5) assisted with investigation, the reproducer, regression test, implementation, and drafting this PR. I reviewed the diff and ran the checks listed above locally.


yangfan-yf-yf (approved):
I rechecked b1c72ec. Using keep_blank_values preserves flag= while sorting still keeps query matching independent of parameter order. The regression also distinguishes a blank-valued parameter from an absent parameter, and the changelog records the behavior.

Local validation on Windows with Python 3.12: the matcher and filter tests passed (53 passed), Ruff lint and format checks passed, and git diff --check passed.

